PR #35, opened 2014-05-30
Acpica 64bit

zetalog:
It is reported resuming failure if we enabled 64-bit FADT addresses favor:
https://bugzilla.kernel.org/show_bug.cgi?id=74021

The root cause is:
1. Linux only supports real mode resuming
  32bit waking vector address is used for real mode environment
  64bit waking vector address is used for 32bit or 64bit plat memory environment
2. BIOS may still look at 32-bit FACS reported in the FADT, setting 32bit waking vector of 64-bit FACS reported in the FADT doesn't work.

The series enables 64-bit favor with the exclusion of the above 2 cases
1. For case 1, it is OSPM feature, so we update AcpiSetFirmwareWakingVector() API so that OSPM can determine the value of the both vectors
2. For case 2, we introduce a new AcpiGbl_Use32BitFacsAddresses option to exclude this case, and we load both 32bit and 64bit FACS, set waking vector for both tables in order to be more robust.

For the rest of the addresses favor:
1. PM register addresses
2. DSDT address
We should be able to do a 64bit favor.

This series has been tested by the reporter.

### DIFF
--- a/source/components/hardware/hwxfsleep.c
+++ b/source/components/hardware/hwxfsleep.c
@@ -124,6 +124,12 @@
 /* Local prototypes */
 
 static ACPI_STATUS
+AcpiHwSetFirmwareWakingVector (
+    ACPI_TABLE_FACS         *Facs,
+    ACPI_PHYSICAL_ADDRESS   PhysicalAddress,
+    ACPI_PHYSICAL_ADDRESS   PhysicalAddress64);
+
+static ACPI_STATUS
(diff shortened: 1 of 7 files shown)


acpibob:
If this seems to work properly in Linux, we can apply to ACPICA.

zetalog:
It works.
And IMO linux will be facilitate with the ability to implement the 64-bit resume environment via this patchset.

Thanks and best regards
-Lv

PR #44, opened 2014-07-22
Acpica efi

zetalog:
This patchset is pulled from Yizhe Wang.
It contains basic EFI support code.
Please check.


zetalog:
Drop.
Will track the patch in efi-acpiexec.

PR #52, opened 2014-09-16
Acpica trace

zetalog:
The method tracing updates.

### DIFF
--- a/generate/linux/make-patches.sh
+++ b/generate/linux/make-patches.sh
@@ -28,7 +28,7 @@ usage()
 {
 	echo "Usage: `basename $0` [-r release] [-u] <old_commit> [new_commit]"
 	echo "Where:"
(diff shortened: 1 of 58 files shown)


zetalog:
updated due to BZ1135, also fixed a memory leak issue in one "Add interpreter tracing mode for method tracing facility".

zetalog:
It is reported that the address of AML opcode/method is not correct.
This is because we still use ACPI_OPERAND_OBJECT's address in the ACPICA log.
This patch uses AML address in the log and thus refreshes all patches.

Also we removed ACPI_NAMESPACE_NODE reference counting because we don't support this yet.
So no reference counting is allowed for ACPI_NAMESPACE_NODE.
Also because ACPI_NAMESPACE_NODE and ACPI_OPERAND_OBJECT doesn't inherit from same parent object, we shouldn't use reference counting API for it.

Thanks and best regards
-Lv

PR #54, opened 2014-09-30
iASL: Add two workarounds for a very old GCC bug.

juikim:
warning: cast from function call of type 'char *' to non-matching type 'long unsigned int'

### DIFF
--- a/source/compiler/aslmapenter.c
+++ b/source/compiler/aslmapenter.c
@@ -288,6 +288,7 @@ MpCreateGpioInfo (
     ACPI_GPIO_INFO          *Info;
     ACPI_GPIO_INFO          *NextGpio;
     ACPI_GPIO_INFO          *PrevGpio;
+    char                    *Buffer;
 
 
     /*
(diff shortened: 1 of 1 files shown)


krytarowski:
I confirm that this fixes problems on NetBSD. GCC from stable 7.x and current are affected.

zetalog:
IMO, it’s better to introduce GCC attribute((forced)) to fix ACPI_CAST_PTR or fix the return value of the function to make the type matched.
It will be a big problem to maintain such kind of code and regression will be easily triggered.

zetalog:
For this case, it seems not possible to fix the return value of the function to fix the warning.
So we may use attribute((forced)).
For example, we can define ACPI_FORCED as attribute((forced)) in acgcc.h, and an empty ACPI_FORCED in acmsvc.h.
Then deploys ACPI_FORCED for ACPI_CAST_PTR.

krytarowski:
I'm OK with the changes as long as GCC will build the sources properly.

PR #61, opened 2015-01-13
Acpica 1501 materials

zetalog:
This patchset includes materials that can be merged during 201501 release cycle.
They include:
1. The GPE patch to return register values.
2. 4 ACPICA community materials.
3. 1 Linux ACPI community material.
4. The 32-bit PAE kernel fix.
5. A script fix detected by kernel bugzilla reporter.

Please review.

### DIFF
--- a/generate/linux/gen-patch.sh
+++ b/generate/linux/gen-patch.sh
@@ -106,9 +106,9 @@ generate_acpica_desc()
 		AUTHOR_EMAIL="robert.moore@intel.com"
 	fi
 	AUTHOR="${AUTHOR_NAME} <${AUTHOR_EMAIL}>"
-	FORMAT="From %H Mon Sep 17 00:00:00 2001%nFrom: $COMMITTER%nData: %aD%nFrom: $AUTHOR%nSubject: [PATCH $INDEX] ACPICA: %s%n%n%b"
+	FORMAT="From %H Mon Sep 17 00:00:00 2001%nFrom: $COMMITTER%nDate: %aD%nFrom: $AUTHOR%nSubject: [PATCH $INDEX] ACPICA: %s%n%n%b"
 	if [ "x$UPSTREAM" = "xyes" ]; then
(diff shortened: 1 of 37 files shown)


zetalog:
1. The reporter of kernel bug 87971 (Paul Menzel) asked for a 3.16.7
   material. The original acpi_physical_address fix thus is split into
   several small patches to be a stable material candidate. I also fixed
   reporters' names in the patch description. The outcome of this task is
   the following commits:
    Linuxize: Fix several issues in the gen-patch.sh.…
    Unix: Cleanup to use ACPI_TO_INTEGER() to calc page offset.…
    Tables: Change AcpiFindRootPointer() to use ACPI_PHYSICAL_ADDRESS.…
    Executer: Cleanup to remove an unnecessary conversion.…
    Utilities: Cleanup to convert physical address printing formats.…
    Utilities: Cleanup to enforce ACPI_PHYSADDR_TO_PTR()/ACPI_PTR_TO_PHYS……
    Utilities: Cleanup to remove useless ACPI_PRINTF/FORMAT_xxx helpers.…
    Tables: Don't release ACPI_MTX_TABLES in AcpiTbInstallStandardTable().…
    Utilities: split IO address types from data type models.… 

2. I had a discussion with Rickard about the deletion of no-reference
   functions. In fact, such kernel requirement has affected ACPICA
   maintanence. Many "#ifdef" is added, and further split into seperate
   source files, finally many local variables have to be global variables.
   So I stated my opinion on the ground rule here:
   a. We should delete the symbol if there is no user even in the ACPICA
      upstream.
   b. We should move a tool specific symbol to the tool's source files.
   c. We could add some "#ifdef" for none kernel symbols, especially, for
      acpiexec/acpihelp/acpinames/iasl/debugger/disassembler. But we needn't
      split such code into new files.
   The outcome of this task is the following commits:
    Utilities: Remove unused AcpiUtCreatePkgStateAndPush().…
    Tables: Move an iasl specific table function to iasl source file.…
    Utilities: Correct conditional compilation definitions.…
    Resources: Correct conditional compilation definitions.…

zetalog:
Updates:
1. Utilities: split IO address types from data type models. …  
    Updated according to Len's comment.
2. Resources: Provide common part for ACPI_RESOURCE_ADDRESS structures. …  
    Thomas Gleixner and Jiang Liu requires this fix to proceed.
    I've asked Rafael to merge its linuxized version before ACPICA releasing it.

PR #67, opened 2015-03-13
Use portable uname option -a to unveil platform details

krytarowski:
Reference: http://pubs.opengroup.org/onlinepubs/009695399/utilities/uname.html

`uname -i` and `uname -o` aren't supported on NetBSD.


krytarowski:
Thanks!

PR #69, opened 2015-03-25
Fix invalid free

lkundrak:
Fixes a crash in case DisasmFilename can't be opened.
DisasmFilename is allocated from the string cache, it should not be
free()d.


acpibob:
There may be a couple of these lying around. Thanks for your help.

PR #83, opened 2015-06-17
Remove the macros potentially defined in ctype.h and string.h.

juikim:
"#undef foo" followed by "#define foo() foo()" may cause failures if any of
these are defined as macros in the system header files and if the system C
library does not contain one of those symbols.  Although it was intended to
preserve the previous behavior of ACPI_* macros, it is not portable because
POSIX allows them to be defined as such.

http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/ctype.h.html
http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/string.h.html

### DIFF
--- a/source/include/platform/acenv.h
+++ b/source/include/platform/acenv.h
@@ -419,52 +419,6 @@
 
 #endif /* ACPI_USE_STANDARD_HEADERS */
 
-/* We will be linking to the standard Clib functions */
-
-#undef strstr
-#undef strchr
-#undef strlen
-#undef strcpy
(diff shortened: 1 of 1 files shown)


juikim:
Yes, it works with GCC 4.8.

acpibob:
I not only removed the macros, I also needed to do some casting throughout the ACPICA code (which was the problem in the first place). These changes have been committed.

juikim:
Confirmed working, thanks!

PR #85, opened 2015-06-25
Efi fixes

zetalog:
-nostdinc is supported now.
printf() stuff is re-implemented using new C-lib style.
Please check.


zetalog:
Drop.
Track the patches in acpica-clib.

PR #86, opened 2015-06-26
MSVC: Fix inclusion order issues.

zetalog:
This patch fixes a regression introduced recently.


zetalog:
crtdbg.h has to be included after including stdlib.h.

PR #88, opened 2015-07-06
Acpica FACS

zetalog:
Linux upstream asked for keeping the old interfaces.
So I generated this commit.
The linuxized code of this commit has been merged by Linux upstream.
So this commit means divergences to us now.

Thanks and best regards
-Lv

### DIFF
--- a/source/components/hardware/hwxfsleep.c
+++ b/source/components/hardware/hwxfsleep.c
@@ -125,7 +125,7 @@
 
 #if (!ACPI_REDUCED_HARDWARE)
 static ACPI_STATUS
-AcpiHwSetFirmwareWakingVector (
+AcpiHwSetFirmwareWakingVectors (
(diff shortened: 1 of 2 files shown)


acpibob:
I should point out that it is getting more difficult to change interfaces as time goes on.
Does this change keep the new interface? It must.

zetalog:
The new interface is kept but renamed to AcpiSetFirmwareWakingVectors() so that OSPMs code won't get changed.
And the old interfaces are restored: AcpiSetFirmwareWakingVector() / AcpiSetFirmwareWakingVector64() so that old OSPM code can still link to them.
And if some OSPMs try to support both resuming environment, they can invoke the new interface AcpiSetFirmwareWakingVectors().

zetalog:
Will do it in Linux.

PR #96, opened 2015-08-25
Acpica debug

zetalog:
This patchset fixes several serious debugger issues.
And also contains a design cleanup for the in-kernel debugger and a command termination support for implementing in-kernel debugger's batch mode.

Please check.

### DIFF
--- a/generate/linux/libacpica.sh
+++ b/generate/linux/libacpica.sh
@@ -60,6 +60,7 @@ acpica_drivers_paths()
 	local paths incs inc
 
 	paths="\
+		components/debugger \
 		components/dispatcher \
 		components/events \
 		components/executer \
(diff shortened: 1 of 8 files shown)


zetalog:
This patchset releases ACPICA debugger to the Linux kernel.

zetalog:
Before applying this patchset:

1. acpiexec -h/-v/-? is broken, it either enters the command loop or breaks EFI builds.
2. exit/quit commands are broken for the in-kernel debugger. AcpiTerminate() shouldn't be invoked.
3. terminate command is broken, we can see segmentation fault.
4. "\n" echoed by the AcpiOsGetLine()cannot be seen by the in-kernel debugger which makes the command output mess.
5. the interactive mode of the debugger cannot be implemented inside of the kernel.
6. for the in-kernel debugger, we can see help messages related to "predefined" command but cannot execute it.

Thanks

zetalog:
Split last 2 patches into other thread.

PR #98, opened 2015-09-25
Acpica debug tid

zetalog:
New API to set debugger main thread ID.
This is on top of acpica-debug pull request.


zetalog:
The commit is included by another pull request (acpi-1509) as it is now divergences between Linux upstream and ACPICA.
So close it.

PR #99, opened 2015-09-25
Acpica debug osl

zetalog:
This is on top of acpica-debug/acpica-debug-tid.

Thanks and best regards
-Lv


zetalog:
Collected in https://github.com/acpica/acpica/pull/116.

PR #100, opened 2015-10-09
Tables: Fix FADT dependency regression

zetalog:
Some logics actually relying on the existence of FADT, currently relies on
the number of loaded tables. This false dependency can easily trigger
regressions. The reported regression can be seen on the following commit:
  Commit: c0b38b4c3982c2336ee92a2a14716107248bd941
  Subject: Tables: Fix global table list issues by removing fixed table
The commit changing the fixed table indexes results in the change of FADT
table index, originally, it was 3 (thus the installed table count should be
greater than 4), while currently it is 0 (and the installed table count may
be 3).

This patch fixes this regression by cleaning up the code. Reported by
Meelis Roos, Fixed by Lv Zheng.

Reference: https://bugzilla.kernel.org/show_bug.cgi?id=105351
Reported-and-tested-by: Meelis Roos <mroos@linux.ee>
Signed-off-by: Lv Zheng <lv.zheng@intel.com>


debox1:
Applied with merge conflicts fixed.

PR #102, opened 2015-10-14
Acpica 1509

zetalog:
ACPICA 1509 release divergences.

The commit is required for building debugger files.
Otherwise link error can be seen.

Thanks
-Lv


zetalog:
Updated to contain the 2 reverted patch.
The issue is root caused and fixed.
The 2 patches are required for this ACPICA release.
Otherwise build breakage can be seen around acpi_terminate() referenced by acpi_db_user_commands().
acpi_terminate() in Linux is marked __init, thus cannot be runtime executed by acpi_db_user_commands().
The Linux build system reports this issue during compile time.

Thanks
-Lv

zetalog:
Updated to contain the thread ID patch.
Without it, Linux kernel built with CONFIG_ACPI_DEBUGGER will fail to boot because of acpi_db_single_step().

Thanks and best regards
-Lv

PR #103, opened 2015-10-29
Acpica repair

zetalog:
String -> Reference conversion.


zetalog:
Updated to include a simple enhancement around AcpiGetObjectInfo().

PR #111, opened 2015-12-15
Acpica module level

zetalog:
We may need this patch so that when the feature is released to Linux, we have more capabilities to tune it from Linux kernel.


zetalog:
Close, will re-open after completing 20151218 release fix materials.

PR #114, opened 2015-12-25
Acpica 1512

zetalog:
Simple commits generated during 20151218 release to eliminate issues.

Thanks
-Lv

PR #117, opened 2016-01-19
Acpica init

zetalog:
It is proven that for _REV >= 2 platforms where ECDT is provided, EC._REG is not required to be invoked, this makes EmbeddedControl another early available operation region like SystemMemory/SystemIo/PciConfig that can be accessed during namespace initialization.
While other evaluations and region accesses may only happen after the namespace is initialized.

ACPI spec proposed a way to offer an ECAV method, in which, when _REV >= 2, EC accesses are allowed, otherwise, EC accesses are only allowed if EC._REG has been executed.
But many platofmrs are not implemented in this way, they set ECOK which should be set in EC._REG to indicate the availability of EC accesses in \\_SB._INI without providing ECAV.
In order not to break such platforms during the bug fixing of early operation region availibility, we need to ensure \\_SB._INI to be the first control method evaluated by the AML interpreter.

We have proven this fact via qemu.

### DIFF
--- a/source/compiler/aslstubs.c
+++ b/source/compiler/aslstubs.c
@@ -192,6 +192,13 @@ AcpiDsStoreObjectToLocal (
 }
 
 ACPI_STATUS
+AcpiEvInstallRegionHandlers (
+    void)
+{
+    return (AE_OK);
+}
+
+ACPI_STATUS
 AcpiEvQueueNotifyRequest (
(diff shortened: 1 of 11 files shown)


zetalog:
This pull request also contains ACPICA release fixes:
https://github.com/acpica/acpica/pull/114
The patches are based on the release fixed results.

PR #118, opened 2016-01-28
aclocal: Put parens around some definitions.

waddlesplash:
Looking at where these are used, this shouldn't result in any behavioral changes, but it's best practices to have them.


acpibob:
Of course. Thanks.
Bob

PR #121, opened 2016-02-19
Acpica tables

zetalog:
New table materials, generated to fix AcpiGetTable()/AcpiPutTable() regressions.

The first 2 patches are Linux materials, and the follow-ups are based on the Linux upstream code base as the feature is requested by Linux kernel developers, they need them to be able to apply on top of current code base.

Reported by Hans de Geode from redhat
Fixed by Lv Zheng

### DIFF
--- a/source/components/tables/tbdata.c
+++ b/source/components/tables/tbdata.c
@@ -158,6 +158,66 @@
 #define _COMPONENT          ACPI_TABLES
         ACPI_MODULE_NAME    ("tbdata")
 
+/* Local prototypes */
+
+static ACPI_STATUS
+AcpiTbCheckDuplication (
+    ACPI_TABLE_DESC         *TableDesc,
+    UINT32                  *TableIndex);
+
+static BOOLEAN
+AcpiTbCompareTables (
+    ACPI_TABLE_DESC         *TableDesc,
+    UINT32                  TableIndex);
+
+
+/*******************************************************************************
+ *
+ * FUNCTION:    AcpiTbCompareTables
+ *
+ * PARAMETERS:  TableDesc           - Table 1 descriptor to be compared
+ *              TableIndex          - Index of table 2 to be compared
+ *
+ * RETURN:      TRUE if both tables are identical.
+ *
+ * DESCRIPTION: This function compares a table with another table that has
+ *              already been installed in the root table list.
+ *
+ ******************************************************************************/
+
+static BOOLEAN
+AcpiTbCompareTables (
+    ACPI_TABLE_DESC         *TableDesc,
+    UINT32                  TableIndex)
+{
+    ACPI_STATUS             Status = AE_OK;
+    BOOLEAN                 IsIdentical;
+    ACPI_TABLE_HEADER       *Table;
+    UINT32                  TableLength;
+    UINT8                   TableFlags;
+
+
+    Status = AcpiTbAcquireTable (&AcpiGbl_RootTableList.Tables[TableIndex],
+        &Table, &TableLength, &TableFlags);
+    if (ACPI_FAILURE (Status))
+    {
+        return (FALSE);
+    }
+
+    /*
+     * Check for a table match on the entire table length,
+     * not just the header.
+     */
+    IsIdentical = (BOOLEAN)((TableDesc->Length != TableLength ||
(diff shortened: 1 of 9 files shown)


zetalog:
It has never been working without modifying ACPICA code.
This is a long term divergence.

ACPICA remembers mapped pointer in table_desc->pointer.
While there is no API for ACPICA to drop this pointer.
And this pointer is not valid after Linux is fully booted.
So there has been a long term divergence between Linux upstream and ACPICA:
Linux sets table_desc->pointer to NULL and returns mapped address/length in a modified ACPICA API.

The whole thing can be improved by introducing a reversal of AcpiGetTable().
In this reversal API, table_desc->pointer will be set to NULL.

The problem is:
1. what the name of the API should be:put should be a reversal of get
2. what should be implemented in this API: it should only contain unmap(table_desc->pointer), but given the fact that the pointer could also be used by ACPICA itself simulatneously, it won't be safe to just unmap right there. Thus I introduced refcnt to protect the unmap.

Thanks

zetalog:
The pull request only contains no-op cleanups and new APIs now.
Tested on a Linux laptop.

Thanks
-Lv

zetalog:
New table materials, generated to fix AcpiGetTable()/AcpiPutTable() regressions.

jwrdegoede:
Hi,

I see that this is not merged yet, what is the status of this  ?

Regards,

Hans

zetalog:
Bob is waiting for me to validate if Windows does ignore table signature when a table is dynamically loaded via Load/LoadTable opcodes.
I'll try to find a way to poke this out this week.

zetalog:
If that's true, ASLTS need to be cleaned up accordingly.

zetalog:
I completed the following test:
1. prepare an OEM table, signature is ABCD
2. prepare a LoadTable call to load it into "\", and check if the object belonging to that table exists
3. Invoke the LoadTable call from \_SB._INI

Table has been successfully loaded, and you can see screen capture in the below attachment.
tablesignature

zetalog:
tablesignature

zetalog:
I'm going to correct the ASLTS.

zetalog:
loadsignature

zetalog:
I found a case broken, so we may need to do the re-work.

If a table is loaded from Buffer, we actually:
1. need to do "compare table" in table install step,  otherwise installed tables will be grown endlessly.
2. need to return AE_ALREADY_EXISTS to let the interpreter to generate an exception to the AML code.

The problem is actually:
1. We cannot do this very early in OS early stage
2. So when we actually executes the table in the OS late stage.

So for the early stage tables, we have to let them "installed" to the global table list.
And for the late stage, we shall prevent them from being installed.

It looks we need to do table comparison
1. both in "install" for dynamic loading tables, and
2. in "load" for static loading tables, and do not return errors for that case.

zetalog:
However I validated one more thing:
Windows doesn't allow "Load" from Buffer/Field, it only allows to "Load" from OperationRegion.

That means it should be possible to use pointer sanity check to return such AE_ALREADY_EXISTS exception.

zetalog:
loadregion

zetalog:
loadfield

zetalog:
loadbuffer

zetalog:
To: Hans

The feature you requested seems should still be based on my old comments made in acpica devel mailing list:
1. Add a flag to indicate if an installed table has been verified.
2. Verify a table when acpi_gbl_verify_table_checksum is changed to true.
3. And always verify a table before installing it to global table list, instead of verifying it before loading.
That's the way won't break things.

I prepared a pull request for this:
https://github.com/acpica/acpica/pull/265

And I'll send linuxized them for you to test.

zetalog:
Rebased to contain linux upstreamed materials:
Tables: Fix a hidden logic related to AcpiTbInstallStandardTable() …  
Tables: Add mechanism to allow to balance late stage AcpiGetT… …  

And simple changes:
Tables: Cleanup table handler invokers …  
Tables: Add sanity check in AcpiPutTable() …   

And table signature changes according to the Windows validation result:
Tables: Do not validate signature for dynamic table load …  
ASLTS: exc_tbl: Fix TLD1 tests …   

Thanks

zetalog:
Closing as #281 has been merged.
Will communicate and upstream ASLTS fix in a different way.

PR #126, opened 2016-03-21
Acpica access

zetalog:
The build issue should be fixed in the macro definition, not the caller side.
Going to add a fix to remove wrong ACPI_ROUND_UP() from AcpiHwRead()/AcpiHwWrite(). This was erroneously introduced by the previous BitOffset fix.


zetalog:
This is released to Linux community and is contained by the pull request 128.
So let me close it.

PR #127, opened 2016-03-22
Improve compatibility with POSIX yacc.

juikim:
- byacc has a bug in "-o" option parsing.  Prefix may not be set properly
if its argument contains ".c" somewhere before the actual output file name.
With "byacc -ofoo.cXXXXX/foo.c -d foo.y", for example, it generates "foo.h"
instead of "foo.cXXXXX/foo.h" because ".cXXXXX/foo.c" is replaced with ".h".
- Prefer a GNU make built-in function, suffix(), rather than a series of
complex variable substitutions, to improve readability.
- Exit with a meaningful exit status to make it more like a real function.

### DIFF
--- a/generate/unix/iasl/Makefile
+++ b/generate/unix/iasl/Makefile
@@ -274,9 +274,11 @@ include ../Makefile.rules
 # Function to safely execute yacc
 #
 safe_yacc = \
-	_d=`mktemp -d $(OBJDIR)/$(1)XXXXXX` &&\
-	$(YACC) $(YFLAGS) -p$(1) -o$$_d/$(1).c -d $(2) &&\
-	mv $$_d/$(1)$(suffix $(3)) $(3);\
+	_d=`mktemp -d $(OBJDIR)/$(1).XXXXXX` &&\
+	cd $$_d &&\
+	$(YACC) $(YFLAGS) -d -p$(1) $(abspath $(2)) &&\
+	cd - &&\
+	mv $$_d/y.tab$(suffix $(3)) $(3);\
 	_r=$$?;\
 	rm -fr $$_d;\
 	exit $$_r
--- a/generate/unix/iasl/Makefile
+++ b/generate/unix/iasl/Makefile
@@ -274,12 +274,12 @@ include ../Makefile.rules
 # Function to safely execute yacc
 #
 safe_yacc = \
-	_f=`echo $(1) | tr '[:upper:]' '[:lower:]'` &&\
-	_d=`mktemp -d $(OBJDIR)/$$_f.XXXXXX` &&\
-	_t=`basename $(3)` &&\
-	$(YACC) $(YFLAGS) -p$(1) -o$$_d/$$_f.c -d $(2) &&\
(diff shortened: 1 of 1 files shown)


acpibob:
I'm not sure that we can use a GNU built-in; things may not work with some vendors.

juikim:
Actually, you were already using GNU make built-in functions, e.g., `findstring`, `shell`, `wildcard`, etc.

acpibob:
Ok, we’ll probably take the patch.

zetalog:
I'm a bit worried about the side effect of removing -o.
With -o, make dependencies are natural.
But with the new safe_yacc, make dependencies will be messed up by intermediate generations and problems can be caused in parallel build.

juikim:
FYI, current implementation already uses intermediate directories & files.  Actually, this patch makes parallel build more safer and I tested it with '-j60' on a 40-core system multiple times.

acpibob:
Sounds interesting, we’ll try it out.
Thanks,
Bob

zetalog:
Then the only question left is:
Is the default output name is ensured by all yacc clones?
OTW, Will errors be reported by "mv", complaining no source file?
  mv $$_d/y.tab$(suffix $(3)) $(3);

Thanks
-Lv

juikim:
> Is the default output name is ensured by all yacc clones?

If it is POSIX-compliant, then yes.

http://pubs.opengroup.org/onlinepubs/9699919799/utilities/yacc.html

> OTOW, will errors be reported by "mv", complaining no source file?

Of course.

Moorre:
If path contains a Space character or other special symbols, the compilation will stop with error:

Mac-Pro-Moorre:~ moorre$ cd /Volumes/2\ UNIT\ 2T/MAC\ OS\ X\ \&\ HACK/ACPI/iASL_MAC/ACPICA_Source/acpica-master_20171214/generate/unix
Mac-Pro-Moorre:unix moorre$ make iasl
-  m4           Preprocess main iASL parser
-  bison        obj/aslcompiler.y
/bin/sh: HACK/ACPI/iASL_MAC/ACPICA_Source/acpica-master_20171214/generate/unix/iasl/obj/aslcompiler.y: No such file or directory
mv: rename /y.tab.h to obj/aslcompiler.y.h: No such file or directory
make[1]: *** [obj/aslcompiler.y.h] Error 1
make: *** [iasl] Error 2
Mac-Pro-Moorre:unix moorre$ /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/bison: extra operand UNIT
Try /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/bison --help for more information.

Look at line:
```/bin/sh: HACK/ACPI/iASL_MAC/ACPICA_Source/acpica-master_20171214/generate/unix/iasl/obj/aslcompiler.y: No such file or directory```
Otherwise, from the home directory, all looks fine.

juikim:
Try this [patch](https://github.com/acpica/acpica/files/1602105/yacc-patch.txt).

PR #130, opened 2016-03-29
Acpica access

zetalog:
Just for review purpose as the solution of BZ 1270 hasn't been determined and the patch order hasn't been fixed.

Thanks
-Lv

### DIFF
--- a/source/components/executer/exfldio.c
+++ b/source/components/executer/exfldio.c
@@ -1018,20 +1018,9 @@ AcpiExInsertIntoField (
 
     AccessBitWidth = ACPI_MUL_8 (ObjDesc->CommonField.AccessByteWidth);
 
-    /*
-     * Create the bitmasks used for bit insertion.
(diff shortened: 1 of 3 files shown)


zetalog:
Hi, since the 201603 linux release cycle has been merged with the AcpiHwRead()/AcpiHwWrite() enhancements dropped, we are not in a hurry now.
Please keep this pull request as a review/discussion material.
I'll add follow-up commits to this pull request before its merging.

Thanks
-Lv

acpibob:
We will need to revert the changes from the git tree for the next ACPICA release. Other reports of this are coming in from other non-Linux customers.

zetalog:
I saw the bug report.
It can be fixed by this commit:
"Hardware: Add optimized access bit width support …“

Probably you should merge this pull request to close this task.
And I can take care of re-ordering Linuxized patches during the next release cycle.

Thanks
-Lv

PR #132, opened 2016-04-13
Acpica module

zetalog:
A new fix was required by a bugzilla reporter.
This fixes a known issue:
_REG(SystemMemory, ..)
_REG(SystemIo, ...)
shouldn't be executed.

Please check.

Thanks
-Lv

### DIFF
--- a/source/components/events/evhandler.c
+++ b/source/components/events/evhandler.c
@@ -222,57 +222,6 @@ UnlockAndExit:
 
 /*******************************************************************************
  *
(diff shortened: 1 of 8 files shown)


zetalog:
I noticed that the comments in evregion.c need to be refreshed and updated the patch.

zetalog:
I got update that we can do this easier due to Windows behavior.

Thanks
-Lv

tapir:
So was there an alternative way? Can you point me to the correct bug report?

PR #133, opened 2016-04-20
acpixtract: use %u format specifier on unsigned int Gbl_TableCount

ColinIanKing:
Minor issue, Gbl_TableCount is an unsigned int, so we should be using
a %u format specifier instead of %d.


acpibob:
Got it, thanks.

PR #134, opened 2016-04-22
Acpica parser

zetalog:
The interpreter enhancement of TermList support for table loading.

It should be OK to merge, because the feature is protected by a global option.
We still have chance to optimize it.

Test cases about this change can be found at: https://github.com/acpica/acpica/pull/135.

Thanks
-Lv

### DIFF
--- a/source/compiler/aslstubs.c
+++ b/source/compiler/aslstubs.c
@@ -119,6 +119,7 @@
 #include "acevents.h"
 #include "acinterp.h"
 #include "acnamesp.h"
+#include "acparser.h"
 
 #define _COMPONENT          ACPI_COMPILER
         ACPI_MODULE_NAME    ("aslstubs")
@@ -138,6 +139,13 @@ AcpiNsExecModuleCodeList (
 ACPI_STATUS
 AcpiNsInitializeObjects (
     void)
+{
+    return (AE_OK);
+}
+
+ACPI_STATUS
+AcpiPsExecuteTable (
+    ACPI_EVALUATE_INFO      *Info)
 {
     return (AE_OK);
 }
--- a/source/components/dispatcher/dsopcode.c
+++ b/source/components/dispatcher/dsopcode.c
@@ -554,6 +554,11 @@ AcpiDsEvalRegionOperands (
     /* Now the address and length are valid for this opregion */
 
     ObjDesc->Region.Flags |= AOPOBJ_DATA_VALID;
+    if (WalkState->ParseFlags & ACPI_PARSE_MODULE_LEVEL)
+    {
+        Status = AcpiUtAddAddressRange (ObjDesc->Region.SpaceId,
+            ObjDesc->Region.Address, ObjDesc->Region.Length, Node);
+    }
     return_ACPI_STATUS (Status);
 }
 
--- a/source/components/events/evrgnini.c
+++ b/source/components/events/evrgnini.c
@@ -690,7 +690,8 @@ AcpiEvInitializeRegion (
                  *
                  * See AcpiNsExecModuleCode
                  */
-                if (ObjDesc->Method.InfoFlags & ACPI_METHOD_MODULE_LEVEL)
+                if (!AcpiGbl_ParseTableAsTermList &&
+                    ObjDesc->Method.InfoFlags & ACPI_METHOD_MODULE_LEVEL)
                 {
                     HandlerObj = ObjDesc->Method.Dispatch.Handler;
                 }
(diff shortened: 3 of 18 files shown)


acpibob:
I want to do more testing on this now that we have a compiler that implements the TermList instead of ObjectList.

zetalog:
I think the compiler change was just a result of this report and this interpreter change.
This fix was sent to the ACPI mailing list last year.
Along with the report of the TermList support in Device/Scope.
You can find the report in the cover-letter of that series.
The TermList support in Device/Scope was revealed by my testing using qemu.
This is the motivation for the ACPICA compiler change.

I'm OK with the testing.
The more testing, the better quality.
But let me be aware of the test case, result.

Thanks and best regards
-Lv

acpibob:
it appears that this change does far more than just supporting TermList.

Also, TermList works today (with change ObjectList->TermList for things like devices), so it is not clear to me why this change is needed.

acpibob:
In any case, please split it up a bit. (TermList, Tests, other...etc.)

acpibob:
In ASLTS, please don't do stores to the Debug object, as it will mess up the existing output, I believe.

acpibob:
Unless debug object is being redirected to a file.

acpibob:
+ * Optionally support module level code by parsing the entire table as
+ * a TermList. Default is TRUE, do execute entire table.
+ * For the disassembler, this is FALSE for now in order not to avoid
+ * introducing too many executer function stubs.

Please explain. I would rather just move to the new grammar and leave the old grammar behind.

acpibob:
As far as description of the TermList stuff and whole-table execution, add "the original implementation was limited to certain opcodes and was not complete."

acpibob:
Overall, looks like it will work well. Seems to work here.

zetalog:
I see, it means I should change the ASLTS case to append to the Debug object.
I'll split this and improve the ASLTS cases.

Thanks

zetalog:
> Please explain. I would rather just move to the new grammar and leave the old grammar behind.
This patch doesn't change disassembler and keeps old behavior for the disassembler.
So that the patch will be minimized.
IMO, disassembler should be changed accordingly after merging this commit.
After that, we can remove this description.

Thanks

zetalog:
> Overall, looks like it will work well. Seems to work here.

Yes, the ASLTS now works well with both interpreter/assembler enabled for the "TermList parsing support in table loading" and "TermList parsing support in scope create/open opcodes".

zetalog:
Pull request #135 can be used to confirm if this is working.
And this commit should be required to make the ASLTS cases implemented in pull request #135  working.

zetalog:
I added a new patch, fixing the bug reported at http://bugs.acpica.org/show_bug.cgi?id=1279.
The problem will be worse when the MLC fix is merged.

Thanks and best regards
-Lv

zetalog:
This patch should have also solved the forward referencing issues for BufferField, BankField, Buffer, Package, Region.

With old support, it appears to be the forward referencing in the arguments of the above opcodes is supported by the ACPICA interpreter, while the investigation result shows that Windows doesn't support the forward referencing even in the arguments.

Per linux bugzilla report, the "initializing ...." debugging informations are lost due to the grammar switch. Then I checked the code, it shows that this is just forward referencing related. And nothing was lost after switching the graamer but the debugging messages. The exception is the Region opcode, we have special logic to track address range in the deferred one, so I added this back into the 1st commit.

Thanks and best regards
-Lv

zetalog:
I removed the regression protection for the disassembler after validating that it is working for all my local tables.
The patch was updated accordingly with patch description enhanced to reflect the BufferField/BankField/OperationRegion/Buffer/Package concerns.

Thanks and best regards
-Lv

zetalog:
I improved the ASLTS cases, and noticed that If/Else/While code blocks under Device/PowerResource/Processor/ThermalZone/Scope were also supported by the original interpreter in a deferred way.
So I updated the patch description to make it clearer.

Thanks and best regards
-Lv

zetalog:
Cases of Type2Opcode support is split into this pull request while cases of Type1Opcode support is in pull request #135. I also added a test case demonstrate the order issue.

zetalog:
Updated with cleanup code fixed.
I was surprised the issue was still there.
I've fixed it before.
Sorry for leaving the same issue in the pull request for twice.

I just have several environments (VC6 and Linux) and the patches were not synced between the environments.

Best regards
-Lv

zetalog:
Experiment is going on in this link:

https://bugzilla.kernel.org/show_bug.cgi?id=120041

Thanks
-Lv

zetalog:
Split one regression fix from the grammar change and make it cleaner.

PR #135, opened 2016-04-27
Acpica aslts

zetalog:
ASLTS changes related to the TermList parsing.
Note that the cases brought by this change can only report OK after merging the following pull request:
https://github.com/acpica/acpica/pull/134

Thanks
-Lv


zetalog:
I improved the case to prove that ACPICA supports If/Else/While under Scope/Device/PowerResource/Processor/ThermalZone.

Thanks and best regards
-Lv

PR #136, opened 2016-04-27
Fix deconstification warnings with '-Wcast-qual'

juikim:
These changes fixes casting `const char *` to `char *` and `void *` warnings.  It is critical for FreeBSD kernel.


acpibob:
ok, we'll get them in the next acpica release.
thanks.

PR #141, opened 2016-05-27
Acpica access2

zetalog:
Alternative solution by fixing AcpiTbInitGenericAddress().


zetalog:
This is just an RFC pull request.
We need to do more unit tests around the commits in it.

acpibob:
Please explain this change in more detail.

zetalog:
Let me drop this fix first.
This is a fix fixing AcpiTbInitGenericAddress() by correctly setting AccessWidth field.
Currently it is said that the breakage can be stopped by previous workaround.

Thanks and best regards
-Lv

PR #148, opened 2016-06-23
Acpica debugger

zetalog:
New OSLs.


zetalog:
This patch can also fix acpiexec termination issue by sorting termination logic strictly:
1. initialize ACPICA core
2. initialize debugger and enter the loop
3. terminate debugger and exit the loop
4. terminate ACPICA core

Thanks

PR #150, opened 2016-06-30
Acpica lock

zetalog:
Lock order is wrong.
During boot, it is namespace -> interpreter.
While we should always ensure interpreter -> namespace.
Linux doesn't trigger this because this happens in boot stage, there is no parallel running threads invoking ACPICA at that time.

Thanks
-Lv

### DIFF
--- a/source/components/namespace/nsload.c
+++ b/source/components/namespace/nsload.c
@@ -161,21 +161,6 @@ AcpiNsLoadTable (
     ACPI_FUNCTION_TRACE (NsLoadTable);
 
 
-    /*
-     * Parse the table and load the namespace with all named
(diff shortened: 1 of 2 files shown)


zetalog:
Hi,

I saw you reverted TermList stuff.
I got it.
I'll do the namespace lock assessment first.
I'll lock namespace mutex in AcpiNsLookup.
And unlock around its invocation, then drop the wrong ordered lock.

But in order to support namespace unloading, we still need to introduce a namespace reference flushing mechanism which isn't necessary to be per-node, but could be a count for the entire namespace.

I'll close this pull request.

Thanks
-Lv

PR #152, opened 2016-07-05
generate/unix/Makefile.common: remove HARDWARE_NAME

tpetazzoni:
The HARDWARE_NAME variable is only used to display some messages saying
whether a 32-bit or a 64-bit build was done, which is pretty
useless. This commit gets rid of this variable entirely.

Signed-off-by: Thomas Petazzoni thomas.petazzoni@free-electrons.com

### DIFF
--- a/generate/unix/Makefile.common
+++ b/generate/unix/Makefile.common
@@ -2,34 +2,11 @@
 # Common make for acpica tools and utilities
 #
 
-#
-# Get the OS machine architecture. Anything with a "64" in the returned
-# string will be treated as a 64-bit OS. Otherwise, the default is 32-bit.
-#
-ifeq ($(HOST), _FreeBSD)
-HARDWARE_NAME := $(shell uname -p)
-else
-ifeq ($(HOST), _QNX)
-HARDWARE_NAME := x86pc
-else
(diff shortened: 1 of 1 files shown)


zetalog:
Looks good.

acpibob:
Theoriginal intent of this was to validate that cross-platform builds were generated correctly (by the correct compiler). Such as 32-bit build on 64-bit OS and 64-bit build on a 32-bit OS.

acpibob:
Given the previous comment, I would rather just leave this in the makefile.

tpetazzoni:
I don't see how it validates anything. It's *only* use is for "echo" statements, nothing else.

Also, 32-bit vs. 64-bit is a pretty useless distinction, especially when (like we're doing in Buildroot) your concern is about cross-compilation. In this case, it's not so much 32 bit vs. 64 bit that matters, but what is the target architecture (x86, ARM, MIPS, etc.)

acpibob:
It's at least useful to us.

acpibob:
We don't think that any of our customers (other operating systems) need this, so we will accept the patch.
Thanks,
bob

acpibob:
I also went ahead and removed the "bit width" from the signon of all of the ACPICA user space tools.

PR #156, opened 2016-07-24
Fix deprecated functions when building for __APPLE__

metacollin:
The current ACPICA code will not compile with gcc or clang if building under OS X as it currently stands.  It calls two deprecated functions, sem_destroy and tmpnam.  Tmpnam is very insecure and can be used quite easily as a vehicle for privilege escalation via interprocess attacks.  mktemp behaves identically, including the way in which it fails (returns -1 etc), but is not deprecated and is secure.  It can be used as a simple drop-in replacement.  It is also safely tucked away inside an #ifdef __APPLE__ code block, so the effect it has is perfectly scoped.  

The other problem is that in a header file specific to the __APPLE__ platform, a macro is already in place and is clearly an attempt to expand sem_destroy (which is deprecated to the point that it is not even actually implemented in OS X, it is an optional part of the POSIX standard) with a perfectly valid alternative, sem_close.  Unfortunately, this macro is defined in a header file that is always included well before <semaphore.h> is included, so the declaration of sem_destroy (which is also where it is explicitly marked as deprecated) in semaphore.h is, just like was defined, expanded to sem_close by the preprocessor.  This results in sem_close being declared as deprecated, and wholly incorrectly, and the end result is nothing is actually solved.  So I simply made sure any such define was removed directly before including semaphore .h, then reapplied after, so it would have the intended effect throughout the rest of the code without the macro expansion polluting the semaphore.h system header.  

Overall, this is almost on the level of a cosmetic change, and only effects OS X users, but it would be nice if typing `make` just worked for us too :).  

Thanks!

### DIFF
--- a/source/include/platform/acmacosx.h
+++ b/source/include/platform/acmacosx.h
@@ -119,6 +119,7 @@
 #include "aclinux.h"
 
 #ifdef __APPLE__
+#undef sem_destroy
 #include <semaphore.h>
 #define sem_destroy         sem_close
 #define ACPI_USE_ALTERNATE_TIMEOUT
--- a/source/include/platform/acmacosx.h
+++ b/source/include/platform/acmacosx.h
@@ -119,6 +119,7 @@
 #include "aclinux.h"
 
 #ifdef __APPLE__
+#include <semaphore.h>
 #define sem_destroy         sem_close
 #define ACPI_USE_ALTERNATE_TIMEOUT
 #endif /* __APPLE__ */
--- a/source/os_specific/service_layers/osunixxf.c
+++ b/source/os_specific/service_layers/osunixxf.c
@@ -829,7 +829,7 @@ AcpiOsCreateSemaphore (
 
 #ifdef __APPLE__
     {
-        char            *SemaphoreName = tmpnam (NULL);
+        char            *SemaphoreName = mktemp (NULL);
 
         Sem = sem_open (SemaphoreName, O_EXCL|O_CREAT, 0755, InitialUnits);
         if (!Sem)


metacollin:
I should mention that I did run the test suite on my modest changes and they did not cause any tests to fail.

source/include/platform/acmacosx.h
```diff
@@ -119,6 +119,7 @@
 #include "aclinux.h"
 
 #ifdef __APPLE__
+#undef sem_destroy
 #include <semaphore.h>
 #define sem_destroy         sem_close
```
zetalog: Where is the original position of this line?
Could we move semaphore.h inclusion prior than that line?

Thanks
-Lv

source/include/platform/acmacosx.h
```diff
@@ -119,6 +119,7 @@
 #include "aclinux.h"
 
 #ifdef __APPLE__
+#include <semaphore.h>
 #define sem_destroy         sem_close
```
zetalog: What if in osunixxf.c:

#ifdef __APPLE__
#include <semaphore.h>
#endif

#include other files

metacollin: This was initially what I did, simply include semaphore before acpi.h with an #ifdef.  But I thought this might be considered 'hacky' or otherwise might be met with disapproval (for no good reason, obviously.  Sorry, I think I just over thought this entire thing!).  But if you have no problems with that solution, then I'll update my pull request to do this instead.  Thanks!

zetalog: That really sounds hackiy. :-)

Then IMO,
The first choice might be:
Converting sem_destroy to sem_close in osunixxf.c.

Does APPLE build always include <semaphore.h>?
I think the problem is the following line:
#include "aclinux.h"
If so, probably the second choice could be writing the following in acmacosx.h:
#define ACPI_USE_SYSTEM_CLIBRARY
#define ACPI_USE_STANDARD_HEADERRS

#ifdef ACPI_USE_STANDARD_HEADERRS
#include <semaphore.h>
#endif 

#include "aclinux.h"

acpibob:
I'll take this when there is agreement on it.
bob

PR #157, opened 2016-07-26
Fix acpidump.efi build

jmarcelof:
Fix 'logical-op' warning generated by gcc 6.0 or above regarding conditionals
as "if (File == stdout || File == stderr)" since in that environment both
stdout and stderr were expanded to ST->ConOut.

Signed-off-by: Marcelo Ferreira <joaomarcelo@lesc.ufc.br>

### DIFF
--- a/source/include/platform/acefi.h
+++ b/source/include/platform/acefi.h
@@ -332,6 +332,6 @@ extern struct _ACPI_EFI_BOOT_SERVICES       *BS;
 
 #define FILE                struct _ACPI_SIMPLE_TEXT_OUTPUT_INTERFACE
 #define stdout              ST->ConOut
-#define stderr              ST->ConOut
+#define stderr              ST->StdErr
 
 #endif /* __ACEFI_H__ */


acpibob:
This may be fixed with pull #158 which was just merged. Let us know.
bob

jmarcelof:
Actually it was not fixed. This problem is related with the new versions of gcc (6.0 or above) as reported at https://gcc.gnu.org/bugzilla/show_bug.cgi?id=69602 .

Thanks

zetalog:
There can be another way to fix this issue.
For example, preparing 2 different global variables for stderr, stdout.
So that we can still have -Wlogical-op enabled.

Thanks

jmarcelof:
That's what I did. Instead of disabling the warning, I changed stderr to ST->StdErr.

zetalog:
Looks reasonable.
Have you tested it?
If it is working, we can switch to use this text output interface.

zetalog:
I just tried acpidump.efi -c "abc" which should leave error messages to the users.
But with stderr defining to ST->StdErr, nothing can be seen from the console.
So your change is not sufficient IMO.

jmarcelof:
Indeed ST->StdErr seems to be invalid in some environments so I'll open a new pull request with a patch that always uses ST->ConOut and fix the issue with gcc 6+.

PR #159, opened 2016-07-28
Fix efi build with gcc 6.0+

jmarcelof:
Avoid 'logical-op' warning caused by stdout and stderr (both defined as ST->ConOut).
This patch replaces the stdout and stderr defines by global variables (both initialized to point to ST->ConOut).


zetalog:
Looks OK to me.

PR #161, opened 2016-08-04
Acpica gas

zetalog:
Current GAS code in Linux upstream.
AcpiHwWrite() support is not included because of issue related to the predefined FADT register length.
However AcpiHwRead() won't be affected.
So it is tested in Linux upstream for months and is proven to be safe.

Our new GAS improvement will be based on this.

Thanks

### DIFF
--- a/source/components/hardware/hwregs.c
+++ b/source/components/hardware/hwregs.c
@@ -285,10 +285,7 @@ AcpiHwValidateRegister (
  *              64-bit values is not needed.
  *
  * LIMITATIONS: <These limitations also apply to AcpiHwWrite>
- *      BitWidth must be exactly 8, 16, or 32.
  *      SpaceID must be SystemMemory or SystemIO.
- *      BitOffset and AccessWidth are currently ignored, as there has
- *          not been a need to implement these.
  *
  ******************************************************************************/
 
@@ -298,7 +295,12 @@ AcpiHwRead (
     ACPI_GENERIC_ADDRESS    *Reg)
 {
     UINT64                  Address;
+    UINT8                   AccessWidth;
+    UINT32                  BitWidth;
+    UINT8                   BitOffset;
     UINT64                  Value64;
+    UINT32                  Value32;
+    UINT8                   Index;
     ACPI_STATUS             Status;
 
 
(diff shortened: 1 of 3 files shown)


acpibob:
I thought that there were issues with using the BitOffset field because some machines had garbage in this field (or perhaps accesswidth)

zetalog:
No garbage BitWidth is reported.
There is a breakage report against garbage AccessWidth and we haven't root caused it.

IMO, This code is stable because it is able to fall back to non-GAS support "when 32-bit favor is set AccessWidth can be set 0".
And the code has been tested in Linux upstream for months, no breakage is reported against read operation.

zetalog:
I'll put AcpiHwWrite() into this pull request as it looks correct.
All issues seem to be related to AcpiHwGetAccessBitWidth().
AcpiHwGetAccessBitWidth() may still be improved according to ASWG feedbacks.
So we should hold this on and wait for the feedbacks.

Thanks

source/components/hardware/hwregs.c
```diff
@@ -164,31 +164,56 @@ AcpiHwGetAccessBitWidth (
...
+    {
+        AccessBitWidth = (1 << (Reg->AccessWidth + 2));
     }
     else
     {
-        return (1 << (Reg->AccessWidth + 2));
+        AccessBitWidth = ACPI_ROUND_UP_POWER_OF_TWO_8 (
+            Reg->BitOffset + Reg->BitWidth);
+    }
+
```
zetalog: IMO, the previous algorithm should be correct.
As it always treat AccessSize as good configuration.
Why should the BIOS developers provide us with a wrong AccessSize field?

On the other hand, we may not need to add ROUND_UP(BitOffset + BitWidth) line.
Because there may not be such a case.
As a "region" GAS, AccessSize should always be non-zero.
And BitOffset + BitWidth should be less than AccessSize.
Otherwise how can this GAS be a region descriptor?
So BIOS developer should have the duty to provide qualified "region" GAS to us.

That's why the previous commit.
However this bug forces me to change the code to this approach:
https://bugzilla.kernel.org/show_bug.cgi?id=151501

I'm still doubt if this approach is correct as I cannot reproduce the issue.
So we'll escalate this to ASWG to see if this is needed.

Before clarification, do not merge.

source/components/hardware/hwregs.c
```diff
@@ -164,31 +164,56 @@ AcpiHwGetAccessBitWidth (
...
+    }
+    else if (Reg->AccessWidth)
+    {
+        AccessBitWidth = (1 << (Reg->AccessWidth + 2));
     }
     else
     {
-        return (1 << (Reg->AccessWidth + 2));
+        AccessBitWidth = ACPI_ROUND_UP_POWER_OF_TWO_8 (
+            Reg->BitOffset + Reg->BitWidth);
```
zetalog: If proven by the spec, we may just put an error message here.

source/components/hardware/hwregs.c
```diff
@@ -164,31 +164,56 @@ AcpiHwGetAccessBitWidth (
...
+
+    /*
+     * Return access width according to the requested maximum access bit width,
+     * as the caller should know the format of the register and may enforce
+     * a 32-bit accesses.
+     */
+    if (AccessBitWidth < MaxBitWidth)
+    {
+        return (AccessBitWidth);
     }
```
zetalog: We could put a warning message here for AccessBitWidth > MaxBitWidth case.
In order to catch unknown cases.

zetalog:
Removed support to mask bits not valid according to BitWidth/BitOffset.
Such masking should be done by the callers as we do not have idea on how we should do the masking.
There are W1C, W0C variations in registers.

zetalog:
Tested with Linux, suspend to freeze and ec works well.

PR #170, opened 2016-08-26
Acpica mlc

zetalog:
The mlc/lock fix in the order that the mlc one (not functional enabled) goes first.
So that when the coverage is not good in the lock fix, mlc needn't be reverted when the lock fix is reverted.

Thanks
Lv

### DIFF
--- a/source/compiler/aslstubs.c
+++ b/source/compiler/aslstubs.c
@@ -119,6 +119,7 @@
 #include "acevents.h"
 #include "acinterp.h"
 #include "acnamesp.h"
+#include "acparser.h"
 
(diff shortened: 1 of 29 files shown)


zetalog:
Tested with linux, both "AcpiGbl_GroupModuleLevelCode = FALSE" and "AcpiGbl_ParseTableAsTermList = TRUE" worked.
Interesting thing is:
MS doesn't support Load/LoadTable opcode in MLC, while ACPICA now allows.
This sounds like that ACPICA is better than MS when dealing with table locks.

Thanks

PR #171, opened 2016-08-30
Efi mem and I/O access services

jmarcelof:
Implementation of OS services to access physical memory and I/O on uefi environment.

Memory access is made directly (no virtual address translation needed) and ports are considered memory-mapped.

Marcelo

### DIFF
--- a/source/include/platform/acefi.h
+++ b/source/include/platform/acefi.h
@@ -319,6 +319,7 @@ struct _ACPI_EFI_FILE_HANDLE;
 struct _ACPI_EFI_BOOT_SERVICES;
 struct _ACPI_EFI_SYSTEM_TABLE;
 struct _ACPI_EFI_PCI_IO;
+struct _ACPI_EFI_PCI_ROOT_BRIDGE_IO;
 
 extern struct _ACPI_EFI_SYSTEM_TABLE        *ST;
 extern struct _ACPI_EFI_BOOT_SERVICES       *BS;
--- a/source/include/platform/acefiex.h
+++ b/source/include/platform/acefiex.h
@@ -968,6 +968,61 @@ typedef struct _ACPI_EFI_PCI_IO {
     VOID                                *RomImage;
 } ACPI_EFI_PCI_IO;
 
+/*
+ * EFI PCI Root Bridge I/O Protocol
+ */
+#define ACPI_EFI_PCI_ROOT_BRIDGE_IO_PROTOCOL \
+  { 0x2f707ebb, 0x4a1a, 0x11d4, {0x9a, 0x38, 0x00, 0x90, 0x27, 0x3f, 0xc1, 0x4d} }
+
+typedef enum {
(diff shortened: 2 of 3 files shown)


zetalog:
The 2nd patch is not correct.
AcpiOsRead/WritePort should only be used to access IO ports (aka, SystemIo address space operation regions).

Thanks

jmarcelof:
Now I understand what you mean, I saw the address range at hwvalid.c. Then I'll further analyse this issue and see what we can do about.

Thanks

jmarcelof:
The 2nd patch is fixed now. I made some primitive tests accessing ports of the RTC region and it seems to be correct.

zetalog:
One thing I do not understand for UEFI is:
It seems UEFI only provides IO functionalities related to the EFI protocols.
Either BRIDGE_IO, PCI_IO or DEVICE_IO, etc.

Which makes IO read/write a device specific manner.

Shouldn't there be a device independent way to achieve IO operations?

jmarcelof:
Actually, the EFI_PCI_ROOT_BRIDGE_IO_PROTOCOL works, for our purposes, as a wrapper to the EFI_CPU_IO_PROTOCOL which has the 'in/out instructions'. Others "specific-device-related EFI protocols" employ EFI_PCI_ROOT_BRIDGE_IO_PROTOCOL to implement their I/O.

So, the independent way of achieving IO operations is primarily by using EFI_CPU_IO_PROTOCOL, but as ACPICA will run after EFI_PCI_ROOT_BRIDGE_IO_PROTOCOL is ready it's better to use it.

Thanks

zetalog:
However ACPI allows SystemIo opregion accesses to be performed on device IO regions other than those owned by the root bridge.
And Linux implements it with in/out instructions:
https://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/tree/drivers/acpi/osl.c?id=2c85025c#n632

zetalog:
I doubt that if AML code enables a PCI device, and accesses its SystemIo operation region in an _INI function. Can EFI_PCI_ROOT_BRIDGE_IO_PROTOCOL handle it?

Thanks
Lv

jmarcelof:
I'm pretty sure it'll work fine but I can replace the RootBridgeIo by the CpuIo protocol if you think it's better.

But I'm no more so sure about the first patch (the physical memory access one). Maybe we'll need to use RootBridgeIo (or CpuIo) too.
Keep me updated about your tests and I'll fix it.

And if there's anything else that I could help just let me know.
Thanks

acpibob:
does this need to be updated, or is it obsoleted by other EFI changes?

zetalog:
The PR didn't pass the test.
However you can close it.
I've collected jmarcelof's code in my local repo with his authorship kept.
I'll take care of helping him to upstream the changes.

acpibob:
Closing as per Lv's comment

PR #172, opened 2016-09-07
Acpica 1608

zetalog:
Divergences generated for 20160831 release.


acpibob:
I went ahead and manually fixed this and committed it.

zetalog:
However I sent the release series before your modification.
It now becomes a divergence.
Let me prepare another pull request for reducing the divergence.

Thanks in advance.

PR #175, opened 2016-09-14
Acpica locks

zetalog:
Additional lock fix code.
Containing an urgent fix for my mistake.

PR #177, opened 2016-09-23
Aslts table

zetalog:
ASLTS 'table' case fixes.
Split from the origin pull request #175.


zetalog:
Hi, Bob

I don't know if MSFT will also execute the below _REG for OPR0.
OperationRegion(OPR0)
Method(_REG) {}
I just made the TLD0.tst9 working without changing the test case.

Care also be taken to make this working for all MLC modes:
1. objectlist, group=true
2. objectlist, group=false
3. termlist

Please check.

source/components/dispatcher/dswload2.c
```diff
@@ -698,10 +698,14 @@ AcpiDsLoad2EndOp (
...
-                AcpiNsGetAttachedObject (Node), FALSE);
-            AcpiExEnterInterpreter ();
+            if (WalkState->PassNumber == ACPI_IMODE_EXECUTE &&
+                !(WalkState->ParseFlags & ACPI_PARSE_MODULE_LEVEL))
+            {
+                AcpiExExitInterpreter ();
+                Status = AcpiEvInitializeRegion (
+                    AcpiNsGetAttachedObject (Node));
+                AcpiExEnterInterpreter ();
+            }
```
zetalog: This is conflict against MLC support.
SystemMemory will become inaccessible during table loading.
I'm considering to fix this by splitting the _REG execution to a new position like UpdateGpes().

zetalog:
I found previous TLD0.tst9 fix is conflict against our MLC support.
In order to support both early region accesses in MLC and auto _REG.
It seems we have to split _REG code a bit more to a new position after loading the table.
Probably same position as UpdateGpes().

This update also adds a new case to do a test around recursive "Load".

PR #180, opened 2016-10-14
avoid resource leak, check for alloc failure

coypoop:
Plus a semantic change renaming ErrorExit.
I've only smoke-tested it by running this code on my laptop.


acpibob:
Still on our list, TBD. Thanks.

acpibob:
got it, thanks for the help.

PR #185, opened 2016-10-28
Acpica loop

zetalog:
Improve infinite loop detection mechanism by adding a deterministic timeout.
This could make it more robust/useful on recent machines.


zetalog:
Moving into #196

PR #186, opened 2016-11-06
acpixtract: support CRLF in input files

Lekensteyn:
Users copying the output of acpidump may accidentally replace LF by
CRLF which confused acpixtract, resulting in only one line.

That happens because `\r\n` is not considered an empty line. Then
AxListTables would continue trying to read a whole header (`\r\nSSDT @
0x...`) and discover that `\r\nSS` is not a valid signature and skip to
the next line. At that point however, the header is partially read and
cannot be matched anymore.

Signed-off-by: Peter Wu <peter@lekensteyn.nl>

### DIFF
--- a/source/tools/acpixtract/acpixtract.c
+++ b/source/tools/acpixtract/acpixtract.c
@@ -159,7 +159,7 @@ AxExtractTables (
 
     /* Open input in text mode, output is in binary mode */
 
-    InputFile = fopen (InputPathname, "rt");
+    InputFile = fopen (InputPathname, "r");
     if (!InputFile)
     {
         printf ("Could not open input file %s\n", InputPathname);
@@ -358,7 +358,7 @@ AxExtractToMultiAmlFile (
 
     /* Open the input file in text mode */
 
-    InputFile = fopen (InputPathname, "rt");
+    InputFile = fopen (InputPathname, "r");
     if (!InputFile)
     {
         printf ("Could not open input file %s\n", InputPathname);
@@ -489,7 +489,7 @@ AxListTables (
 
     /* Open input in text mode, output is in binary mode */
 
-    InputFile = fopen (InputPathname, "rt");
+    InputFile = fopen (InputPathname, "r");
     if (!InputFile)
     {
         printf ("Could not open input file %s\n", InputPathname);
--- a/source/tools/acpixtract/axutils.c
+++ b/source/tools/acpixtract/axutils.c
@@ -334,7 +334,7 @@ AxCountTableInstances (
     unsigned int            Instances = 0;
 
 
-    InputFile = fopen (InputPathname, "rt");
+    InputFile = fopen (InputPathname, "r");
     if (!InputFile)
     {
         printf ("Could not open input file %s\n", InputPathname);
--- a/source/tools/acpixtract/axutils.c
+++ b/source/tools/acpixtract/axutils.c
@@ -172,9 +172,9 @@ AxIsEmptyLine (
         Buffer++;
     }
 
-    /* If end-of-line, this line is empty */
+    /* Line is empty when a Unix or DOS-style line terminator is found. */
 
-    if (*Buffer == '\n')
+    if ((*Buffer == '\r') || (*Buffer == '\n'))
     {
         return (1);
     }


zetalog:
I'm not sure if this is useful, because:
1. acpidump result shouldn't be messed up;
2. if something is wrong with acpidump output, you can correct it via dos2unix.

Lekensteyn:
It is useful, countering your arguments:

 1. When the acpidump textual output is copied to a website (pastebin), it might be messed up (by the website, or because the user pasted it on Windows).
 2. Currently acpixtract gives no hint/warning and just produces unexpected results. For robustness, it should handle this realistic case. dos2unix is a workaround which should be unnecessary.

zetalog:
You can consider acpidump result as a kind of binary format and then never modify it, in which case, it won't be messed up by the Windows users or pastbin.

Lekensteyn:
Counter-example: using acpidump iasl-win-20160930.zip from https://acpica.org/downloads/binary-tools on Windows 7 produces a file with CRLF. `acpixtract.exe -l acpidump-win.txt` gives all tables, the same file on Linux gives only one file.

zetalog:
Yes.
Windows acpidump will output such text format via -o or console redirection.

How is it handled by Windows acpixtract and why cannot it be handled by the Linux acpixtract?

zetalog:
By reviewing, I can see:
1. fopen() in AxCountTableInstances() is not correct.
2. You may also need to validate and improve the code in AxIsDataBlockHeader()/AxProcessOneTextLine()/AxGetTableHeader(), they are all invoked after reading an "\r\n" line as AxIsEmptyLine().

Lekensteyn:
Why is fopen AxCountTableInstances incorrect? While `t` in fopen is non-standard, the behavior should be correct, although it should not hurt to remove `t` or even change it to `b`. There are also some callers with mode `"rt"` in acpixtract.c, would you like these to be updated as well?

Note that `acpidump/apmain.c` deliberately opens files in text mode when the `-o` mode is in use (`ApOpenOutputFile`). Are you sure that acpidump output should be treated as binary?

Please correct me if I'm wrong, but I don't see anything that should be changed in AxCountTableInstances/AxIsDataBlockHeader/AxProcessOneTextLine wrt AxIsEmptyLine. AxIsEmptyLine() returns true when the input contains a CR or LF at the beginning of the line.

zetalog:
> Why is fopen AxCountTableInstances incorrect?
It's better not to fopen it again, but just pass the FILE * and rewind it.
Can be put into a different cleanup.

> Are you sure that acpidump output should be treated as binary?
I'm not sure now.
If console redirection is a common use case (i.e., acpidump > acpidump.txt), the text mode is enforced.

> I don't see anything that should be changed in AxCountTableInstances/AxIsDataBlockHeader/AxProcessOneTextLine wrt AxIsEmptyLine.
You can confirm that and write down justifications in the patch description for the reviewers.
For the reviewers, they only need to:
1. check if your description is persuasive;
2. check if all functions invoked after fgets() are mentioned in your patch description.

Lekensteyn:
I have seen your verbose explanations on the ACPI patches in Linux, they are quite well explained :-)

I have just force-pushed some patches with longer explanations. Can you check if they are OK to you?

Btw, the acpixtract list code has some strange logic. It uses AxGetTableHeader to read *at most* three lines. It aborts earlier when a line does not contain data. One possible trace is:
 - `while (fgets(...))` reads signature
 - `AxGetTableHeader` reads three data lines (containing header)
 - `while (fgets(...))` reads one line (let's assume data line)
 - line starts with space, so skipped.
 - `while (fgets(...))` reads one line (let's assume empty line)
 - line is empty, so skipped.
 - `while (fgets(...))` reads one line (let's assume signature).
 - `AxGetTableHeader` reads three data lines (containing header)
 - and so on.

After writing this out it seems working, but it was not obvious at first instance (`fgets` input is discarded).

zetalog:
> I have seen your verbose explanations on the ACPI patches in Linux, they are quite well explained.

I'm still struggling in writing patch descriptions...

> I have just force-pushed some patches with longer explanations. Can you check if they are OK to you?

Much better now. I won't be confused this time.
Your solution looks to be:
As all affected functions allow empty lines between valid lines, by treating "\r" as empty lines, we can handle this issue now.

> Btw, the acpixtract list code has some strange logic.

It means acpixtract allows empty line between data lines. So your commit will also work for it. Rigth?

Besides:
from "rt" to "t"
Should this be from "rt" to "r"?
I'm not sure if this change is necessary, is there any issue if we do not change?

Thanks

Lekensteyn:
> As all affected functions allow empty lines between valid lines, by treating "\r" as empty lines, we can handle this issue now.

Yes, that is the summary of it :smile:

> It means acpixtract allows empty line between data lines. So your commit will also work for it. Rigth?

Yes, it will (still) work for empty lines since these are skipped. (I just found the code to be confusing at the first time.)

>  Should this be from "rt" to "r"?
>  I'm not sure if this change is necessary, is there any issue if we do not change?

Right, I have updated the commit message to fix this typo and also added a note that all other callers in ACPICA do not use fopen with `t`. This change is strictly not necessary, but it does bring more consistency and avoids use of non-standard functionality.

Lekensteyn:
Should I make any other changes or can you merge (cherry-pick) the changes?

zetalog:
Maybe not further changes are needed.
I don't have write access to the acpica upstream repo.
I'll help to sync with Bob during a bi-weekly internal meeting.

acpibob:
I'm not 100% convinced about this. On Linux/Unix, there are lots of tools that cannot handle CR/LF line terminator.

What about the reverse case, where CR/LF is replaced by LF?

>>>Windows 7 produces a file with CRLF.  acpixtract.exe -l acpidump-win.txt  gives all tables, the same file on Linux gives only one file.

Isn't this the real problem?

acpibob:
>>>Windows 7 produces a file with CRLF. acpixtract.exe -l acpidump-win.txt gives all tables, the same file on Linux gives only one file.

You mean that once this .txt file is transferred to Linux, it doesn't work? I just want to get this clear for me.

Then, you want a portable .txt file across all operating systems, which is not currently supported.

Lekensteyn:
> On Linux/Unix, there are lots of tools that cannot handle CR/LF line terminator.

Similarly, there are lots of tools that can handle CR/LF.

> What about the reverse case, where CR/LF is replaced by LF?

Windows can read acpidumps from Linux (with just LF instead of CRLF), just verified this (again).
Linux cannot read acpidumps from Windows in the CRLF format (without this patch).

The original issue was someone sending me an acpidump from Linux, but then it got mangled because it was pasted into a textarea on a website. Text is text and both LF and CRLF should be treated as line terminator.

acpibob:
Ok, I don't have any problem with this.

Lekensteyn:
> Ok, I don't have any problem with this.

Does that mean it is ready for merging?

acpibob:
Works well here, both on Linux and windows.
Couple of items before merge:

1) please fully parenthesize this statement:
if ((*Buffer == '\r') || (*Buffer == '\n'))

2) Comments.
change all of these
/* Open input in text mode, output is in binary mode */
to something like:
/* Open both input and output files (binary mode) */

for this one:
    /* If end-of-line, this line is empty */

    if ((*Buffer == '\r') || (*Buffer == '\n'))

something like:
    /* Handle end-of-line for both unix and dos-style terminators */

    if ((*Buffer == '\r') || (*Buffer == '\n'))

acpibob:
Sorry, these are OK
/* Open input in text mode, output is in binary mode /

Lekensteyn:
Added parentheses and changed the preceding comment.

acpibob:
got it, thanks for your help. We're doing an acpica release this week, so this just made it in.
bob

PR #192, opened 2016-11-30
Linux-specific header: Add support for s390x compilation.

ColinIanKing:
Adds s390x as a 64-bit architecture.


acpibob:
Got it, thanks.

PR #202, opened 2017-01-23
Acpica reg

zetalog:
BZ 1339.

_REG seems to contain many issues.
But changes it suddenly may not be wise.
This is a validated first step.


timofonic:
@zetalog @acpibob @SchmErik Any news about this? Will it get merged someday?

acpibob:
Closing for now, pending additional design and code review. The _REG support is very sensitive code. Also, need to validate Window behavior before we even think about integrating this code.

PR #209, opened 2017-02-23
Acpica 64bit

zetalog:
BZ 1287.
Contains 64-bit support of internal GAS.
External API then can invoke internal ones to reduce code redundancy.


acpibob:
what prompted this change? was it requested or needed by a user?

acpibob:
Is this bugzilla 1287?

zetalog:
Yes, this is the BZ 1287 related material.

zetalog:
> what prompted this change? was it requested or needed by a user?
In theory, AcpiRead/AcpiWrite also need to support GAS.
Instead of enhancing the external APIs, this fix enhances internal APIs and let the external ones invoke the enhanced internal APIs.

PR #219, opened 2017-03-01
Makefile: enable compilation of ACPICA tools as 32-bit programs

SchmErik:
This is to facilitate compiling ACPICA tools as 32-bit programs on 64-bit machines.

To build as 32 bit programs, set the M32 variable to true when running make

Example: make M32=TRUE

Note: if the above results in a compilation error, there is a possibility
that 32-bit libraries may not be installed. If this is the case, run the
following command (on Ubuntu compiling with gcc)

    sudo apt-get install g++-multilib libc6-dev-i386

and run make again.

### DIFF
--- a/generate/unix/Makefile.config
+++ b/generate/unix/Makefile.config
@@ -148,6 +148,15 @@ ACPICA_HEADERS = \
 OPT_CFLAGS ?= $(CWARNINGFLAGS)
 
 #
+# Common compiler flags
+# The _GNU_SOURCE symbol is required for many hosts.
+#
+ifeq ($(M32),TRUE)
+CFLAGS +=-m32
+LDFLAGS +=-m32
+endif
+
+#
 # Optionally disable optimizations. Optimization causes problems on
 # some compilers such as gcc 4.4
 #


generate/unix/Makefile.config
```diff
@@ -148,6 +148,15 @@ ACPICA_HEADERS = \
 OPT_CFLAGS ?= $(CWARNINGFLAGS)
 
 #
+# Common compiler flags
+# The _GNU_SOURCE symbol is required for many hosts.
+#
+ifeq ($(M32),TRUE)
+CFLAGS +=-m32
+LDFLAGS +=-m32
+endif
```
zetalog: Normally, LDFLAGS can derive from CFLAGS on most compilers.
I'm not sure if ACPICA Makefile is written in this way.
If so, you only need to add CLFAGS here.

Just a reminder.

SchmErik: I tried calling removing LDFLAGS and calling make M32=TRUE. It turns out that the ACPICA Makefiles do not derive LDFLAGS from CFLAGS. Removing LDFLAGS results in an error when calling make.

PR #238, opened 2017-03-21
ASLTS: edit err function to print line numbers

SchmErik:


zetalog:
Why not just convert error index into __LINE__?
So that you needn't change the err function (or only need to change its verbosity).

And in the future we can introduce a macro to replace err invokers to put __LINE__ automatically.

zetalog:
err functions contain 2 very dirty parameters:
test index and error index, they actually can be converted into __FILE__/__LINE__.

acpibob:
Yes, the goal is to use FILE/LINE, of course.

This first batch of changes introduces LINE instead of the index. Change to _ERR is small.

tests/aslts/src/runtime/cntl/common.asl
```diff
@@ -795,13 +813,14 @@ Method(err, 7)
 
 /*
  * Report parameters of error
- * arg0 - absolute index of file reporting the error
+ * arg0 - absolute index of file reporting the error and line number
  * arg1 - index of error
  * arg2 - absolute index of file initiating the checking
  * arg3 - name of Method initiating the checking
  * arg4 - index of checking
+ * arg5 - line number of error
```
zetalog: I mean - we can directly convert arg4 or arg1 into __LINE__, there is no need to introduce arg5 here.

tests/aslts/src/runtime/cntl/common.asl
```diff
@@ -795,13 +813,14 @@ Method(err, 7)
 
 /*
  * Report parameters of error
- * arg0 - absolute index of file reporting the error
+ * arg0 - absolute index of file reporting the error and line number
  * arg1 - index of error
  * arg2 - absolute index of file initiating the checking
  * arg3 - name of Method initiating the checking
```
zetalog: If arg0 contains __FILE__, arg2/arg3 are also useless.

acpibob:
I think the idea here is to do _LINE_ first, then do _FILE_ in the next batch of changes

acpibob:
Yes, the idea is to get rid of the “index” parameters.

PR #241, opened 2017-03-23
Acpica gcc

zetalog:
Facilities to allow excluding gcc's stdarg.h.
Tested in efi builds by adding -DACPI_USE_BUILTIN_STDARG in Makefiles.

### DIFF
--- a/source/include/platform/acgcc.h
+++ b/source/include/platform/acgcc.h
@@ -156,7 +156,17 @@
  * Use compiler specific <stdarg.h> is a good practice for even when
  * -nostdinc is specified (i.e., ACPI_USE_STANDARD_HEADERS undefined.
  */
+#ifndef va_arg
+#ifdef ACPI_USE_BUILTIN_STDARG
+typedef __builtin_va_list       va_list;
+#define va_start(v, l)          __builtin_va_start(v, l)
+#define va_end(v)               __builtin_va_end(v)
+#define va_arg(v, l)            __builtin_va_arg(v, l)
+#define va_copy(d, s)           __builtin_va_copy(d, s)
+#else
 #include <stdarg.h>
+#endif
+#endif
 
 #define ACPI_INLINE             __inline__
 
--- a/source/include/platform/acintel.h
+++ b/source/include/platform/acintel.h
@@ -156,7 +156,9 @@
  * Use compiler specific <stdarg.h> is a good practice for even when
  * -nostdinc is specified (i.e., ACPI_USE_STANDARD_HEADERS undefined.
  */
+#ifndef va_arg
 #include <stdarg.h>
+#endif
 
 /* Configuration specific to Intel 64-bit C compiler */
 


acpibob:
I think I would like to do this in a more compiler-independent way, so that the original va_arg macros are configurable regardless of the compiler.

zetalog:
No, the original va-arg macro is not working for gcc.
Here is an example in GNU-EFI.
https://github.com/vathpela/gnu-efi/blob/master/inc/efistdarg.h

And attached is a photo I took by running an acpidump.efi whose va_arg macros are implemented by the original ACPICA default ones - so they are MSVC dependent.
We can see many other open source projects containing same macros, but they are only use them for MSVC va_arg stubs.
img_6095 1

zetalog:
The built EFI binary cannot run, it crashes, dumping unreadable contents.
Looks like it dumps things from mystery memory chunks (looks like firmware mapped).
And the program cannot run into an end.

So the va_arg implementation is compiler specific.
And we don't have a position in headers can allow we to put such default compiler stuffs without messing up the inclusion order.
It seems the actual order should be:

include <host configurable that can affect compilers>
include <compiler specific pre-requisites>
<possibly could put it here>
include <host configurable that can implement host specific compiler behavior>
include <compiler specific host-affected implementation>
include <host specific implementation>

But it will do no good but allow some build test to pass.
What if we can provide compiler specific correct implementations to let it not just pass the build test but also pass the runtime test?

zetalog:
Maybe the "acpidump -h" result is more persuasive.

zetalog:
All va-arg result are crashed:

img_6099

zetalog:
Why originally solaris build works?
Actually originally we don't have the following lines in acgcc.h.
```
#include <stdarg.h>
```

And in acenv.h, original va_arg implementation is wrapped by a check:
```
#ifndef va_arg
...
#endif
```

So for solaris, it actually only need to exclude <stdarg.h>.

Why we need solaris to specify -DACPI_USE_EXTERNAL_STDARG now?

As both BSDs/Linux kernels include <stdarg.h> (as freestanding compiler specific header), we collect the inclusion in acgcc.h for all gcc builds.
Otherwise each ACPICA user should do include stdarg.h prior than including acpi.h.
Then we can:
```
#ifndef va_arg
#include <stdarg.h>
#endif
```
In this way, Solaris makefiles do not need to be changed.
However we can hardly ask all BSDs/Linux (and more gcc built hosts) to change.
So we should:
```
#ifndef ACPI_USE_EXTERNAL_STDARG
#include <stdarg.h>
#endif
```
And Solaris could be the only one that must change by adding -DACPI_USE_EXTERNAL_STDARG to their makefiles.
Also we can see strict order of including stdarg.h for gcc, it must appear in acgcc.h, cannot be moved to acgccex.h.
As a conclusion, this is a required change to Solaris to use cleaned ACPICA compiler headers and the most cleaning/right-ordered way for all other OSPMs.

zetalog:
Why we need solaris to specify -DACPI_USE_EXTERNAL_STDARG now?

As both BSDs/Linux kernels include <stdarg.h> (as freestanding compiler specific header), we collect the inclusion in acgcc.h for all gcc builds.
Otherwise each ACPICA user should do include stdarg.h prior than including acpi.h.
Then we can:
```
#ifndef va_arg
#include <stdarg.h>
#endif
```
In this way, Solaris makefiles do not need to be changed.
However we can hardly ask all BSDs/Linux (and more gcc built hosts) to change and for this case, Solaris seems to be the only special case.
So we can:
```
#ifndef ACPI_USE_EXTERNAL_STDARG
#include <stdarg.h>
#endif
```
And Solaris could be the only one that must change by adding -DACPI_USE_EXTERNAL_STDARG to their makefiles.
Also we can see strict inclusion order requirement of stdarg.h for gcc, it must appear in acgcc.h, cannot be moved to acgccex.h.

As a conclusion, this is a required change to Solaris to use new versioned ACPICA headers and the most cleaning/right-ordered way for all other OSPMs.

zetalog:
I can to utilize "#ifndef va_arg" in acgcc.h to see if it works for all hosts.

acpibob:
Is this ready for merge?

zetalog:
> Is this ready for merge?

I think yes.

acpibob:
I originally meant: create an option to use either:
1) stdarg.h or
2) our local versions of the va_arg stuff

Default would be stdarg.

In this way, the user can choose 1 or 2 regardless of what compiler is used or what environment.

zetalog:
The problem is the "local version of va_arg stuff" is harmful to other compilers.
Users of other compilers won't be willing to see them included.
We've managed to achieve the cleanup.
We shouldn't put them back to mess things up.

PR #242, opened 2017-03-23
Change %u to %lu to fix compilation error.

SchmErik:


acpibob:
instead changed the BadBytes to UINT32 from ACPI_SIZE

PR #245, opened 2017-03-29
Acpica debugger

zetalog:
https://bugzilla.kernel.org/show_bug.cgi?id=194845

### DIFF
--- a/source/components/dispatcher/dsutils.c
+++ b/source/components/dispatcher/dsutils.c
@@ -791,12 +791,6 @@ AcpiDsCreateOperand (
         if ((OpInfo->Flags & AML_HAS_RETVAL) ||
             (Arg->Common.Flags & ACPI_PARSEOP_IN_STACK))
         {
-            ACPI_DEBUG_PRINT ((ACPI_DB_DISPATCH,
-                "Argument previously created, already stacked\n"));
-
-            AcpiDbDisplayArgumentObject (
-                WalkState->Operands [WalkState->NumOperands - 1], WalkState);
-
             /*
              * Use value that was already previously returned
              * by the evaluation of this argument


acpibob:
This entire call to the debugger should be macro-ized under the ACPI_DEBUG_EXEC macro.

Also, I'm not sure if we should just check for 0 arguments at this point since this debugger call is used for single-stepping, it would seem.

zetalog:
I'm not sure what you mean here.
Linux kernel now uses these functions, so that are not AcpiExec specific.

All debugger APIs are already made as stub in a different way.
1. Most of the debugger APIs invoked by the rest of the ACPICA subsystem are made stub by ACPI_DBR_DEPENDENT_XXX macros in the header files, and source files that contain these functions won't compile if ACPI_DEBUGGER is not configured.
2. During runtime, these functions are bypassed due to some runtime check, for example, if it is single stepping mode and etc.

There might be functions haven't been cleaned up according to the above rules, but they are not related to this bug fix IMO.

zetalog:
> Also, I'm not sure if we should just check for 0 arguments at this point since this debugger call is used for single-stepping, it would seem.

My point is: AML_HAS_RETVAL now doesn't look like a useful flag to determine if a result is already in the operands arrary. It seems in whatever cases, when these lines are reached, operands array is always empty. As there are another invocation below to dump all arguments, we can ommit this invocation for AML_HAS_RETVAL.

However for ACPI_PARSEOP_IN_STACK, it really means something in the operands array, it might be OK to use NumOperands - 1 as an index to access the arrary. Even if there are bugs related to ACPI_PARSEOP_IN_STACK, it is not related to this repoted bug. We can leave it unchanged, and if it is triggered again due to ACPI_PARSEOP_IN_STACK, then we should validate the entire ACPI_PARSEOP_IN_STACK support.

So I prefer to exclude AML_HAS_RETVAL rather than adding if (> 0) check which may make ACPI_PARSEOP_IN_STACK logic implicit.

zetalog:
What the purpose of this debugging message?
%
ArgObj:    00873048 <Obj>             Integer 0000000000000000
ArgObj:    00873540 <Obj>             Integer 00000000FFFF00FF <=
ArgObj:    00873540 <Obj>             Integer 00000000FFFF00FF <= redundant previous ArgObj
ArgObj:    00888008 <Obj>             Integer 000000000000AB00
It only dumps redundant objects.
What if we just remove the entire call to DisplayArgument?

zetalog:
It looks all AcpiDsObjStackPush() invocations are paired with AcpiDbDisplayArgumentObject().
So the purpose of these lines are confusing.

acpibob:
I do in fact see NumOperands at zero. That is also why runtime UBSAN reports the -1 index. 
I'm investigating the whole thing further.

zetalog:
> I do in fact see NumOperands at zero. That is also why runtime UBSAN reports the -1 index.
 I'm investigating the whole thing further.

I can also see that (NumOperands = 0) when the crash is encountered, but all such cases are seen when AML_HAS_RETVAL set but ACPI_PARSEOP_IN_STACK unset.

acpibob:
>>>Linux kernel now uses these functions, so they are not AcpiExec specific

I mean that they still need to be configurable for other operating systems, which I hope they are. The only thing I don't quite like is that we usually use a macro with uppercase letters for this type of thing (like ACPI_DEBUG_EXEC, etc.), which immediately tells the reader that the code within the macro is configurable.

In any case, I've committed the changes, and even pulled out the debug message.

zetalog:
>> The only thing I don't quite like is that we usually use a macro with uppercase letters for this type of thing (like ACPI_DEBUG_EXEC, etc.), which immediately tells the reader that the code within the macro is configurable.

It's not a good practice in Linux. Configurable code are managed in Linux at function boundary.
It's basically now allowed to do
```
#ifdef XXX
code
#endif
```
Inside of a function. Developers who are allowed to do so will make less engineering effort to make their code more compact. We may consider a better way to survive this coding style.

PR #246, opened 2017-03-29
Explicitly cast 1 to UINT32.

SchmErik:
The runtime errors caused when acpica tools are compiled with
-fsanitize=shift imply that these 1s are stored in integers.
This cast insures that 1 is stored in unsigned integers.

### DIFF
--- a/source/components/utilities/utownerid.c
+++ b/source/components/utilities/utownerid.c
@@ -225,14 +225,14 @@ AcpiUtAllocateOwnerId (
                 break;
             }
 
-            if (!(AcpiGbl_OwnerIdMask[j] & (1 << k)))
+            if (!(AcpiGbl_OwnerIdMask[j] & ((UINT32) 1 << k)))
             {
                 /*
                  * Found a free ID. The actual ID is the bit index plus one,
                  * making zero an invalid Owner ID. Save this as the last ID
                  * allocated and update the global ID mask.
                  */
-                AcpiGbl_OwnerIdMask[j] |= (1 << k);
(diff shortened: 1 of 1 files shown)


zetalog:
This looks scary, the compiler warning is actually meaningful.
It's undefined compiler behaviors if a shift operation wrapped over.
See ACPI_MASK_BITS_ABOVE/BELOW_64.
You may need to check if the change can cause some cases failure at the boundary of the data types.

SchmErik:
The runtime warnings are as follows:

../../../source/components/utilities/utownerid.c:228:47: runtime error: left shift of 1 by 31 places cannot be represented in type 'int'
../../../source/components/utilities/utownerid.c:235:46: runtime error: left shift of 1 by 31 places cannot be represented in type 'int'
../../../source/components/utilities/utownerid.c:331:13: runtime error: left shift of 1 by 31 places cannot be represented in type 'int'

Which alludes to the fact that this result is being stored in an int only when compiled using -fsanitize=shift.

acpibob:
The key here is that everything is defined to be UINT32, especially the target, and the whole operation is surrounded by this loop:

        for (k = AcpiGbl_NextOwnerIdOffset; k < 32; k++)
        {
UBSAN is complaining about a shift of 31 (at runtime), which is OK. 32 and above can't happen. Since it is at runtime, the warning happens very many times.

So, I don't have any problem with simply adding a cast to make UBSAN happy. We are attempting to run UBSAN with the entire ASLTS.

zetalog:
> k < 32

That explains this fix. It's better to put comments before the casts.

PR #248, opened 2017-04-01
Optionally use designated initializers for function pointers

kees:
The Linux kernel is going to soon be using a gcc plugin for automatically
randomizing structure layouts of structures that are entirely function
pointers. To support this, initializers need to use designated
initialization rather than ordered initialization. This provides a macro,
ACPI_STRUCT_INIT, to handle both cases, keeping ordered initialization
for non-Linux builds.

### DIFF
--- a/source/components/hardware/hwxfsleep.c
+++ b/source/components/hardware/hwxfsleep.c
@@ -184,9 +184,18 @@ AcpiHwSleepDispatch (
 
 static ACPI_SLEEP_FUNCTIONS         AcpiSleepDispatch[] =
 {
-    {ACPI_HW_OPTIONAL_FUNCTION (AcpiHwLegacySleep),    AcpiHwExtendedSleep},
-    {ACPI_HW_OPTIONAL_FUNCTION (AcpiHwLegacyWakePrep), AcpiHwExtendedWakePrep},
-    {ACPI_HW_OPTIONAL_FUNCTION (AcpiHwLegacyWake),     AcpiHwExtendedWake}
+    {ACPI_STRUCT_INIT (legacy_function,
+                       ACPI_HW_OPTIONAL_FUNCTION (AcpiHwLegacySleep)),
+     ACPI_STRUCT_INIT (extended_function,
+                       AcpiHwExtendedSleep) },
+    {ACPI_STRUCT_INIT (legacy_function,
+                       ACPI_HW_OPTIONAL_FUNCTION (AcpiHwLegacyWakePrep)),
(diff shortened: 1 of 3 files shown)


source/components/hardware/hwxfsleep.c
```diff
@@ -180,13 +180,29 @@ AcpiHwSleepDispatch (
 #define ACPI_WAKE_PREP_FUNCTION_ID      1
 #define ACPI_WAKE_FUNCTION_ID           2
 
+/*
+ * Some compilers can handle designated initializers, which is needed
+ * under Linux kernel builds for structures that are entirely function
+ * pointers.
+ */
+#ifdef __GNUC__
```
zetalog: Such ifdef should only appear in acenv.h.

source/components/hardware/hwxfsleep.c
```diff
@@ -180,13 +180,29 @@ AcpiHwSleepDispatch (
...
+#define ACPI_SLEEP_FUNCTION_INIT(legacy, extended) \
+              { .legacy_function   = legacy,     \
+                .extended_function = extended }
```
zetalog: Is that possible to introduce a mechanism rather than a special case?

zetalog: Is that possible to introduce a mechanism rather than a special case?
For example:
```
#define ACPI_DEFINE_FUNCITON(name, pointer) \
...
#define ACPI_SLEEP_FUNCTION_INIT(legacy, extended) \
{ \
    ACPI_DEFINE_FUNCITON(legacy_function, legacy), \
    ACPI_DEFINE_FUNCITON(extended_function, extended) \
}
```

zetalog: Anyway, I'm not sure what the benefit of doing this.
Even it is possible, we still cannot change the function definition order inside of ACPICA.
As it can break other compilers.

kees: I'll try your approach to making a general purpose initializer, thanks!

As to the definition order: this shouldn't matter, right? The structures is not shared outside of the kernel under Linux, yes?

kees:
Friendly ping. What do you think of the revised patch? Thanks!

SchmErik:
Looks good! Just one thing: ACPICA is a OS-independent project that needs to be compiled using many different compilers and its meant to be consumed by linux as well as other operating systems. Could you generalize your commit message to avoid using keywords like GCC and Linux?

kees:
Okay, I've removed mention of that in the changelog.

PR #251, opened 2017-04-07
External op refactor02

SchmErik:
This pull request changes the compiler behavior when there are externals and named objects with the same name. If these externals and named objects appear in different definition blocks, the externals are references to the named object. This case is considered a valid use case and passes compilation. If these externals and named objects appear in the same definition block, this is considered a redefinition of that named object or external and results in an error. With this new change in behavior, the external(_ERR, MethodObj) declaration in ehandle.asl within ASLTS has to be deleted to avoid redefinition of _ERR.

### DIFF
--- a/source/compiler/aslload.c
+++ b/source/compiler/aslload.c
@@ -797,23 +797,30 @@ LdNamespace1Begin (
             {
                 /*
                  * Allow one create on an object or segment that was
-                 * previously declared External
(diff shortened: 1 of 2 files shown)


SchmErik:
Corner case has been implemented in [https://github.com/acpica/acpica/pull/251/commits/7ef9b537098a4f9fabfc87cd6667256140f6e993](url)

PR #253, opened 2017-04-14
Fix typo

SchmErik:

### DIFF
--- a/source/components/tables/tbfadt.c
+++ b/source/components/tables/tbfadt.c
@@ -597,8 +597,8 @@ AcpiTbCreateLocalFadt (
  * The 64-bit X fields are optional extensions to the original 32-bit FADT
  * V1.0 fields. Even if they are present in the FADT, they are optional and
  * are unused if the BIOS sets them to zero. Therefore, we must copy/expand
- * 32-bit V1.0 fields to the 64-bit X fields if the the 64-bit X field is
- * originally zero.
+ * 32-bit V1.0 fields to the 64-bit X fields if the 64-bit X field is originally
+ * zero.
  *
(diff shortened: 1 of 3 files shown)


source/components/tables/tbutils.c
```diff
@@ -257,7 +257,7 @@ AcpiTbCheckDsdtHeader (
  *
  * FUNCTION:    AcpiTbCopyDsdt
  *
- * PARAMETERS:  TableDesc           - Installed table to copy
+ * PARAMETERS:  TableIndex         - Index of installed table to copy
```
zetalog:
```
- Index of ...
```
Seems not correct.
Has it been aligned to the parameter description column?
It should be a column number of multiple-4.

Normally I'll duplicate this line and change 1 of them so that I can align it correctly.

zetalog: Also I can see problems in the patch description:
1. Should the body be empty? It's OK of being empty in ACPICA upstream, but not OK for linux upstream IMO.
2. There shouldn't be empty line between SOBs.

source/components/tables/tbfadt.c
```diff
@@ -597,7 +597,7 @@ AcpiTbCreateLocalFadt (
  * The 64-bit X fields are optional extensions to the original 32-bit FADT
  * V1.0 fields. Even if they are present in the FADT, they are optional and
  * are unused if the BIOS sets them to zero. Therefore, we must copy/expand
- * 32-bit V1.0 fields to the 64-bit X fields if the the 64-bit X field is
+ * 32-bit V1.0 fields to the 64-bit X fields if the 64-bit X field is
```
zetalog: Maybe the answer is yes, but you need to check if the next line should still be wrapped.

PR #262, opened 2017-05-10
skip building tools if they exist

michealgong:


acpibob:
We want to force a build of the tools to absolutely guarantee that the tests are run with the latest version. It is a relatively small overhead.

acpibob:
Also, we are in the process of replacing make with python.

zetalog:
To Michael:
Try "aslts.sh -u"
For more options, try aslts.sh -h.


To Bob:
> We want to force a build of the tools to absolutely guarantee that the tests are run with the latest version. It is a relatively small overhead.

Yes, there is a case.
When developers are correcting aslts and the C source code needn't be changed.
They'll run aslts.sh many times, and surely don't want the tools to be rebuilt for each run.

zetalog:
Anyway, this commit is useless, please drop it.
We can use "aslts.sh -u".

PR #265, opened 2017-05-15
Acpica tables2

zetalog:
Support deferred table verification, useful for Linux.


zetalog:
Close as merged to #121 per upstream request.

PR #270, opened 2017-06-09
Acpica aslts

zetalog:
Some ASLTS case is not written in a Windows compliant way.
This patchset fixes such problems.
By converting in-package NameString into the following style:

Store(RefOf(name), Index(pkg, ...))

And find correct position to add DerefOf to extract the element.

### DIFF
--- a/tests/aslts.sh
+++ b/tests/aslts.sh
@@ -116,15 +116,13 @@ run_aslts() {
 	version=`$ASL | grep version | awk '{print $5}'`
 	rm -rf $ASLTSDIR/tmp/aml/$version
 
-	if [ "x$TEST_CASES" = "x" ]; then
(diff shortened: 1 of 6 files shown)


zetalog:
Hi, guys

I think we are not synchronized well.

Let me add one more comment.
Applying this pull request won't change ASLTS result.
I just confirmed this by:
1. run aslts.sh to generate first RESULT folder, say <new>;
2. apply this series, and run aslts.sh to generate second RESULT folder, say old;
3. comparing new/opt|nopt/32|64/__STATUS_OF_TEST_CASES with old/opt|nopt/32|64/__STATUS_OF_TEST_CASES generates empty differences

So this series is a no-op.

zetalog:
Lost track of this branch...

PR #274, opened 2017-06-15
Acpica loop

zetalog:
Current loop implementation.

### DIFF
--- a/source/components/dispatcher/dscontrol.c
+++ b/source/components/dispatcher/dscontrol.c
@@ -234,7 +234,8 @@ AcpiDsExecBeginControlOp (
             WalkState->ParserState.PkgEnd;
         ControlState->Control.Opcode =
             Op->Common.AmlOpcode;
-
+        ControlState->Control.LoopTimeout = AcpiOsGetTimer () +
+           (UINT64) (AcpiGbl_MaxLoopIterations * ACPI_100NSEC_PER_SEC);
 
         /* Push the control state on this walk's control stack */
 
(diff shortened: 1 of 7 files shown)


zetalog:
I just have a concern with this solution and Len/Rafael didn't answer me.

I feel we need to apply timeout mechanism as an addition to the count mechanism.
Because if we use timeout mechanism only, it may just loop 1-2 times to blame an AE_TIME while originally it have looped too many times..

acpibob:
I would like to point out one thing:

AcpiExec changes the default loop count to 0x400 (1K) for this reason:

During the loading of ACPI tables, the _INI, etc. methods are run, as per usual in ACPICA. However, there is of course no hardware support, and if there is initialization code in the DSDT/SSDTs that attempts to touch the actual hardware, the AML can very possibly (and often does) get into an infinite loop while waiting for some hardware event that will never happen.

Therefore, we want AcpiExec to timeout rather quickly on these hardware loops.

When we move to the use of AcpiOsGetTimer for the infinite loop detection, we should set the default timeout to something rather short, perhaps a second or two. Otherwise, AcpiExec could get hung up for many seconds if not minutes.

zetalog:
I've changed the default timeout to 1seconds for acpiexec.

zetalog:
Default OSPM 30seconds value:

In Linux, NMI times out 10seconds. NMI should have a shortest timeout than other mechanisms.
Network protocols commonly have 120seconds timeout.
And AML can access IPMI operation region which can have an underlying msg based network lower layer.
We actually want an INFINITE value here, but we cannot do that in case Linux runs some BIOS code that acesses Linux unsupported hardware.

From these point of views, uses 30seconds as a default value that is between the shortest and the theoretical.

PR #276, opened 2017-06-20
iASL: Add support for the SDEI table

james-morse-arm:
ACPI 6.2 adds support for the Software Delegated Exception Interface,
which is described by "Software Delegated Exception Interface (SDEI)"
ARM DEN0054A.

Add the necessary types in the ACPICA header files and support for
compiling/decompiling the table.

Signed-off-by: James Morse <james.morse@arm.com>


source/include/actbl2.h
```diff
@@ -1418,6 +1419,21 @@ typedef struct acpi_mtmr_entry
 
 } ACPI_MTMR_ENTRY;
 
+/*******************************************************************************
+ *
+ * SDEI - Software Delegated Exception Interface Descriptor Table
+ *
+ * Conforms to "Software Delegated Exception Interface (SDEI)" ARM DEN0054A,
+ * May 8th, 2017. Copyright 2017 ARM Ltd.
```
SchmErik: Just out of curiosity, why is there a copyright notice here but not in this commit?
https://github.com/acpica/acpica/pull/269/commits/d00a4eb86e64bb4fa70f57ab5e5ca0a4ca2ad8ef

james-morse-arm: Hi Erik,

I picked the SLIC table as a template, it has a date and copyright statement like this. I guess Robin picked a different example.

Thanks,

acpibob:
A couple things:

1) DtCompileSDEI: This doesn't look like this function is necessary. A NULL entry in AcpiDmTableData causes the "generic" code to be executed. Same with AcpiDmDumpSDEI.

2) For consistency, function names of this form DtCompileSDEI should be DtCompileSdei.

james-morse-arm:
Hi Robert,

This was tricky, the DTCompileDataTable() call fails with invalid-parameter because the FieldList parameter indicates there is no more work to do.
I've added a patch to allow this when compiling simple tables.

I will try and close this pull-request and open a new one with the new pre-requisite patch.

PR #277, opened 2017-06-20
Executer: adapt buffer length for Field Attrib Raw Process in Ops Region

bentiss:
Detected on the Surface 3:
The MSHW0011 driver uses Field Attrib Raw Process to return information
for the ACPI Battery. The DSDT declares a parameter of 2 though
functions like _BST or _BIX require a much bigger out buffer.

It looks like the incoming buffer has the requested size and we can
work around the issue by using this input size as the output and
parameters size.

As metioned by Stephen Just on the kernel bugzilla (comment 44):
"Lv, after reading through 5.5.2.4.5.3 of the ACPI spec several times,
it appears that AccessLength is ignored for AttribRawProcessBytes
accesses - the access length is always specified by the user - and
in fact, the example omits the AccessLength argument entirely.

More specifically, comparing 5.5.2.4.5.3.9 and 5.5.2.4.5.3.10, it
appears that the only difference between AttribRawProcessBytes and
AttribRawBytes is that AttribRawProcessBytes does not seem to pay
attention to AccessLength, and lets the driver deal with it."

Link: https://bugzilla.kernel.org/show_bug.cgi?id=106231

Signed-off-by: Benjamin Tissoires <benjamin.tissoires@redhat.com>

### DIFF
--- a/source/components/executer/exfield.c
+++ b/source/components/executer/exfield.c
@@ -553,6 +553,15 @@ AcpiExWriteDataToField (
              *     Data[x-1]: (Bytes 2-x of the arbitrary length data buffer)
              */
             Length += 2;
+
+            /*
+             * When using AttribRawProcessBytes, it seems the access length
+             * is always specified by the user.
+             * So just take the incoming buffer length as the reference.
+             */
+            if (AccessorType == AML_FIELD_ATTRIB_RAW_PROCESS)
+                Length = SourceDesc->Buffer.Length;
+
             Function = ACPI_WRITE | (AccessorType << 16);
         }
         else /* IPMI */


hadess:
Looks good from my side. :+1:

zetalog:
TBH, IMO, this fix is not correct.

The root cause of the issue is:
ACPI spec declears a call-by-reference ABI, while ACPICA uses AcpiUtCopyIobjectToIobject() breaking this (it has to do that, because the stacked objects' reference counts are already broken), making some objects passed just like call-by-value.

So this is only a workaround, in order to have the end user problem working under this circumstances.

Cheers

bentiss:
> So this is only a workaround, we can ship just in order to have the end user problem fixed under this circumstances.

So, do you want me to send a pull request amending the comment to explain what you put in the comment above?

acpibob:
Is this change going to cause a problem? I guess we will see. I certainly don't like making random workarounds, so we may just revert this.

zetalog:
> So, do you want me to send a pull request amending the comment to explain what you put in the comment above?

You needn't. :)

> Is this change going to cause a problem?

I'm not sure for now. At least this fixes an end user issue.

> I guess we will see. I certainly don't like making random workarounds, so we may just revert this.

Tuning operand objects' references looks a big deal (need a big change and many test efforts) and we should be careful as the current code at least working for most cases.
If problem occurs, we can put an option around it and prepare a kernel boot parameter for different end users.
So IMO, we can keep this change in upstream for now.

bentiss:
> I certainly don't like making random workarounds, so we may just revert this.

@acpibob, given the above explanation from @zetalog, would you agree to revert the revert, or do I need to resubmit a PR, or is your decision final?

acpibob:
We are still looking at this.

BTW, do you know what the behavior is on Windows? It would be very, very helpful to know this.
thanks,
Bob

bentiss:
> We are still looking at this.

Thanks :)

> BTW, do you know what the behavior is on Windows? It would be very, very helpful to know this.

I wish I could help, but I have no ideas on how to debug this under Windows. I know how to override the DSDT[1] under Windows, but that's the top of my knowledge. If you have specifics for me to test, feel free to ask.

[1] it's BTW linked here: https://bugzilla.kernel.org/attachment.cgi?id=187171

jwrdegoede:
Hi All,

Talking To Bastien and Benjamin about battery monitoring on the Surface 3 has made me take a look at this. It looks like Benjamin's initial patch is mostly correct and we really need something like this.

Let me explain, looking at "5.5.2.4.5.3.10 Raw Block Process Call (AttribRawProcessBytes)" (page 240) in the ACPI_6.0.pdf 2 things stand out:

1) Unlike "5.5.2.4.5.3.9 Raw Read/Write N Bytes (AttribRawBytes)" (page 238) there is no mention of "The actual number of bytes to read or write is specified as part of the AccessAs attribute." IOW the specification does NOT say we should use access_length to determine the length, the current ACPICA code does use access_length and I believe this is wrong.

2) The example code for AttribRawProcessBytes uses just "AttribRawProcessBytes" NOT "AttribRawProcessBytes(x)"

These both indicate that the access_length passed to the AccessAs-AccessAttribute should not be used for an ACPI_ADR_SPACE_GSBUS access with an accessor_type of AttribRawProcessBytes.

Further more the example code contains the following:

Name(BUFF, Buffer(34){}) // Create SerialBus buf as BUFF
CreateByteField(BUFF, 0x00, STAT) // STAT = Status (Byte)
CreateByteField(BUFF, 0x01, LEN) // LEN = Length (Byte)

Store(8, LEN) // Length of the valid data
Store(Store(BUFF, FLD0), BUFF) // Execute the PC
if (LEqual(STAT, 0x00)) // Test the status
{
// BUFF now contains information returned from PC
// LEN now equals size of data returned
}

Indicating that the input length is passing in through the LEN field of the passed in buffer and that LEN afterwards is set to the amount of data returned. Note that the buffer is an input/output buffer and we do not know how much data will be returned.

Since AttribRawProcessBytes does an atomic write + readback operation, the passed in source-buffer will also be the output buffer. So we know that the amount of data returned will be max SourceDesc->Buffer.Length bytes and the input length naturally also must be that size or less.

So the only correct way to deal with this is to allocate a buffer of that size as Benjamin's patch is doing.

The only problem I see with Benjamin's patch is that there is no way for the opregion handler to know the size of the passed in buffer. Passing this in requires an extra argument to acpi_ex_access_region and all the opregion handlers. A simpler way of passing this would be to store it in the field.access_length, which is otherwise unused for ACPI_ADR_SPACE_GSBUS accesses with an accessor_type of AttribRawProcessBytes.

So something like this (linux kernel sources diff):
```
--- a/drivers/acpi/acpica/exfield.c
+++ b/drivers/acpi/acpica/exfield.c
@@ -366,6 +366,16 @@ acpi_ex_write_data_to_field(union acpi_operand_object *source_desc,
                } else if (obj_desc->field.region_obj->region.space_id ==
                           ACPI_ADR_SPACE_GSBUS) {
                        accessor_type = obj_desc->field.attribute;
+
+                 /*
+                  * AttribRawProcessBytes uses a variable length,
+                  * stored as in/out param inside the buffer,
+                  * which may use up to the entire buffer size.
+                  */
+                 if (AccessorType == AML_FIELD_ATTRIB_RAW_PROCESS)
+                         obj_desc->field.access_length =
+                                 source_desc->buffer.length - 2;
+
                        length =
                            acpi_ex_get_serial_access_length(accessor_type,
                                                             obj_desc->field.

```

Regards,

Hans

jwrdegoede:
p.s.

As for this causing regressions, I've a collection with acpidumps from over 70 different machines and  the Surface 3 is the only one using the AttribRawProcessBytes keyword, so I don't think we need to worry about that too much, since it simply is almost never used.

jwrdegoede:
Hi again,

Upon closer reading of this pull-req I see that the original analysis matches with mine, but that the reason the fix is not liked is because the REAL fix would be to simply pass a reference to the original buffer rather then passing a copy at which point the original buffer size would automatically be retained and everything is fine.

I understand that that would be better, but we are one year later now and battery monitoring is still broken on the Surface 3 because of this.

Also the revert message is wrong, it says: "This appears to be a workaround for some bad ASL or other code. This type of code should not be in the ACPICA code."  But the bug is not in the ASL code or in code outside of ACPICA.

The bug is in ACPICA, specifically the bug is that since ACPICA for now still makes a copy of the buffer, it is using the wrong buffer size while making that copy. Benjamin originals commit fixes ACPICA using the wrong buffer size and as such it,  or my alternative which allows buffer overflow checking in the handler, should be applied until the making of the copy is removed all together.

Regards,

Hans

jwrdegoede:
Ping? This is a real issue with a known fix and no real reasons not to implement the fix,  can we at leats get some reaction as to why the fix is not being re-considered?

SchmErik:
Sorry for the late response. I'll take a look and let you know.

acpibob:
Yes, this is on our list of things to look at.

SchmErik:
We found this code

AccessAs (BufferAcc, AttribRawProcessBytes (0x02)), 
BB00,   8

Do you have any idea why the 0x02 is used?

SchmErik:
We were thinking about making this work for both cases. Can you try this?
```
From b4b8938d7163458278ece12fc9e6da6d56942681 Mon Sep 17 00:00:00 2001
From: Erik Schmauss <erik.schmauss@intel.com>
Date: Thu, 2 Aug 2018 13:06:55 -0700
From: Erik Schmauss <erik.schmauss@intel.com>
Subject: [PATCH 1] ACPICA: Executer: choose the input buffer as access length if it is larger than the original

---
 exfield.c |    6 ++++++
 1 file changed, 6 insertions(+)

diff -Nurp linux.before_name/drivers/acpi/acpica/exfield.c linux.after_name/drivers/acpi/acpica/exfield.c
--- linux.before_name/drivers/acpi/acpica/exfield.c	2018-08-02 13:09:08.691845563 -0700
+++ linux.after_name/drivers/acpi/acpica/exfield.c	2018-08-02 13:09:02.889845171 -0700
@@ -379,6 +379,12 @@ acpi_ex_write_data_to_field(union acpi_o
 			 *     Data[x-1]: (Bytes 2-x of the arbitrary length data buffer)
 			 */
 			length += 2;
+			if (accessor_type == AML_FIELD_ATTRIB_RAW_PROCESS &&
+			    source_desc->buffer.length >
+			    obj_desc->field.access_length) {
+				length = source_desc->buffer.length;
+			}
+
 			function = ACPI_WRITE | (accessor_type << 16);
 		} else {	/* IPMI */
 ```

acpibob:
Here is an issue that we have and need to figure out.

From the ACPI spec:
Access Attributes that require an AccessLength argument:
• AttribBytes (AccessLength)
• AttribRawBytes (AccessLength)
• AttribRawProcessBytes (AccessLength)

Clearly, the AccessLength for AttribRawProcessBytes is not something optional and it is not to be ignored. It is required. And the spec should also say that this is the transfer length, as the other similar Attribs (The AML grammar goes way out of its way to stuff the AccessLength into unused existing bits for these Attribs.)

As you already know, ACPICA uses the AccessLength field for the transfer length. We feel that this is in keeping with the ACPI spec, even though it is a bit difficult to figure it out.

So, we need to figure out where the real bug is.
1) The ASL itself could be wrong (length of 2), or
2) Windows has a bug or a workaround.

This puts us on a path to figure out exactly what Windows does.

We have some assumption that the Surface 3 runs windows correctly (could be wrong); so perhaps Windows does either the wrong thing, or it does something similar to Erik's patch above. Yet, we need to know for sure what is happening (and why) before we add something like this to the code base.

Bob

jwrdegoede:
Hi,

On 08/03/2018 06:18 AM, Robert Moore wrote:
> Here is an issue that we have and need to figure out.
> 
>  From the ACPI spec:
> Access Attributes that require an AccessLength argument:
> • AttribBytes (AccessLength)
> • AttribRawBytes (AccessLength)
> • AttribRawProcessBytes (AccessLength)

As both zetalog's and my later independent analysis show
this clearly is a "bug" in the spec, the text discussing
how AttribRawProcessBytes actually works shows clearly
that the second byte in the buffer gives the input /
output size of the command and the spec also says that
optionally AttribRawProcessBytes may be used without
an AccessLength argument.

Regards,

Hans

acpibob:
There are at least two lengths that come into play here.  It is our understanding that the second byte of the buffer gives the length of the "valid data" in the buffer rather than the actual length of the buffer.  This seemingly allows a large buffer with a small command header (the valid data) to be passed to the handler in order to allow for a relatively large amount of return data (_BIF is a probably a good example).

If the AccessLength is not used, then why is it in the spec/grammar in the first place?

What appears to be missing in the spec section for AttribRawProcessBytes is the text "The actual number of bytes to read or write is specified as part of the AccessAs attribute" This text is used for
AttribRawBytes and AttribBytes.

> spec also says that
> optionally AttribRawProcessBytes may be used without
> an AccessLength argument.

Where does it say this?

BTW, the ASL and AML grammars for AttribRawProcessBytes do in fact specify that the AccessLength is a required argument (AttribBytes, and AttribRawBytes as well), text extracted from the spec below:

ASL
AccessAttribKeyword :=
AttribQuick | AttribSendReceive | AttribByte | AttribBytes (n) | AttribRawBytes (n) |
AttribRawProcessBytes (n) | AttribWord | AttribBlock |AttribProcessCall |
AttribBlockProcessCall     // Note: Used for SMBus and GenericSerialBus BufferAc

AML
AccessType := ByteData // Bits 0:3 - Same as AccessType bits of FieldFlags.
// Bits 4:5 - Reserved
// Bits 7:6 - 0 = AccessAttrib = Normal Access Attributes
// 1 = AccessAttrib = AttribBytes (x)
// 2 = AccessAttrib = AttribRawBytes (x)
// 3 = AccessAttrib = AttribRawProcessBytes (x)
//
// x' is encoded as bits 0:7 of the AccessAttrib byte

The iASL compiler implements these.

BTW, in the spec, the example ASL for the use of AttribRawProcessBytes is simply wrong and will not compile.

Bob

jwrdegoede:
Hi,

On 03-08-18 16:28, Robert Moore wrote:
> 
>     spec also says that
>     optionally AttribRawProcessBytes may be used without
>     an AccessLength argument.
> 
> Where does it say this?

ACPI_6.0.pdf page 776 "19.2.7 ASL Parameter Keyword Terms" :

First it is lusted with an AccessLength argument:

"AccessAttribKeyword :=
AttribQuick | AttribSendReceive | AttribByte | AttribBytes (n) | AttribRawBytes (n) |
AttribRawProcessBytes (n) | AttribWord | AttribBlock |AttribProcessCall | ..."

And then again (which I take to mean this is optional) without:

"ExtendedAccessAttribKeyword :=
AttribBytes | AttribRawBytes | AttribRawProcessBytes
// Note: Used for GenericSerialBus BufferAcc only."

Note that this specifically says it is for use with GenericSerialBus BufferAcc.

Also the examples in the spec for ProcessBytes use it *without* an AccessLength
argument:

ACPI_6.0.pdf page 240 "5.5.2.4.5.3.10 Raw Block Process Call (AttribRawProcessBytes)" :

"Field(TOP1, BufferAcc, NoLock, Preserve)
{
     Connection(I2CSerialBus(0x5a,,100000,,"\_SB.I2C",,,,RawDataBuffer(){1,6}))
        AccessAs(BufferAcc, AttribRawProcessBytes) // Use the Raw Bytes Process Call protocol
        FLD0, 8
}"

Note no AccessLength argument.

Regards,

Hans

acpibob:
Here are the full terms:

ExtendedAccessAttribTerm :=
    ExtendedAccessAttribKeyword (
        AccessLength //ByteConst
    )

ExtendedAccessAttribKeyword :=
    AttribBytes | AttribRawBytes | AttribRawProcessBytes
    // Note: Used for GenericSerialBus BufferAcc only.

AccessAsTerm :=
    AccessAs (
        AccessType,     // AccessTypeKeyword
        AccessAttribute // Nothing | ByteConstExpr |
                        // AccessAttribKeyword | ExtendedAccessAttribTerm
    )


What this says to us:

1)	AccessAs can contain an AccessAttribute that is an ExtendedAccessAttribTerm
2)	ExtendedAccessAttribTerm contains an ExtendedAccessAttribKeyword and a length
3)	ExtendedAccessAttribKeyword is one of AttribBytes, AttribRawBytes, or AttribRawProcessBytes

Therefore, the AccessLength is not optional for AttribBytes, AttribRawBytes, or AttribRawProcessBytes.


As far as the example, the AccessLength appears to be simply missing and the code is wrong. The code will not compile for this and other reasons (syntax errors).

acpibob:
In any case, we are trying to figure this whole thing out, and hopefully we can talk to the originator(s) of these GenericSerialBus protocols. Otherwise,  we are just guessing.

acpibob:
Note, from the ASL grammar:

AccessAttribute // Nothing | ByteConstExpr |
 // AccessAttribKeyword | ExtendedAccessAttribTerm

I understand the intended meaning of this, but the notation is incorrect. It should be:

AccessAttribute := Nothing | ByteConstExpr |  AccessAttribKeyword | ExtendedAccessAttribTerm


And from the description of the AccessAs operator:

19.6.1 AccessAs (Change Field Unit Access)

AccessLength is a required argument for some of the Access Attributes.

Access Attributes that require an AccessLength argument:
• AttribBytes (AccessLength)
• AttribRawBytes (AccessLength)
• AttribRawProcessBytes (AccessLength)

acpibob:
Another possibility might be to use the LEN that is specified in the data block. This would be the data length for the write portion of the protocol (write-then-read).


5.5.2.4.6.3.10 Raw Block Process Call (AttribRawProcessBytes)
The GenericSerialBus Raw Write-Read Block Process Call protocol (AttribRawProcessBytes) transfers a
block of data bi-directionally (performs a Write Block followed by a Read Block as an atomic transaction)

// Create the GenericSerialBus data buffer as BUFF
Name(BUFF, Buffer(34)()) // Create GenericSerialBus data buffer as BUFF
CreateByteField(BUFF, 0x00, STAT) // STAT = Status (Byte)
CreateByteField(BUFF, 0x01, LEN) // LEN = Length (Byte)
...
Store(8, LEN) // Length of the valid data

jwrdegoede:
Hi,

On 08/08/2018 05:41 PM, Robert Moore wrote:
> Another possibility might be to use the LEN that is specified in the data block. This would be the data length for the write portion of the protocol (write-then-read).

No that will not work here, only a few bytes are written (a request) and then a much
bigger data block is returned as answer.

Microsoft (in its role as implementer of the specific DSTD we are talking about) seems
to have used GenericSerialBus + RawProcessBytes as a way to build a command - response
protocol here. Note that these GenericSerialBus requests are not going to actual
hardware, they are intercepted bu overriding the GenericSerialBus opregion handler
for this device and this device only.

If you look at the DSDT it cannot work with the normal Opregion handler since there
is no valid address for the i2c resource referenced.

> 5.5.2.4.6.3.10 Raw Block Process Call (AttribRawProcessBytes)
> The GenericSerialBus Raw Write-Read Block Process Call protocol (AttribRawProcessBytes) transfers a
> block of data bi-directionally (performs a Write Block followed by a Read Block as an atomic transaction)

As zetalog commented on this issue already, the real issue here is that ACPICA
makes a copy of the buffer while the spec says this should be passed by reference
to the opregion handler. If things where passed by reference there would not be
a problem, since the opregion implementation could simply do what it wants then.

So the real problem is we're not going to fix the problem of ACPICA passing a fresh
buffer, rather then passing a reference to the actual where the response will be stored,
at least not any time soon.

So we are going to need a workaround to make sure that this freshly allocated buffer is
big enough and that is what the patches from this pull-req do. I know that workarounds
are frowned upon, but given that the patch is not that big / invasive I believe that
we should just move forward with the workaround.

Regards,

Hans

acpibob:
This line from the example code (duplicated across the bi-directional protocols) implies a call-by-value to us:

Store(Store(BUFF, FLD0), BUFF) // Execute the PC

A call-by-reference would be a simple store:

Store (BUFF, FLD0)

acpibob:
I'm putting together some pseudo-code that might help with this discussion.

One problem is that we seem to be reading between the lines of the spec, since the entire section for the protocols is weak.

I'd also like to take this to email to refine the behaviors.

jwrdegoede:
Hi,

On 08-08-18 21:29, Robert Moore wrote:
> I'm putting together some pseudo-code that might help with this discussion.
> 
> One problem is that we seem to be reading between the lines of the spec, since the entire section for the protocols is weak.

Trying to get some clarification of the spec here seems like a good thing and I agree that the spec is ambiguous.

But getting clarification on this from the specification committee is likely going to take a long time and more importantly even if the spec gets clarified we still have to do something to support existing DSTDs out there.

It has been more then a year since this pull-request was opened and in the mean time battery monitoring on the Surface 3 is still broken under linux / freebsd because of this.

AFAICT (from a whole bunch of DSTDs I've access too) the Surface 3 is the only device actually using ProcessBytes, so a workaround should not cause any issues elsewhere and
passing a too large buffer to the opregion handler should never be a problem.

So in the mean time can we please merge the fix which Erik suggested where we make the buffer passed to the opregion handler the biggest of AccessLengh and source_desc->buffer.length ?

Regards,

Hans

bentiss:
> AFAICT (from a whole bunch of DSTDs I've access too) the Surface 3 is the only device actually using ProcessBytes, so a workaround should not cause any issues elsewhere and passing a too large buffer to the opregion handler should never be a problem.

Actually, I have been reported that the Surface Pro 2017 (the latest one) is also using the same AttribRawProcessBytes behavior. See the Device(_SAN) in the attached [DSDT](https://github.com/acpica/acpica/files/2272949/dsdt.txt).

And again, this is used in the exact same way (except this time the Operation Region is on an UART connected device). BTW, the Operation region looks like an I2C handler but the device is on UART, so I think Microsoft really considers this ProcessBytes as a function call from one driver to one other, 'ignoring' the spec, or abusing the blurry part of it.

The `0x2` as parameter seem to be the input ones, and as discussed here, the output parameter is simply missing.

Hope this helps

acpibob:
We want to clarify the proper behavior before we do anything else with this.

I realize that this has been unresolved for some time, but we lost Zetalog from the team and this has slowed things down considerably.

I am completing an outline of an implementation that I will send out hopefully today.

In the meantime, I have a question:

From the spec:
typedef struct
{
    BYTE  Status; // Byte 0 of the data buffer
    BYTE  Length; // Byte 1 of the data buffer
    BYTE[x-1]Data; // Bytes 2-x of the arbitrary length data buffer,
} // where x is the last index of the overall buffer

“length” is the same as LEN. The question is, should LEN include the 2 bytes of the header (Status and Length) or is it simply the number of bytes following in Data?

Thanks,
Bob

acpibob:
As far as the use of "bidirectional":

It appears that the term "bidirectional" is somewhat meaningless, or at best, refers to a lower-level implementation of the protocol, not the ASL itself. In other words, the ASL input buffer is not automatically the same as the output buffer.

Thus, we see ASL code in the examples with the double store:
Store(Store(BUFF, FLD0), BUFF)

this could very well be:
Store(Store(BUFF, FLD0), Local0)

In this case, BUFF would not be overwritten. This is our interpretation.
Bob

acpibob:
From the example in the spec, it appears that LEN refers to the entire buffer from byte 0:

Name(BUFF, Buffer(34)()) // Create GenericSerialBus data buffer as BUFF
CreateByteField(BUFF, 0x00, STAT) // STAT = Status (Byte)
CreateByteField(BUFF, 0x01, LEN) // LEN = Length (Byte)
CreateWordField(BUFF,0x02, DATW) // Data (Bytes 2 and 3), first two bytes of DATA
CreateField(BUFF, 0x10, 256, DATA) // Data (Block) (32 bytes)
...
Store("ACPI", DATA) // Fill in outgoing data
Store(8, LEN) // Length of the valid data

jwrdegoede:
Hi,

On 09-08-18 16:19, Robert Moore wrote:
> We want to clarify the proper behavior before we do anything else with this.
> 
> I realize that this has been unresolved for some time, but we lost Zetalog from the team and this has slowed things down considerably.
> 
> I am completing an outline of an implementation that I will send out hopefully today.
> 
> In the meantime, I have a question:
> 
>  From the spec:
> typedef struct
> {
> BYTE Status; // Byte 0 of the data buffer
> BYTE Length; // Byte 1 of the data buffer
> BYTE[x-1]Data; // Bytes 2-x of the arbitrary length data buffer,
> } // where x is the last index of the overall buffer
> 
> “length” is the same as LEN. The question is, should LEN include the 2 bytes of the header (Status and Length) or is it simply the number of bytes following in Data?

I've been unable to find a clear answer to this, the Linux implementation of
the I2C GenericSerialBus opregion handler stores the number of
bytes actually transferred there (for block transfers only), so if 8 bytes
are transfered then it stores 8 at bytes[1] and stores the data in bytes[2-9].

But looking at all the DSDTs which I've from acpidumps of various devices
not a single DSDT actually uses AttribBlock, so this code is likely untested.

On 09-08-18 16:29, Robert Moore wrote:
 > As far as the use of "bidirectional":
 >
 > It appears that the term "bidirectional" is somewhat meaningless, or at best, refers to a lower-level implementation of the protocol, not the ASL itself. In other words, the ASL input buffer is not automatically the same as the output buffer.
 >
 > Thus, we see ASL code in the examples with the double store:
 > Store(Store(BUFF, FLD0), BUFF)
 >
 > this could very well be:
 > Store(Store(BUFF, FLD0), Local0)
 >
 > In this case, BUFF would not be overwritten. This is our interpretation.

The spec is talking about this being an atomic r/w transfer so
from my reading that means that the only valid way to use this
would be:

Store(Store(BUFF, FLD0), BUFF)

But that is just my 2 cents and I'm in no way an expert on this.

Regards,

Hans

PR #280, opened 2017-06-24
iASL: Ensure that the target node is valid in AcpiExCreateAlias

al3xtjames:
The following ACPI table contains an invalid target node within the
Alias operator:

```
DefinitionBlock ("", "SSDT", 1, "Bug", "BugTable", 0x00001000)
{
    Scope (_SB)
    {
        Device (DEV0)
        {
            Name (_ADR, 1)

            Device (DEV1)
            {
                Alias (_ADR, _ADR)
            }
        }
    }
}
```

If an ACPI table contains such an invalid target node in an Alias
operator, a segmentation fault will occur when the target node is
dereferenced within AcpiExCreateAlias. Add a check for such an invalid
target node in AcpiExCreateAlias.

Signed-off-by: Alex James <theracermaster@gmail.com>

### DIFF
--- a/source/components/executer/excreate.c
+++ b/source/components/executer/excreate.c
@@ -203,6 +203,12 @@ AcpiExCreateAlias (
         TargetNode = ACPI_CAST_PTR (ACPI_NAMESPACE_NODE, TargetNode->Object);
     }
 
+    /* Ensure that the target node is valid */
+    if (!TargetNode)
+    {
+        return_ACPI_STATUS (AE_NULL_OBJECT);
+    }
+
     /*
      * For objects that can never change (i.e., the NS node will
      * permanently point to the same object), we can simply attach


acpibob:
A few comments:

1) The code being changed is kernel ACPICA code, so you cannot use AcpiOsPrintf. In fact, there is no point to the printf since we know the node is a null pointer.

2) We rarely if ever use AE_ERROR, especially in the kernel. We have an exception for this null case, and the return macro should be used:

        return_ACPI_STATUS (AE_NULL_OBJECT);

3) The compiler should catch this as well as the kernel runtime code. This change only fixes the runtime code. You might want to take a look at this.

al3xtjames:
Thanks for the help. I amended the commit, how does it look? I will add the compiler changes soon.

acpibob:
This looks good, I'm taking it now. Any compiler changes can come later.

SchmErik:
I've filed a compiler bug here:
[https://bugs.acpica.org/show_bug.cgi?id=1406](url)

PR #290, opened 2017-07-10
Acpica 1706

zetalog:
ACPICA 201700629 release materials.
No significant divergences can be seen except 2 small changes. They are there because Linux side merged 2 urgent fixes and thus the fixes were not in ACPICA upstream.

This series back ports the ACPICA parts of the 2 commits.

### DIFF
--- a/source/components/hardware/hwregs.c
+++ b/source/components/hardware/hwregs.c
@@ -228,7 +228,7 @@ AcpiHwGetAccessBitWidth (
     }
     else if (Reg->AccessWidth)
     {
-        AccessBitWidth = (1 << (Reg->AccessWidth + 2));
+        AccessBitWidth = ACPI_ACCESS_BIT_WIDTH (Reg->AccessWidth);
     }
     else
     {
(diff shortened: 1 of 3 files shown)


SchmErik:
For 0013cb824aad02bd3b9840a9dd43c3e13172ac3b what exactly is broken on these platforms? Could you provide more information about this breakage?

zetalog:
It looks there are solid proof for reversion:
https://patchwork.kernel.org/patch/9700593/

Anyway it is not such useful to put such a check especially this additional byte has been defined by ACPI spec as a checksum (obviously won't be 0).

zetalog:
We should revert the wrong commit and provide a new one instead of that to check if the last byte is a zero or a checksum.

acpibob:
Yes, it looks like we need to revert this. As per the comment, both iASL and ASL insert a zero for this byte. It would appear that the problem results from a hand-crafted resource descriptor. But, I would like to see the actual ASL code so that we know for sure exactly what is happening. Further, this will help us craft an appropriate comment in the code to ensure that the history of this is tracked.

acpibob:
If we have cases that are non-zero, then I don't think we can even bother looking at the byte. There are too many strange possibilities to be able to separate a checksum from a non-checksum.

acpibob:
Erik and I just remembered that the reason the code was added in the first place was for the binary AML comparisons in the ASLTS. I think it seemed "reasonable" for all cases, which apparently it was not. Just to note that the code was not added for no reason.

PR #302, opened 2017-08-16
Acpica aslts

zetalog:
Cleanups for ASLTS.
This is only the 1st step.
The next step should be to split install_test_cases into aml+/aml-/asl+ targets.


zetalog:
Merged via #305, closing...

PR #304, opened 2017-08-28
acpiexec:change acpiexec's max loop count

SchmErik:
This increases the max loop global and results in fixing a few ASLTS test cases due to large loop iterations.

### DIFF
--- a/source/tools/acpiexec/aemain.c
+++ b/source/tools/acpiexec/aemain.c
@@ -600,7 +600,7 @@ main (
 
     /* ACPICA runtime configuration */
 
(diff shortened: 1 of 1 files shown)


zetalog:
Thanks for the reminder. :)
We've determined to use timeout mechanism instead of this.
Your PR reminds me to update acpiexec option in the determined commit - https://github.com/acpica/acpica/pull/274.

SchmErik:
This is a mistake. I meant to equate this to ACPI_MAX_LOOP_COUNT but this global is already assigned to that during initialization. We should just get rid of this line.

zetalog:
Yes, I removed this line in #274.

PR #306, opened 2017-08-31
acpisrc: don't fclose a file twice on an error exit path

ColinIanKing:
The check on Gbl_CheckAscii in function AsProcessOneFile jumps to the
ErrorFree exit path that fcloses a file a second time.  Avoid this
double fclose by moving first fclose to close to the end path of the
function.

Detected by CoverityScan, CID#175021 ("Use after close")

Fixes: a2c83e75a595 ("acpisrc: Add option to check files for non-printable characters")


acpibob:
got it, as per our email.
thanks,
bob

PR #307, opened 2017-08-31
Fix an off-by-one error in AcpiGetTimerDuration().

juikim:
Delta calculation has an off-by-one error when there is a rollover.
For example, when StartTicks is 0x00FFFFFF and EndTicks is 0x00000000
(for 24-bit timer), DeltaTicks should be 1 (one) but it was 0 (zero).


acpibob:
We'll get it for the next acpica release in September, thanks.

PR #312, opened 2017-09-08
Add converter test sequence

SchmErik:
This pull request adds the converter test sequence to ASLTS

This is similar to the ASLTS disassembler test sequence.
The converter test sequence converts each test case using iasl -ca flag
and then recompiles using flags from a normal test case compilation.
After recompilation, normally compiled AML is compared with the
converted and recompiled AML.


SchmErik:
Updated - add a 60 second timeout for loops

PR #316, opened 2017-09-18
iASL: remove redundant initialization of CommentOption

ColinIanKing:
CommentOption is being initialized and then this value is being
overwritten by a subsequent assignment to CommentOption, hence the
first setting of CommentOption is redundant and can be removed. Cleans
up clang warning:

Value stored to 'CommentOption' during its initialization is never read


acpibob:
looks good, thanks.

PR #328, opened 2017-10-13
Namespace: fix memory leak from building prefixed pathname

SchmErik:


SchmErik:
updated

PR #336, opened 2017-11-02
ACPI 6.2: Additional PPTT flags

jlinton:
The ACPI 6.2 spec has flags to describe cache
allocation, write back, and whether
it is an instruction, data or unified cache.

Signed-off-by: Jeremy Linton <jeremy.linton@arm.com>


acpibob:
We'll get these for the next release, ~ 2 weeks. Thanks.

acpibob:
Needs support in the Data Table Compiler as well as the table disassembler, we can do this.

acpibob:
Looks like the table compiler and disassembler are already implemented.

timofonic:
@acpibob So is it going to be in upstream soon? Thanks!

acpibob:
Changes have been released in ACPICA. Then it is merged into Linux which takes some additional time.

PR #339, opened 2017-11-10
Fix build on a platform where char is unsigned.

juikim:


juikim:
Without this fix, it fails to build like this:
```
In file included from ../../../source/compiler/aslcompiler.l:864:
../../../source/compiler/aslsupport.l:550:48: error: comparison of constant -1 with expression of type 'char' is always true [-Werror,-Wtautological-constant-out-of-range-compare]
    for (i = 0; (yytext[i] != 0) && (yytext[i] != EOF); i++)
                                     ~~~~~~~~~ ^  ~~~
1 error generated.
```

SchmErik:
Thank you for your pull request. Could you change your while loop to a for loop?

juikim:
I realized that the fix is not entirely correct.  I'll open a new issue soon.

juikim:
Please see #341.

PR #341, opened 2017-11-13
Fix build on a platform where char is unsigned.

juikim:
`yytext[]` should never be tested against `EOF` because `lex(1)` cannot
produce `yytext` containing `EOF`.  In fact, `(yytext[i] != EOF)` is always
true on those platforms because `EOF` is defined as an implementation
specific *negative* integer by the ANSI C, often times defined as `-1`.
Specifically for the above expression, `yytext[i]` is implicitly promoted
to a signed integer before comparison and it can never be a negative
value because `sizeof int` is greater than `sizeof char`.

### DIFF
--- a/source/compiler/aslsupport.l
+++ b/source/compiler/aslsupport.l
@@ -525,7 +525,7 @@ static void
 count (
     int                 Type)
 {
-    int                 i;
+    char                *p;
 
 
     switch (Type)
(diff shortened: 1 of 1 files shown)


acpibob:
got it, thanks.

Could I ask why this didn't appear earlier?

juikim:
Actually, it appeared long ago but I had no time to look into it.  I had to fix it because I recently bought Raspberry Pi 2 (running FreeBSD/arm) and 3 (running FreeBSD/arm64). ;-)

BTW, I was told it does not build on FreeBSD/powerpc64 but I have no way to verify it ATM.

PR #352, opened 2018-01-04
License and copyright changes

SchmErik:
Updated the copyright to 2018 and removed the string "All rights reserved" due to its obsolescence.


SchmErik:
next step: insert SPDX license. This requires a little more work because we need to insert different comment styls for .h and .c files.

PR #355, opened 2018-01-25
Linuxize script: ignore actbinfo.h

SchmErik:


SchmErik:
ACPICA version 20180105 creates a few file called actbinfo.h. This is not necessary for Linux so this change is meant to ignore this file

PR #359, opened 2018-02-13
Recognize the _OSI string "Windows 2017.2"

superm1:
Dell uses this string to activate Thunderbolt native mode on supported
machines.

This is necessary to support the work that was done in the Linux kernel for native mode:
https://marc.info/?l=linux-acpi&m=151853954129169&w=2


acpibob:
Is there a “Windows 2017.1” string?

superm1:
I don't believe so.  If so, Dell isn't using it.

acpibob:
I’ve looked at the Microsoft _OSI document, and 2017.2 is not there.  Is this an MS-approved _OSI string?

superm1:
I guess Microsoft still needs to update their spec.

Dell is definitely using it and it works properly when you run on the latest released version of Windows (Fall creators update - RS3).

acpibob:
Ok, we will take, thanks.
Bob

superm1:
FYI, Microsoft has updated documentation around this here:
https://docs.microsoft.com/en-us/windows-hardware/drivers/acpi/winacpi-osi

PR #362, opened 2018-02-21
Define ACPI_IGNORE_PACKAGE_RESOLUTION_ERRORS for FreeBSD.

juikim:
This commit is analogous to a6c3c725c44dd44ad9d3f2b2a64351fdbe6e0014
for Linux.


acpibob:
You might not need this, depending on your init sequence and types of platforms you run on.

I don’t like to just ignore errors, and this feature will report issues with ASL package objects very early (once tables are fully loaded).

Bob

acpibob:
BTW, we are going to make this a runtime option, so you might be able to leave it disabled in the general case, allowing the customer enable it if need be.

The symptom of this problem is seen on Linux as a bunch of NOT_EXIST errors appearing in the dmesg file. It is very rare condition however.


From: Moore, Robert
Sent: Friday, February 23, 2018 7:33 AM
To: 'acpica/acpica' <reply@reply.github.com>; acpica/acpica <acpica@noreply.github.com>
Cc: Subscribed <subscribed@noreply.github.com>
Subject: RE: [acpica/acpica] Define ACPI_IGNORE_PACKAGE_RESOLUTION_ERRORS for FreeBSD. (#362)

You might not need this, depending on your init sequence and types of platforms you run on.

I don’t like to just ignore errors, and this feature will report issues with ASL package objects very early (once tables are fully loaded).

Bob

juikim:
Understood.  Thanks for the info!

PR #363, opened 2018-02-22
Initialize FixedStatus/FixedEnable variables

anatol:
Latest versions of GCC are able to track variable initialization/usage.
And the compiler produces warnings that FixedStatus/FixedEnable can be
used without any initialization. It can happen if AcpiHwRegisterRead()
fails, the function does not assign output value in this case. But
despite that FixedStatus/FixedEnable are still used. Initialize these
variables to 0 to silense the warning.

### DIFF
--- a/source/components/events/evevent.c
+++ b/source/components/events/evevent.c
@@ -341,8 +341,8 @@ AcpiEvFixedEventDetect (
     void)
 {
     UINT32                  IntStatus = ACPI_INTERRUPT_NOT_HANDLED;
-    UINT32                  FixedStatus;
-    UINT32                  FixedEnable;
+    UINT32                  FixedStatus = 0;
+    UINT32                  FixedEnable = 0;
     UINT32                  i;
 
 


acpibob:
What version of gcc?

anatol:
gcc 7.3.0 with `LTO` enabled.

SchmErik:
I'm using 7.3.1 distributed by the fedora package manager. I've tried adding the -flto flag but I am unable to reproduce what you're seeing...

anatol:
I use `-O3`. Also I am using Arch Linux, maybe the compiler itself been compiled with different set of flags.

But it is better to look at the code itself to understand the issue.  `AcpiHwRegisterRead` initializes output value only [if status is `SUCCESS`](https://github.com/acpica/acpica/blob/0d5f6f68453d770a4917cb7f0bb502712c3dc97c/source/components/hardware/hwregs.c#L749). If status is a failure then this stack variable value is undefined.

Later `AcpiEvFixedEventDetect` does not check result status of `AcpiHwRegisterRead` (by the way why?) but tries to use invalid values of FixedStatus/FixedEnable. So either this variables need to be initialized to some known value. Or result of `AcpiHwRegisterRead` need to be checked and if it is a failue then do no use FixedStatus/FixedEnable.

acpibob:
I think that calls to AcpiHwRegisterRead should always check the return status, exit on failure.

SchmErik:
new pull request #370

acpibob:
we did this slightly differently, so I'm closing this one.

PR #368, opened 2018-03-01
Add support for hex AML C header file generation

samimujawar:
The iAsl compiler provides a list of options to support
'Firmware Support - C Text Output'.

One of the options, '-tc' supports generation of AML hex
tables in C. This command generates a character array that
represents the AML hex table data.
        e.g. unsigned char AmlCode[] = {
               // HEX AML data
               ...
               };

However, there is no option to change the name of the
generated character array (AmlCode) in the output hex file.
This limits the ability to reference more than one table from
the firmware code.

Also, the generated output does not have header file include
guards. This is ideally desired if the hex file is to be
included from a C source file.

A solution to this is to modify the Hex C output such that
the name of the array is derived from the input file name,
and to add header file include guards.

To address the above, without breaking existing implementations,
a new option '-th' is introduced in the list of supported
'Firmware Support - C Text Output' commands. This option
generates an AML hex table output in a C header file suitable
for including from a C source file.

The '-th' option:
1. Uses the input file name as a prefix for the generated
   C array name.
   e.g. 'unsigned char DadtAmlCode[]'

2. Adds a header file include guard.
   e.g.
        #ifndef DSDT_HEX_
        #define DSDT_HEX_
        ...

        unsigned char DadtAmlCode[] = {
                // HEX AML data
                ...
        };

        #endif // DSDT_HEX_

  Where Dadt.aml is the input file name.

  Note: The header file include guard uses '<InputFileName>_HEX_'
        format instead of '<InputFileName>_H_' to avoid possible
        collision with an existing header guard.

Signed-off-by: Sami Mujawar <sami.mujawar@arm.com>
Signed-off-by: Evan Lloyd <evan.lloyd@arm.com>

### DIFF
--- a/source/compiler/aslcompile.c
+++ b/source/compiler/aslcompile.c
@@ -544,7 +544,8 @@ AslCompilerSignon (
             Prefix = "; ";
         }
         else if ((Gbl_HexOutputFlag == HEX_OUTPUT_C) ||
-                 (Gbl_HexOutputFlag == HEX_OUTPUT_ASL))
+                 (Gbl_HexOutputFlag == HEX_OUTPUT_ASL) ||
+                 (Gbl_HexOutputFlag == HEX_OUTPUT_H))
         {
             FlPrintFile (ASL_FILE_HEX_OUTPUT, "/*\n");
             Prefix = " * ";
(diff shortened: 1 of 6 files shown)


SchmErik:
Please change the style for if statements so that braces are on separate lines like this:
if
{
    // code
}
else
{
    // code
}

samimujawar:
Fixed patch based on review comments and new pull request generated at https://github.com/acpica/acpica/pull/372

acpibob:
Implemented, closing

PR #377, opened 2018-04-09
iASL: add check for illegal forward references

SchmErik:
This commit changes the behavior of iASL to emit errors on forward
references within control methods. Changes in ASLTS are made to
account for this compiler change.


SchmErik:
this has already been committed. closing

PR #378, opened 2018-04-10
Fix potential infinite loop in AcpiRsDumpByteList

ColinIanKing:
There is a potential infinite loop if AcpiRsDumpByteList is
called with a Length greater than 255 since the current loop
counter is just a UINT8 and will wrap to zero and never reach
the desired value in Length.  Fix this by making the loop
counter the size type as Length.


acpibob:
got it, thanks.
bob

PR #379, opened 2018-04-12
iASL: add a check for some definition block parameters for compliance

SchmErik:
This commit addds a check during the compilation of a definition
block. It ensures that OEM table ID is at most 8 characters and the
OEM ID is at most 6 characters.


SchmErik:
closing, already implemented

PR #383, opened 2018-05-03
acpidump and acpixtract should handle table sizes greater than 1MB

tsoome:
There is a bug in the interaction of acpidump and acpixtract when the table
size is greater than 1MB. The acpixtract code will stop parsing a table if
the first character on a line is not a space (' '). The acpidump code will
overflow the offset into the first character after 1MB. 

This patch is developed in illumos.

### DIFF
--- a/source/components/utilities/utbuffer.c
+++ b/source/components/utilities/utbuffer.c
@@ -205,7 +205,7 @@ AcpiUtDumpBuffer (
     {
         /* Print current offset */
 
-        AcpiOsPrintf ("%6.4X: ", (BaseOffset + i));
+        AcpiOsPrintf ("%7.4X: ", (BaseOffset + i));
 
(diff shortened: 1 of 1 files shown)


acpibob:
Could you post a binary table > 1mb and some examples. For our testing and documentation. Thanks.

acpibob:
May as well go to %8.4 and be done with it, yes?

We will have to check if there are any side effects with the data table compiler

We now have a > 1mb dsdt, so ignore the previous message.

tsoome:
yes, i think %8.4 would be fine.

acpibob:
Sorry, but even %7.4 will not work properly without changes to acpixtract.

(try a binary compare after a dump->xtract sequence)

However, I will try to fix it.

acpibob:
Completed and committed. Thanks.

PR #390, opened 2018-06-14
AcpiExec: fixing -fi option

SchmErik:
Field elements listed in the init file used to be initialized after
the table load and before executing module-level code blocks. The
recent changes in module-level code means that the table load becomes
a method execution. If fields are used within module-level code and
we are executing with -fi option, then these values are populated
after the table has finished loading. This commit changes the
initialization of objects listed in the init file so that field unit
values are populated during the table load.


SchmErik:
This is ready now. Go ahead and look it over.

PR #392, opened 2018-07-02
debugger: fix build warnings on set but not used variable NumStatements

ColinIanKing:
When building the Linux kernel integrated version of ACPICA with clang
we get warnings that NumStatements is set but not used. Fix this by
adding more #if guarding around this variable so that it is only
declared and set when it is being required.

Fixes Linux clang build warning:
warning: variable 'num_statements' set but not used [-Wunused-but-set-variable]

### DIFF
--- a/source/components/debugger/dbmethod.c
+++ b/source/components/debugger/dbmethod.c
@@ -412,7 +412,9 @@ AcpiDbDisassembleAml (
     char                    *Statements,
     ACPI_PARSE_OBJECT       *Op)
 {
+#ifdef ACPI_DISASSEMBLER
     UINT32                  NumStatements = 8;
+#endif
 
 
     if (!Op)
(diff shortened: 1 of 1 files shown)


acpibob:
we need a bit of clarification.

So, this is a build that includes the AML debugger but not the disassembler?

If so, probably the entire "List" command should go away in this case.

ColinIanKing:
I believe this in the AML debugger mode, as when the disassembler mode is enabled with  ACPI_DISASSEMBLER defined, then the reference to num_statements is being made and the warning won't occur in that latter case.

acpibob:
So it sounds like the debugger (can be) in the kernel, but the disassembler is not.

I'm going to completely disable the "List" command in this case. This is the only caller for AcpiDbDisassembleAml.

ColinIanKing:
Hrm.  Is the "List" command useful in the kernel debugger?  So disabling it may be reducing functionality to kernel / ACPI developers.

ColinIanKing:
Ah, OK, understood.  In which case yes, List can be removed :+1:

acpibob:
I can take it from here.

acpibob:
It looks like the "Debug" (single-step) command depends on the disassembler also.

acpibob:
Closing as per comments above.

PR #410, opened 2018-08-28
Use %d for signed int print formatting instead of %u 

ColinIanKing:
… is defined

When building ACPICA in the Linux kernel with Clang with ACPI_DISASSEMBLER
not defined we get a the following warning on variable DisplayOp:

warning: variable 'display_op' set but not used [-Wunused-but-set-variable]

Fix this warning by guarding the declaration of DisplayOp with some
ACPI_DISASSEMBLER #ifdefs

### DIFF
--- a/source/compiler/aslcompile.c
+++ b/source/compiler/aslcompile.c
@@ -858,7 +858,7 @@ CmCleanupAndExit (
 
     if (AslGbl_ExceptionCount[ASL_ERROR] > ASL_MAX_ERROR_COUNT)
     {
-        printf ("\nMaximum error count (%u) exceeded\n",
+        printf ("\nMaximum error count (%d) exceeded\n",
             ASL_MAX_ERROR_COUNT);
     }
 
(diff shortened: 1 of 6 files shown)


SchmErik:
Hi @ColinIanKing, thanks for the pull request. We'll take %u commit. The debugger change can be done by encapsulating lines 333-444 into a separate function function surrounded by an ifdef. I can take care of this and give you credit for the suggestion. It would be great if you could update this pull request to include an updated version of the first commit.

Could you send me a patch that you used in order to build the kernel with clang?

Thanks

ColinIanKing:
Hi there,

1. I've re-pushed my %u commit so it's now in sync with the current ACPICA repo.
2. Sure, do what ever is required for the debugger change, and thanks for crediting me with the fix suggestion.
3. I'm using the following for a kernel clang build:

make -Oline CC=clang-7  -k CC_HAVE_ASM_GOTO=1 CC=clang-7 KBUILD_CFLAGS=-D__BPF_TRACING__ 

That's a hack :-)

PR #26, opened 2014-04-14
Acpica endian

zetalog:
This patchset is a demonstration to add combined unalignement/endianness support for ACPICA.

The new feature can be an opt-in feature.
After we cleaned up all of the existing code, the old features could then opt-out.

PR #411, opened 2018-09-16
Provide support for big-endian machines

ahs3:
This represents a few years of collected patches and a lot of testing
with a lot of different ACPI tables in source and binary form.

Getting ACPICA tools to build and run on big-endian machines is not
that difficult.  The issue is that binary forms of the ACPI tables
must, by definition, be in little-endian format.  Without providing
the changes to read/write tables in little-endian, running these
tools on big-endian machines doesn't really work correctly.

This patch touches all the places in ACPICA where it is assumed that
the data being manipulated is in little-endian form.  In those cases,
changes are made so that the actual data format is irrelevant, but
when it is read or written from a binary table, it is always read or
written as little-endian and converted into a form usable on the
current host.

There may be better ways to do this; as mentioned, this represents an
accretion of changes over many years.  At some point, some clever
refactoring is probably in order.

The veracity of the ACPI tables produced has been checked with some
basic testing.  The first test used takes an ACPI table in source form,
compiles it with iasl, and compares the binary table with the known
good version from an x86_64 version of iasl.  This verifies that we can
write out data in proper little-endian format.  The second test is that
the binary form of a table is read, and then disassembled into source
form.  The source form is then compared to a known good version from
x86_75 iasl; this verifies we can read little-endian format.

Note, also, that these changes have been used by a number of end
users on Debian and Fedora distributions over the last several years,
as well.  Hopefully, this has provided even more extensive testing.

Signed-off-by: Al Stone <ahs3@ahs3.net>


acpibob:
In theory, most of the support is there already, but we don't get many requests for it these days. 

We will need to closely review this, after vacations are over.

acpibob:
So, let me ask a possibly dumb question. For applications/tools, doesn't the OS have a "misalignment handler" that is invoked on a misalignment fault? Then, the OS pieces together the misaligned request and returns from the fault?

ahs3:
> So, let me ask a possibly dumb question. For applications/tools, doesn't the OS have a "misalignment
> handler" that is invoked on a misalignment fault? Then, the OS pieces together the misaligned request
> and returns from the fault?

You mean, like when you ask for misaligned 16 bits from a 64 bit register?  Well, I've worked on machines that do accommodate that and machines that do not, especially many older ones.  I've seen those as traps caught by run-time libraries, by the OS, or compilers that just don't generate such cases.
Is there a specific case in question, or a specific part of this merge request?  That would really help me understand the question better.

SchmErik:
How old are these machines that do not support misalignment?

ahs3:
> In any case, this will certainly take us a good chunk of time to review and make suggestions. The sheer size and extent of these changes gives me a fright.

Understood.  And sorry about that -- it's a multi-year accumulation, unfortunately.

ahs3:
> We see two different issues here:
> 
> 1. Big-endian support where data must be swapped back to little-endian for things like ACPI tables, etc.
> 2. Support for machines that don't automatically handle misalignment issues, defined as:
> 
> * reading/writing a 64-bit number that is not on a "natural" 64-bit boundary
> * reading/writing a 32-bit number that is not on a "natural" 32-bit boundary
> * reading/writing a 16-bit number that is not on a "natural" 16-bit boundary
> * 8-bit values can always be read/written.
> 
> We've done some work on the misalignment and big-ending support, but as you've discovered, it is not complete.

Oh, I think I might understand a bit better.  I really haven't seen much of any problems with 64-, 32- or 16-bit values not being on a usable boundary.  There a couple of exceptions on PowerPC and there's a different patch for the specific problem found there.  Just because of the way the tables are defined, alignment is usually pretty solidly 16-, 32- or 64-bit aligned.  If it is not, it's usually a byte array, or there's a reserved field in the right place making is easy to use the ACPI_MOVE_x_TO_y macros; does that help answer the questions?

acpibob:
Actually, there are lots of tables that are not "aligned". There are at least 60 different tables. Fundamentally, the ACPI tables are defined to be little-endian, and natural alignment is not guaranteed.  Often, there is an attempt to define the tables such that they are in fact aligned, but this is not always the case.

Certainly things like the DSDT and SSDTs have no alignment.

acpibob:
some issues after applying the patch:

obj/acpiexec ../../../source/components/tables/tbutils.c
../../../source/components/tables/tbutils.c: In function ‘AcpiTbGetRootTableEntry’:
../../../source/components/tables/tbutils.c:349:14: error: cast from pointer to integer of different size [-Werror=pointer-to-cast]
  Address64 = (UINT64) TableEntry;
              ^
In file included from ../../../source/include/acpi.h:165:0,
                 from ../../../source/components/tables/tbutils.c:152:
../../../source/include/actypes.h:652:48: error: cast to pointer from integer of different size [-Werror=int-to-pointer-cast]
 #define ACPI_CAST_PTR(t, p)             ((t *) (ACPI_UINTPTR_T) (p))
                                                ^
../../../source/components/tables/tbutils.c:362:43: note: in expansion of macro ‘ACPI_CAST_PTR’
         return ((ACPI_PHYSICAL_ADDRESS) (*ACPI_CAST_PTR (
                                           ^~~~~~~~~~~~~
cc1: all warnings being treated as errors
make[1]: *** [../Makefile.rules:20: obj/tbutils.o] Error 1
make[1]: Leaving directory '/cygdrive/c/acpica/generate/unix/acpiexec'
make: *** [generate/unix/Makefile.common:7: acpiexec] Error 2

obj/iasl ../../../source/compiler/aslcodegen.c
../../../source/compiler/aslcodegen.c: In function ‘CgWriteTableHeader’:
../../../source/compiler/aslcodegen.c:651:13: error: ‘TableHeader’ undeclared (first use in this function)
     DWord = TableHeader.Length;
             ^~~~~~~~~~~

acpibob:
FYI, Lv worked on the endian/alignment stuff back in 2014, you may want to take a look at the pull request, it is #26

acpibob:
Did you get a chance to look at the "alternate" endian/alignment code posted as a pull request?

ahs3:
> Did you get a chance to look at the "alternate" endian/alignment code posted as a pull request?

My apologies, but not yet.  I've had to work on some other issues for a while, but hope to get back to this soon.

ahs3:
Yo, @acpibob: I'm going to close this PR.  1) I've let it get too crufty so there's some bit rot (patch rot?); 2) I noticed the addition of AcpiNs{In,Ex}ternalizeName functions -- so, I think the proper path is to convert the changes started here to use these new functions instead.  Sorry about the delays, but thanks for the review, time and energy spent.

ahs3:
The content of this PR can be replaced with calls to AcpiNs{In,Ex}ternalizeName functions instead.  So, cancelling this PR in order to properly patch with the new functions.

PR #412, opened 2018-09-17
Add a man page for each tool

ahs3:
For each of the various ACPICA tools, create a Linux man page.

This patch adds the source for each man page, and changes the Makefiles
in order to install them properly.

Signed-off-by: Al Stone <ahs3@ahs3.net>

### DIFF
--- a/generate/unix/Makefile.config
+++ b/generate/unix/Makefile.config
@@ -42,10 +42,12 @@ CC ?=    gcc
 #
 OBJDIR =     obj
 BINDIR =     bin
+MANDIR =     man
 COMPILEOBJ = $(CC) -c $(CFLAGS) $(OPT_CFLAGS) -o $@ $<
 LINKPROG =   $(CC) $(OBJECTS) -o $(PROG) $(LDFLAGS) $(OPT_LDFLAGS)
 PREFIX ?=    /usr
 INSTALLDIR = $(PREFIX)/bin
+MANINSTALLDIR = $(PREFIX)/share/man/man1
 UNAME_S := $(shell uname -s)
 
 #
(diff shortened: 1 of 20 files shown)


acpibob:
This comes up once in a while.

I could be wrong, but I seem to recall that the man files are not what we would call "OS-independent", as the man program is different across the various unix-flavor operating systems.

acpibob:
we'll take a closer look when Erik gets back from vacation.

acpibob:
These look good, and we will take the changes.

We'd like to put the man pages at acpica/man-pages. The "generate" directory is not appropriate because this directory is primarily used for makefiles and project files.

Some of the dates in the makefiles look odd:
.TH ACPIEXEC 1 "January 23, 2013"

We moved the source repository to several different source control systems before git, so I think the date should simply be the copyright on all of tools, 2000.  Any month and day would probably be ok.

ahs3:
I'd almost forgotten about this patch set.  Is there still interest?  Clearly, some updating would be necessary, and then be resubmitted.

sacdintel:
Hi @ahs3 ,

I will have a chat with @acpibob next week in the office and we can decide on what (if) changes need to be made to merge this PR. I think from what I understand, this PR adds a bunch of lines of code thereby essentially bloating ACPICA. A typical client BIOS image is usually <2MB or so, we don't want to burden that further, although these days memory is not generally a bottleneck. It would be nice to see the man pages as optional with the help of a flag to enable it or so. Let me know your thoughts, thanks.

-Saket Dumbre

sacdintel:
On further inspection, it does look like you have applied install and remove flags for the man page, so I don't have a big problem to merge it, but comments and feedback from @acpibob would be great. Thanks a lot!

SaketADumbre:
Hi @ahs3 ,
 
Hope you are doing alright! I would like to convey my interest in merging this PR if and when the prior comments by @acpibob are addressed and this PR updated. This would be a great tool to add for the obvious educational reasons. Please let me know your thoughts, I hope it is not too late to try and bring this stale PR back to life!

maciejwieczorretman:
Stale PR and we're looking at a different way to document the project. Closing

PR #417, opened 2018-10-11
AML interpreter: add region addresses in global list during initializ…

SchmErik:
…ation

The table load process omitted adding the operation region address
range to the global list. This omission is problematic because the OS
queries the global list to check for address range conflicts before
deciding which drivers to load. This commit may result in warning
messages that look like the following:

[    7.871761] ACPI Warning: SystemIO range 0x00000428-0x0000042F conflicts with OpRegion 0x00000400-0x0000047F (\PMIO) (20180531/utaddress-213)
[    7.871769] ACPI: If an ACPI driver is available for this device, you should use it instead of the native driver

However, these messages do not signify regressions. It is a result of
properly adding address ranges within the global address list.

Link: https://bugzilla.kernel.org/show_bug.cgi?id=200011
Tested-by: Jean-Marc Lenoir <archlinux@jihemel.com>

### DIFF
--- a/source/components/dispatcher/dsopcode.c
+++ b/source/components/dispatcher/dsopcode.c
@@ -586,6 +586,9 @@ AcpiDsEvalRegionOperands (
         ObjDesc, ACPI_FORMAT_UINT64 (ObjDesc->Region.Address),
         ObjDesc->Region.Length));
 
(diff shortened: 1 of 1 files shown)


SchmErik:
We can wait to merge this until it gets to Linux upstream for more testing (around end of October).

PR #425, opened 2018-10-25
iASL: properly detect continuation on dos files

SchmErik:
If the '\\' is next to a "\r\n", it is also a continuation


SchmErik:
updated, this effectively "fixes" our support for CRLF

PR #428, opened 2018-11-09
iASL: restrict integer package elements to integers and constants

SchmErik:
Package elements that are integer expressions are not supported on
any known AML interpreter implementations. This change in the
compiler restricts integer package elements to integer literals and
constants.


SchmErik:
Closing for now, I will update next week

PR #436, opened 2018-12-19
Correct DSDT Address field in FACP tables

ahs3:
The FADT allows either the DSDT Address or XDSDT Address field to be
zero.  However, the table definition used by the table compiler still
requires the DSDT Address to be non-zero, which is not correct.  So,
remove the DT_NON_ZERO flag from the field.

Signed-off-by: Al Stone <ahs3@redhat.com>

### DIFF
--- a/source/common/dmtbinfo.c
+++ b/source/common/dmtbinfo.c
@@ -292,7 +292,7 @@ ACPI_DMTABLE_INFO           AcpiDmTableInfoFacs[] =
 ACPI_DMTABLE_INFO           AcpiDmTableInfoFadt1[] =
 {
     {ACPI_DMT_UINT32,   ACPI_FADT_OFFSET (Facs),                    "FACS Address", 0},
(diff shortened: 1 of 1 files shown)


paulmenzel:
Looks good.

SchmErik:
This definition allows both fields to be zero. We should probably write logic in the table compiler to ensure that at least one of these fields are non-zero..

acpibob:
Agree.

SchmErik:
So #568 suggests that people are still using 32-bit operating systems... Maybe it's just safer to leave it as-is... Is there anything wrong with iASL forcing support for 32-bit operating systems?

jlevon:
This is a 32-bit acpixtract binary, not a 32-bit kernel.

SchmErik:
oh my mistake. Thanks for the clarification!

acpibob:
Ok, got it. Now, if both the 32-bit and 64-bit fields are zero, an error is thrown.

PR #442, opened 2019-01-28
ASLTS: split variable definitions for clarity, no functional change

SchmErik:


SchmErik:
This is a cosmetic change, I was browsing the source and found that this variable naming makes it a little easier to understand

PR #456, opened 2019-03-29
Multiple file compilation feature

SchmErik:
This feature allows multiple ASL files to be compiled in the same namespace. This provides a way to detect unresolved external declarations as well as redundant named object declarations during compilation.


SchmErik:
Hi @acpibob this pull request is ready

PR #458, opened 2019-04-09
debugger: use correct printf format specifiers for various integers

ColinIanKing:
Use lu for unsigned longs and zd for sizeof() sizes. Detected by
static analysis with Coverity.


acpibob:
The problem we have had in the past was that “lu” and “zd” are not supported by all compilers. An additional complication is that we need to support both 32-bit and 64-bit compilers.

acpibob:
Technically, not all C libraries.

ColinIanKing:
OK fair comment.

PR #477, opened 2019-07-04
Debugger: remove redundant assignment on ObjDesc

ColinIanKing:
Pointer ObjDesc is being initialized with a value that is never
read and it is being updated later with a new value. The initialization
is redundant and can be removed.

Addresses-Coverity: ("Unused value")


SchmErik:
looks good to me

PR #490, opened 2019-08-07
Static operation region in method

irp20:


acpibob:
Actually, this applies to most all named objects (warn on declaring an object within a method). With the possible exceptions of Integers, Strings, and Buffers.

SchmErik:
closing due to conflicts. Rebased/edited pull request available here: https://github.com/acpica/acpica/pull/551

PR #491, opened 2019-08-07
Check reserved bytes field unit does not extend beyond region limit

irp20:


SchmErik:
This code seems to generate false positives...

SchmErik:
rebased version of this pull request available here: https://github.com/acpica/acpica/pull/587

closing

PR #503, opened 2019-08-27
ASLTS: add option to only execute data table compiler tests

SchmErik:


SchmErik:
updated to include a fix for the -tp option

PR #506, opened 2019-09-26
make acpi_load_table() return table index

SchmErik:
For unloading an ACPI table, it is necessary to provide the index of
the table. The method intended for dynamically loading or hotplug
addition of tables, AcpiLoadTable(), should provide this information
via an optional pointer to the loaded table index.

This patch fixes the table unload function of acpi_configfs.

Reported-by: Andy Shevchenko <andriy.shevchenko@linux.intel.com>
Fixes: d06c47e3dd07f ("ACPI: configfs: Resolve objects on host-directed
table loads")
Signed-off-by: Nikolaus Voss <nikolaus.voss@loewensteinmedical.de>

### DIFF
--- a/source/components/debugger/dbfileio.c
+++ b/source/components/debugger/dbfileio.c
@@ -253,7 +253,7 @@ AcpiDbLoadTables (
     {
         Table = TableListHead->Table;
 
-        Status = AcpiLoadTable (Table);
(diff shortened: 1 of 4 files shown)


SchmErik:
updated with acpiexec tests to exercise new load/unload interfaces

PR #507, opened 2019-10-03
Disassembler: IORT: fix output of optional padding field

SchmErik:
The disassembly of the optional padding field resulted in dumping
the remainder of the length in the subtable. This is problematic
when there are mappings that are describes after the padding field.

The length of the mappings are a part of the subtable length. When
dumping the optional padding field, the mapping length should be
subtracted from the rest of the field so that the proper length of
padding should be dumped.


acpibob:
Does this fix Bugzilla #1469?

PR #514, opened 2019-10-19
iASL: fix memory leak of LocalityBuffer on error return path

ColinIanKing:
The error return path is not free'ing previously allocated
resource LocalityBuffer and hence there is a memory leak. Fix this
by free'ing it before returning.

Fixes: c22c0b1df98c ("Data table compiler: Add error messages for SLIT locality buffers")


SchmErik:
Looks good to me. Thanks!

PR #564, opened 2020-02-26
Fix GCC 10 / -fno-common

SoapGentoo:
* GCC 10 switches its default symbol emission mode
  to cause linker errors when an object is defined
  more than once.

Bug: https://bugs.gentoo.org/706672

### DIFF
--- a/source/compiler/aslcompiler.l
+++ b/source/compiler/aslcompiler.l
@@ -156,7 +156,7 @@
 
 #include <stdlib.h>
 #include <string.h>
-YYSTYPE AslCompilerlval;
+#define YYSTYPE AslCompilerlval;
 
 /*
  * Generation:  Use the following command line:
--- a/source/compiler/dtparser.l
+++ b/source/compiler/dtparser.l
@@ -208,7 +208,7 @@ NewLine         [\n]
 /*
  * Local support functions
  */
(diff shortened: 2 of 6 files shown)


source/compiler/aslcompiler.l
```diff
@@ -156,7 +156,7 @@
 
 #include <stdlib.h>
 #include <string.h>
-YYSTYPE AslCompilerlval;
+#define YYSTYPE AslCompilerlval;
```
SchmErik: I'm taking a look at this right now but I don't have access to gcc-10 (yet). What happens when you remove this #define? Could you show me the error that you get?

SoapGentoo: Indeed, this line can be removed without error. Want me to remove it?

SoapGentoo:
@SchmErik you don't need GCC 10 to test this, just use `OPT_CFLAGS="-fno-common"` and you'd get the errors you'd see with GCC 10.

SchmErik:
Thanks for the report. I was able to get these fixed using your suggestions as well as a reporter on our bugzilla. I combined all of our solutions and submitted a pull request here: #566 Since not all of anyone's solution made it, I just decided to credit you as a reporter. Let me know if this is a problem. I'll close this but let's take the discussion to #566 and I'll close this.

PR #566, opened 2020-03-04
Bz1522

SchmErik:
This pull request contains commits that allow ACPICA to build with GCC 10 which (by default) is more strict about redundant variable declarations.

### DIFF
--- a/source/compiler/aslcompiler.l
+++ b/source/compiler/aslcompiler.l
@@ -156,7 +156,6 @@
 
 #include <stdlib.h>
 #include <string.h>
-YYSTYPE AslCompilerlval;
 
 /*
  * Generation:  Use the following command line:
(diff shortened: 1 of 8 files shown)


SchmErik:
@juikim could you test this? I installed byacc on my system but it gives me the max table size error... Could you give me a link to the one that has the max table size override?

juikim:
> @juikim could you test this?

It builds fine.

> I installed byacc on my system but it gives me the max table size error... Could you give me a link to the one that has the max table size override?

You have to rebuild byacc from source.

1.  Download the latest byacc source.

https://invisible-island.net/datafiles/release/byacc.tar.gz

2.  Configure it with ``--with-max-table-size`` option.

For example,
```
./configure --with-max-table-size=2147483647
```
Note the value in the above example is ``0x7fffffff`` and it is ``INT_MAX`` on FreeBSD.

3.  Compile byacc.

Run ``make`` to build.

SoapGentoo:
@SchmErik looks good

SchmErik:
@acpibob please merge this pull request

PR #568, opened 2020-03-13
utilities: fix sprintf()

jlevon:
On 32-bit, the provided sprintf() is non-functional: with a size of
ACPI_UINT32_MAX, String + Size will wrap, meaning End < Start, and
AcpiUtBoundStringOutput() will never output anything as a result.

The symptom we saw of this was acpixtract failing to output anything.

Signed-off-by: John Levon <john.levon@joyent.com>

### DIFF
--- a/source/components/utilities/utprint.c
+++ b/source/components/utilities/utprint.c
@@ -565,7 +565,7 @@ vsnprintf (
     if (Size != ACPI_UINT32_MAX) {
         End = String + Size;
     } else {
-        End = (char *)(uintptr_t)ACPI_UINT32_MAX;
+        End = ACPI_CAST_PTR(char, ACPI_UINT32_MAX):
     }
 
     for (; *Format; ++Format)
--- a/source/components/utilities/utprint.c
+++ b/source/components/utilities/utprint.c
@@ -560,7 +560,13 @@ vsnprintf (
 
 
     Pos = String;
-    End = String + Size;
+
+
+    if (Size != ACPI_UINT32_MAX) {
+        End = String + Size;
+    } else {
+        End = (char *)(uintptr_t)ACPI_UINT32_MAX;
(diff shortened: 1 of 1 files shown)


acpibob:
Are you saying that the 32-bit version of acpixtract never outputs anything?

jlevon:
```
# acpixtract -a acpidump.out 

Intel ACPI Component Architecture
ACPI Binary Table Extraction Utility version 20180629
Copyright (c) 2000 - 2018 Intel Corporation

Could not open output file 
```

Because of:

```
open("", O_RDWR|O_CREAT|O_TRUNC, 0666)		Err#2 ENOENT
```

which is due to the aforementioned sprintf() bug.

acpibob:
So, what does "open("", O_RDWR|O_CREAT|O_TRUNC, 0666)" have to do with sprintf?

jlevon:
https://github.com/acpica/acpica/blob/master/source/tools/acpixtract/acpixtract.c#L288

So Gbl_OutputFilename stays ""

acpibob:
I don't see where our makefile or MSVC project file for acpixtract actually uses utprint.c; are you using a custom build?

jlevon:
Yes, this is from an import into our OS (illumos).
For what it's worth, I would also be happy with complete deletion of utprint.c (which is what we've done). However, none of that changes the fact this sprintf() implementation is broken.

acpibob:
You'll need to change this line:
End = (char *)(uintptr_t)ACPI_UINT32_MAX;
To something like:
End = ACPI_CAST_PTR(char, ACPI_UINT32_MAX);
Because on some systems uintptr_t doesn't exist.

jlevon:
Fixed, thanks.

acpibob:
got it, thanks.

PR #577, opened 2020-03-23
iASL: implement custom iASL macro for expected errors

SchmErik:


acpibob:
got it, thanks.

acpibob:
I think you may need to add new structs to acpisrc, such as ASL_LOCATION_NODE.

PR #578, opened 2020-03-23
acpixtract: fix AX_IS_TABLE_BLOCK_HEADER macro

SchmErik:
This macro needs to be surrounded by parens. Otherwise, a logical
statement that applies a logical not operator to this macros could
result in a computation that applies the operator to the left side of
the logical and but not the right.

Reported-by: John Levon <john.levon@joyent.com>


acpibob:
got it, thanks.

PR #587, opened 2020-04-08
iASL: Check Reserved_Bytes Fields don't extend beyond region limit

SchmErik:

### DIFF
--- a/source/compiler/aslxref.c
+++ b/source/compiler/aslxref.c
@@ -149,6 +149,12 @@
  *
  *****************************************************************************/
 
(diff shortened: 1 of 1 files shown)


SchmErik:
this isn't ready yet

SchmErik:
It's likely that this does not account for cases when offset and length of the operation region are variables

acpibob:
We better check it out.

rafaeljw:
This got outdated and needs to be closed and possibly refreshed.

acpibob:

PR #589, opened 2020-04-09
compiler/aslanalyze: Fix test condition on 2nd char for GPE method

ElyesH:
The test '((Name[1] != 'L') && (Name[1] != 'E'))' is always false.
As the 2nd char must be L or E, we would use:
if !((Name[1] == 'L') || (Name[1] == 'E'))
	return;

Signed-off-by: Elyes HAOUAS <ehaouas@noos.fr>

### DIFF
--- a/source/compiler/aslanalyze.c
+++ b/source/compiler/aslanalyze.c
@@ -587,7 +587,7 @@ ApCheckForGpeNameConflict (
      * 3rd/4th chars must be a hex number
      */
     if ((Name[0] != '_') ||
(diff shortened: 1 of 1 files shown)


acpibob:
I disagree with "The test '((Name[1] != 'L') && (Name[1] != 'E'))' is always false".
By DeMorgan's theorem:
!(Cond1 || Cond2) is equivalent to (!Cond1 && !Cond2)

ElyesH:
indeed, my mistake.
Thx

PR #591, opened 2020-04-17
Dispatcher: add status checks to avoid NULL pointer dereference on fi…

SchmErik:
…eld objects

Reported-by: Kurt Kennett <kurt_kennett@hotmail.com>


ElyesH:
Now, we write on "Status = AE_OK" at "source/components/dispatcher/dsfield.c" line 347 and we never use it 
we overwrite on it at line 400

PR #600, opened 2020-04-28
psparse: Do not parse if DSDT AML has zero size

zamaudio:
Bypasses a page fault when the DSDT is present but has empty aml.

Linux seems to not experience a fault from this, perhaps there is a workaround in Linux for this (?)

In any case, it seems like a waste of cpu cycles to attempt parsing an empty AML.

Signed-off-by: Damien Zammit <damien@zamaudio.com>

### DIFF
--- a/source/components/parser/psparse.c
+++ b/source/components/parser/psparse.c
@@ -579,6 +579,13 @@ AcpiPsParseAml (
         return_ACPI_STATUS (AE_BAD_ADDRESS);
     }
 
+    /* If size of AML is zero, we succeeded. There is nothing to do */
+
+    if (!WalkState->ParserState.AmlSize)
+    {
+        return_ACPI_STATUS (AE_OK);
+    }
+
     /* Create and initialize a new thread state */
 
     Thread = AcpiUtCreateThreadState ();


SchmErik:
How did you test this on linux?

zamaudio:
@SchmErik I used the same qemu machine to boot Linux as my customised GNU/Hurd with ACPICA integrated into the mach kernel.  Linux seems to boot in QEMU anyway, I've never seen it fail on parsing the qemu DSDT.  The page fault was something I experienced without the patch applied using my custom Hurd OS.  With this patch, the problem goes away.

SchmErik:
QEMU will expose ACPI tables including a non-null dsdt to the guest operating system. Are you saying that you're trying to run linux with a null dsdt?

SchmErik:
Closing this doesn't seem to be an issue. QEMU exposes ACPI tables to linux. If you're running an ACPI enabled operating system and the firmware does not contain a DSDT, it is better to take the fault because the DSDT is a fundamental and necessary component of ACPI firmware. Closing

paulmenzel:
Error handling would still be useful, wouldn’t it? That means, printing an error message helping to debug a possible issue?

Th3Fanbus:
> Closing this doesn't seem to be an issue.

Well, it *is* an issue: in some cases, it causes a page fault. While one could reason that returning success in such an unexpected situation is a bad idea, it would be less bad than causing a page fault. For instance, one could return `AE_BAD_ADDRESS` instead, as done when the DSDT is missing. Thoughts?

PR #602, opened 2020-05-11
Add parsing for IVRS IVHD 40h and device entry F0h

amonakov:
I was investigating a Linux boot issue related to AMD IVRS table, and noticed that iasl could not fully decode the table on my laptop: support for sub-table type 0x40 and entry type 0xf0 were missing. This pull request adds the necessary support.


SchmErik:
Could you give us a link to a document that contains the IVRS table definition?

amonakov:
Sure, it's available from AMD at https://www.amd.com/en/support/tech-docs in the document titled "AMD I/O Virtualization Technology (IOMMU) Specification". Here's a direct link to the PDF: https://www.amd.com/system/files/TechDocs/48882_IOMMU_3.05_PUB.pdf

source/common/dmtbdump2.c
```diff
@@ -588,6 +588,14 @@ AcpiDmDumpIvrs (
                     InfoTable = AcpiDmTableInfoIvrs8c;
                     break;
 
+                /* Variable-length entries */
+
+                case ACPI_IVRS_TYPE_HID:
+
+                    EntryLength = 22;
```
ElyesH, line 595: regarding AMD I/O Virtualization Technology (IOMMU) Specification (48882—Rev 3.05-PUB—Jan 2020)
" EntryLength"  = Size of IVHD block in bytes, starting from Type field and including IVHD device entries.

By my understand "EntryLength" should be greater than 22. it should be  ( 40 + n +1), or I'm wrong ?

amonakov: 22 is correct, see table 99 in the document. You are probably confusing entry length with the header that precedes a list of entries, which is described in table 89.

ElyesH: indeed, 22 is correct.
Thank you.

acpibob:
Could you update the template for IVRS to include the 40h and device entry F0h? (use iASL -tc to get the hex dump of the compiled input file.)

acpibob:
Start with the existing template (iasl -T ivrs), modify ivrs.asl to add the new stuff, compile it the file with the -tc flag (iasl -tc ivrs.asl). then add the hex output (ivrs.hex) to the template file (dttemplate.h)

amonakov:
Sorry, but the 10-month delay is not inspiring. I'm going to need assurances that everything else in the patch is fine, and the additions will be reviewed in a timely manner to even consider doing the extra work. This was a volunteer effort. Could you merge it as-is instead?

SchmErik:
Hey, it doesn't look like there's compiler support for this. Do you have an example of an IVRS that you dumped using these changes? It would be helpful for us to finish the implementation for the compiler and templates if we had an example. Thanks

amonakov:
Here's an example, uploaded as a Github attachment: [ivrs-example.zip](https://github.com/acpica/acpica/files/6154958/ivrs-example.zip).

acpibob:
Thanks. Sorry for the delay.

acpibob:
Do you have any more IVRS examples? I would prefer a couple of them to make sure that the compiler support fully works.
thanks,
bob

amonakov:
Several ACPI tables dumped from consumer hardware, including IVRS, can be found in this repository: https://github.com/linuxhw/ACPI . At the moment it seems to include 32 samples for IVRS, some of which are recent enough to include sub-table 0x40.

acpibob:
Thanks.
BTW, I have the prototype compiler support for IVRS up and running.

acpibob:
Unfortunately, I don’t think I can use the IVRS files in this repository, since they are disassembled with an older version of the compiler, and there are errors like this:

**** Unknown IVRS subtable type 0x11

I might be able to clean some of these up however.

amonakov:
> Unfortunately, I don’t think I can use the IVRS files in this repository, since they are disassembled

Hm? Hexdumps of raw tables are still available there, no need for some sort of manual cleanup.

acpibob:
Ok, those are the .bin files, correct?

acpibob:
BTW, it looks like AMD just published a new version of the IOMMU spec, which contains the IVRS.

amonakov:
> Ok, those are the .bin files, correct?

There's also a hexdump immediately following a (partially) disassembled table, under the `Raw Table Data:` heading. But yeah, many of the dumps have corresponding `.bin` files that could be easier to work with.

PR #609, opened 2020-06-13
ASLTS: Replace inaccurate language

kees:
The tests/ subdirectory contains confusing (and offensive) language
that is not tied to any "external" specifications. The thread worker
naming convention is more accurately named "Worker". (The controlling
thread was already named "Controller".) Rename one file and update the
language throughout.

This removes 209 instances, leaving only the 67 that are tied to
specification language.

For background and further details on the technical and societal
rationale for these changes, please see:
https://tools.ietf.org/id/draft-knodel-terminology-01.html

Signed-off-by: Kees Cook <kees@outflux.net>


acpibob:
What exactly is "offensive" concerning the original language? Do tell.

acpibob:
OK, I just read the document at the link you provided.  Thanks for the changes, we will probably integrate them.

PR #612, opened 2020-07-06
iASL: clean up loop counter mismatch with upper bound

ColinIanKing:
The loop counter i is a U16 where as the upper bound is a UINT32 which
in theory could lead to an infinite loop if AslGbl_ElevatedMessagesIndex
is greater than 65535.  In practice this is probably never going to
occur, but make code fully safe by making the loop counter a UINT32.

Addresses-Coverity: ("Infinite loop")


acpibob:
OK, thanks.

PR #621, opened 2020-08-04
ACPI 6.3: HMAT updates

michalbiesek:
- rename Processor to Attached Initiator
- drop deprecated bits from Memory Proximity Domain flag
- bump ACPI version in HMAT comment

Signed-off-by: Michal Biesek <michal.biesek@intel.com>


SchmErik:
looks good to me

PR #623, opened 2020-08-10
iASL: parse valid package elements returned by predefined control met…

SchmErik:
…hods

### DIFF
--- a/source/compiler/aslprepkg.c
+++ b/source/compiler/aslprepkg.c
@@ -318,6 +318,17 @@ ApCheckPackage (
          */
         for (i = 0; i < Count; i++)
         {
+            if (!Op)
+            {
+                /*
+                 * If we get to this point, it means that the package length
+                 * is larger than the initializer list. Stop processing the
+                 * package and return because we have run out of package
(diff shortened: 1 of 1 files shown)


acpibob:
From the ACPI spec:
Package (NumElements) {PackageList} =>Package
...
"If NumElements is present and greater than the number of elements in the PackageList, the default entry of type Uninitialized is used to initialize the package elements beyond those initialized from the PackageList."

It doesn't look like this change handles the case above, but I'm not sure.

SchmErik:
Thanks for bringing this up. No, this PR doesn't address this. It just prevents a null dereference. @acpibob, are you saying that this behavior should be implemented in the interpreter or should the uninitialized elements be inserted in the AML bytecode?

acpibob:
I think it should be implemented in the interpreter; the code already exists somewhere, AFAIK.

Which predefined name is this from?

SchmErik:
The original report was from _PSL

SchmErik:
Here's the dsdt from the original report: https://github.com/acpica/acpica/files/5049983/DSDT_crash.zip

PR #624, opened 2020-08-13
ACPI 6.3: Remove last level memory cache

michalbiesek:
Update System Locality Latency and Bandwidth Information Flags possible
values: remove "Last level memory" and update rest of values

Signed-off-by: michalbiesek <michal.biesek@intel.com>


SchmErik:
looks good to me

PR #632, opened 2020-09-23
tests/aslts.sh: Remove unnecessary tabs at end of line

ElyesH:
unnecessary 2 tabs at end of line removed


acpibob:
Got it, thanks.

PR #633, opened 2020-09-23
generate/release/build.sh: Remove tabs at end of line

ElyesH:
unnecessary  2 tabs at end of line removed


acpibob:
got it, thanks.

PR #634, opened 2020-09-23
generate/release/release.sh: Remove tabs at end of lines

ElyesH:
unnecessary tabs at end of lines removed


acpibob:
got it, thanks.

PR #638, opened 2020-10-09
IORT: Updates for revision E

shamiali2008:
IORT revision E(ARM DEN 0049E) contains a few additions like,
    -Added an identifier field in the node descriptors to aid table
     cross-referencing.
    -Introduced the Reserved Memory Range(RMR) node. This is used
     to describe memory ranges that are used by endpoints and requires
     a unity mapping in SMMU.
    -Introduced a flag in the RC node to express support for PRI.


shamiali2008:
The IORT Rev E doc can be found here,
https://developer.arm.com/documentation/den0049/latest/

Thanks,
Shameer

SchmErik:
looks good to me

acpibob:
merged, thanks.

shamiali2008:
> looks good to me

Thanks for taking a look. ARM team has indicated that they will be updating the IORT spec soon with some changes to RMR. So the Linux kernel patches are now waiting for that update.  Can we wait for the updated IORT RMR spec before these changes land in the Linux kernel, please?

Thanks,
Shameer

SchmErik:
What does that mean for IORT revision E right now? Are you saying that there IORT revision E is basically a prototype and there is no real-world use case for this?

shamiali2008:
Yes, AFAIK, there are no current users for revision E. That is mainly because ARM wanted to update few things around RMR specification and I am told by ARM folks that the revised spec will be out soon.  Hence decided not to push the kernel patches for now.

As I type this, just checked, there is an E.b of IORT spec out now,
https://developer.arm.com/documentation/den0049/latest/

I will check with ARM folks to confirm that is a final one and work on it.

Thanks,
Shameer

PR #640, opened 2020-10-18
Fix exception code class checks

qzed:
Fixes exception code class checks in `acexcep.h`.

Currently, the exception code class check macros are broken. For example, `ACPI_ENV_EXCEPTION(Status)` will always evaluate to zero, whereas `ACPI_AML_EXCEPTION(Status)` will evaluate to non-zero for `AE_CODE_PROGRAMMER`, `AE_CODE_ACPI_TABLES`, as well as `AE_CODE_AML`. The other macros have similar problems.

This issue likely has not been detected before as these classification macros do currently not see much use.

The first commit in this PR fixes those macros, the second commit cleans up existing checks of the form `((Status & AE_CODE_MAKS) == <class>)` by replacing them with the corresponding `ACPI_<class>_EXCEPTION(Status)` macro.

### DIFF
--- a/source/components/debugger/dbobject.c
+++ b/source/components/debugger/dbobject.c
@@ -201,7 +201,7 @@ AcpiDbDumpMethodInfo (
 
     /* Ignore control codes, they are not errors */
 
-    if ((Status & AE_CODE_MASK) == AE_CODE_CONTROL)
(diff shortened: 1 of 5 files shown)


acpibob:
Ok, thanks. We will get to this soon.

PR #646, opened 2020-10-30
Provide support for big-endian systems

ahs3:
Let's try this a different way.  In prior attempts at providing support for big-endian systems, replacing macros was sort of working but it was getting very fragile as upstream code improved.  What this attempt does is reduce the functions needed to a a small handful constrained to a single file (source/components/utilities/utendian.c), and then make as few changes as possible elsewhere.  In general, we only had to change code when we actually had to read or write little-endian values to a file.

So, for example, AML code generation was pretty well contained whether assembling or disassembling.  For the static tables, most of the time very few lines needed changing.  The only places where things got a little weird was in handling all of the resource types in AML/ASL files; the patches to fix SSDT/DSDTs are the only really large ones as a result.

Even with the changes, what look like function calls basically turn into assignment statements on little-endian machines so in theory the only performance impact is when running on big-endian machines.

After all of these changes, ASLTS still passes all of the tests (big- or little-endian), and we always produce correct binary little-endian files regardless of where we are running (this was tested with ~2000 different ACPI binary tables of all types).

I know this is a lot to review.  Hopefully breaking this up into a lot of patches, each specific to a type of table, makes it a little simplet (most of the static table patches are very similar, for example).

Let me know if there are any questions.

### DIFF
--- a/source/common/dmtbdump2.c
+++ b/source/common/dmtbdump2.c
@@ -1103,11 +1103,13 @@ AcpiDmDumpMpst (
     UINT16                  SubtableCount;
     UINT32                  PowerStateCount;
     UINT32                  ComponentCount;
+    UINT32                  TableLength = AcpiUtReadUint32 (&Table->Length);
+    UINT16                  Tmp16;
 
 
     /* Main table */
 
-    Status = AcpiDmDumpTable (Table->Length, 0, Table, 0, AcpiDmTableInfoMpst);
+    Status = AcpiDmDumpTable (TableLength, 0, Table, 0, AcpiDmTableInfoMpst);
     if (ACPI_FAILURE (Status))
     {
         return;
@@ -1115,13 +1117,14 @@ AcpiDmDumpMpst (
 
     /* Subtable: Memory Power Node(s) */
 
-    SubtableCount = (ACPI_CAST_PTR (ACPI_TABLE_MPST, Table))->PowerNodeCount;
+    Tmp16 = (ACPI_CAST_PTR (ACPI_TABLE_MPST, Table))->PowerNodeCount;
+    SubtableCount = AcpiUtReadUint16 (&Tmp16);
     Subtable0 = ACPI_ADD_PTR (ACPI_MPST_POWER_NODE, Table, Offset);
 
-    while ((Offset < Table->Length) && SubtableCount)
+    while ((Offset < TableLength) && SubtableCount)
     {
         AcpiOsPrintf ("\n");
-        Status = AcpiDmDumpTable (Table->Length, Offset, Subtable0,
+        Status = AcpiDmDumpTable (TableLength, Offset, Subtable0,
             sizeof (ACPI_MPST_POWER_NODE), AcpiDmTableInfoMpst0);
         if (ACPI_FAILURE (Status))
         {
@@ -1130,8 +1133,8 @@ AcpiDmDumpMpst (
 
         /* Extract the sub-subtable counts */
 
-        PowerStateCount = Subtable0->NumPowerStates;
-        ComponentCount = Subtable0->NumPhysicalComponents;
+        PowerStateCount = AcpiUtReadUint16 (&Subtable0->NumPowerStates);
+        ComponentCount = AcpiUtReadUint16 (&Subtable0->NumPhysicalComponents);
         Offset += sizeof (ACPI_MPST_POWER_NODE);
 
         /* Sub-subtables - Memory Power State Structure(s) */
@@ -1142,7 +1145,7 @@ AcpiDmDumpMpst (
         while (PowerStateCount)
         {
             AcpiOsPrintf ("\n");
-            Status = AcpiDmDumpTable (Table->Length, Offset, Subtable0A,
+            Status = AcpiDmDumpTable (TableLength, Offset, Subtable0A,
                 sizeof (ACPI_MPST_POWER_STATE), AcpiDmTableInfoMpst0A);
             if (ACPI_FAILURE (Status))
             {
@@ -1165,7 +1168,7 @@ AcpiDmDumpMpst (
 
         while (ComponentCount)
         {
-            Status = AcpiDmDumpTable (Table->Length, Offset, Subtable0B,
+            Status = AcpiDmDumpTable (TableLength, Offset, Subtable0B,
                 sizeof (ACPI_MPST_COMPONENT), AcpiDmTableInfoMpst0B);
             if (ACPI_FAILURE (Status))
             {
@@ -1182,22 +1185,24 @@ AcpiDmDumpMpst (
         SubtableCount--;
         Subtable0 = ACPI_ADD_PTR (ACPI_MPST_POWER_NODE, Subtable0,
             sizeof (ACPI_MPST_POWER_NODE) +
-            (sizeof (ACPI_MPST_POWER_STATE) * Subtable0->NumPowerStates) +
-            (sizeof (ACPI_MPST_COMPONENT) * Subtable0->NumPhysicalComponents));
+            (sizeof (ACPI_MPST_POWER_STATE) *
+                AcpiUtReadUint16 (&Subtable0->NumPowerStates)) +
+            (sizeof (ACPI_MPST_COMPONENT) *
+                AcpiUtReadUint16 (&Subtable0->NumPhysicalComponents)));
     }
 
     /* Subtable: Count of Memory Power State Characteristic structures */
 
     AcpiOsPrintf ("\n");
     Subtable1 = ACPI_CAST_PTR (ACPI_MPST_DATA_HDR, Subtable0);
-    Status = AcpiDmDumpTable (Table->Length, Offset, Subtable1,
+    Status = AcpiDmDumpTable (TableLength, Offset, Subtable1,
         sizeof (ACPI_MPST_DATA_HDR), AcpiDmTableInfoMpst1);
     if (ACPI_FAILURE (Status))
     {
         return;
     }
 
-    SubtableCount = Subtable1->CharacteristicsCount;
+    SubtableCount = AcpiUtReadUint16 (&Subtable1->CharacteristicsCount);
     Offset += sizeof (ACPI_MPST_DATA_HDR);
 
     /* Subtable: Memory Power State Characteristics structure(s) */
@@ -1205,10 +1210,10 @@ AcpiDmDumpMpst (
     Subtable2 = ACPI_ADD_PTR (ACPI_MPST_POWER_DATA, Subtable1,
         sizeof (ACPI_MPST_DATA_HDR));
 
-    while ((Offset < Table->Length) && SubtableCount)
+    while ((Offset < TableLength) && SubtableCount)
     {
         AcpiOsPrintf ("\n");
-        Status = AcpiDmDumpTable (Table->Length, Offset, Subtable2,
+        Status = AcpiDmDumpTable (TableLength, Offset, Subtable2,
             sizeof (ACPI_MPST_POWER_DATA), AcpiDmTableInfoMpst2);
         if (ACPI_FAILURE (Status))
         {
--- a/source/common/dmtbdump2.c
+++ b/source/common/dmtbdump2.c
@@ -1300,18 +1300,23 @@ AcpiDmDumpNfit (
     ACPI_STATUS             Status;
     UINT32                  Offset = sizeof (ACPI_TABLE_NFIT);
     UINT32                  FieldOffset = 0;
-    UINT32                  Length;
+    UINT32                  TableLength = AcpiUtReadUint32 (&Table->Length);
     ACPI_NFIT_HEADER        *Subtable;
     ACPI_DMTABLE_INFO       *InfoTable;
     ACPI_NFIT_INTERLEAVE    *Interleave = NULL;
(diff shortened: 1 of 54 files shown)


SchmErik:
We're working on getting a test system right now... It might take a while.. Out of curiosity, which system did you test this on?

ahs3:
> We're working on getting a test system right now... It might take a while.. Out of curiosity, which system did you test this on?

Fedora rawhide on s390x, plus x86_64, armhf, aarch64, ppc64le

ahs3:
I forgot a couple: this also works on Fedora i686, RHEL (very different compiler path, all arches), and Debian for s390, x86_64 and aarch64

SchmErik:
hey Al, we're curious as to why you decided to use functions instead of the existing macros to differentiate little vs big endian. Could you explain?

ahs3:
> hey Al, we're curious as to why you decided to use functions instead of the existing macros to differentiate little vs big endian. Could you explain?

Sure.  I've been using the macros for several years.  The patches for big-endian support that used the macros were getting more and more fragile.  Something would improve (thank you, btw :) and getting the macros just right again would get harder somewhere.  It got to where I just needed to fix it.

My goal was to make these changes so that the internal representation of everything could stay in the host-native form, and was only converted to little-endian when it absolutely had to be -- when it was read from or written to an external representation (ASL or binary files, basically).  I even tried some experiments in the lexer and the parser code to see if I could make it completely invisible (too much code outside of those relies on endianness, it turns out).

What I found was that I could not rely on the macros to be used consistently.  For example, ACPI_MOVE_32_TO_32 was sometimes used, and ACPI_COPY_NAMESEG was also used, in both cases to copy a name segment (sometimes even the same one) -- except of course when memcpy() or strcpy() was used.  The cleanup in the use of the macros over time (replacing ACPI_MOVE_32_TO_32 with ACPI_COPY_NAMESEG or memcpy() when that is appropriate) is a good thing, but the transition makes things kind of fuzzy to me.  So, part of it was just trying to be clearer about what was happening, and stop overloading the semantics of a macro that was already overloaded.

I also found cases where ACPI_MOVE_x_TO_y  was used as a memcpy() but then it would be used on the same value at different times during compilation or disassembly.  What I would end up with is a value that was always the wrong endianness at some phase of processing (if you fixed one spot, it broke the other, and vice versa).  With the functions, I could hope to keep the semantics clearer.

Maintainability touched on several things.  With the macros, I ended up using a pattern of putting a value into a temporary location, and then ACPI_MOVE_x_TO_y,  A bit clunky, but it worked.  In a lot of cases, I could use a function return value in the original assignment instead, making things simpler to read.  I couldn't do that everywhere, but I got a bunch of them.

This also prevented me from having a whole group of macros to deal with all of the implications of casts to and from all of the different permutations (1 byte -> 2,4, 8 and back, 2 bytes -> 1,2,4,8 and back, and so on).  In a bunch of the cases, I could now replace a series of one or more casts with a single function call.  This also let me use the compiler to do some type checking that the macros side-step.

As a side effect, the functions let me move bytes around more safely.  I can always be sure that the original value is untouched and has not been overwritten unless I explicitly choose to do so.  So, while compiling a parse tree node, an internal structure, I can keep it in host-native form always and only swap bytes when I need to.  For example, a lot of the tables and subtables and opcodes have length fields; if I used ACPI_MOVE_x_TO_y, I would change the endianness of the value.  But, I may need to set the value and then increment it multiple times in multiple different places -- I would have to keep track of when I have and have not swapped bytes,  With the functions, I always return a value on the stack so I can be very clear about when a value has to change.

And finally, most compilers these days can handle optimizing out leaf functions.  At that point, there's not a lot of difference in the resulting executable when using macros vs the functions.  In theory, the functions were written in such a way that they would only impose additional minor overhead on big-endian machines (maybe) and would be such simple leaf functions on little-endian machines that the call would get optimized out.

That's a bit of stream of consciousness there, but those are the things that led me to this approach.

acpibob:
I’ve noticed a couple of things about #646:

1) It looks like tabs are used in the code. (Or at least some things don’t line up.)
2) There is no space after many function invocations before the argument list (inconsistent at best). Same with the sizeof operator. We have to be careful about these because of the Linuxization process.
3) Function declarations should have 1 line per argument, same as the prototypes.

Otherwise, the changes look pretty good.

BTW, does your patch remove the "ACPI_MOVE_x_TO_y" macros?  If not, should they be removed?

Thanks,
Bob

ahs3:
Oh, bother.  You're correct; I missed a few of the ACPICA conventions and used Linux conventions instead.  And I do know that can mess up the "Linuxization" so no excuse, really.  Sorry about that.  Should those all get fixed first?

No, I don't think we can remove the ACPI_MOVE_x_TO_y macros yet; my patch does not.  They are used a *lot* (especially ACPI_MOVE_32_TO_32).  That sounds like a great job for a summer intern -- go through and replace them all with memcpy() or assignments.  Or it's something I can try to whittle away at over time.

ahs3:
For version 2:

- Cleaned up all of the uses of sizeof and function calls so that they have proper spacing, consistent with ACPICA
- Cleaned up the declarations and definitions of the new endian functions so that they're consistent with ACPICA usage
- Dropped two patches for MTMR and VRTC tables since those are deprecated tables and no longer in use (see also [issue #649](https://github.com/acpica/acpica/issues/649))

ahs3:
For version 3:

Somehow missed changes to the iASL Unix Makefile that accidentally removed all the tabs.  Put them back in so it all builds properly.  Note, too, that I should have mentioned that cleaning up the spacing and tabs was part of version 2, also.  This is what removed them from the Makefile (silly thinko).

SchmErik:
@ash3, could you squash the Makefile tab/space commit to your first commit?

ahs3:
> @ahs3, could you squash the Makefile tab/space commit to your first commit?

Sure, Erik.  I'll see if I can get to it today or tomorrow.  Still digging out from the holidays ...

SchmErik:
> > @ahs3, could you squash the Makefile tab/space commit to your first commit?
> 
> Sure, Erik. I'll see if I can get to it today or tomorrow. Still digging out from the holidays ...

No problem. I completely understand. I'm doing the same..

acpibob:
Couple of things:
Did you compile this support in both 32-bit and 64-bit compiler mode?
Did you generate the code on MSVC (2017)?

acpibob:
Also, did you run a static analyzer such as Clang on the code?


From: Moore, Robert
Sent: Wednesday, January 13, 2021 8:46 AM
To: acpica/acpica <reply@reply.github.com>; acpica/acpica <acpica@noreply.github.com>
Cc: Comment <comment@noreply.github.com>
Subject: RE: [acpica/acpica] Provide support for big-endian systems (#646)

Couple of things:
Did you compile this support in both 32-bit and 64-bit compiler mode?
Did you generate the code on MSVC (2017)?

acpibob:
Re:
#if defined(ACPI_BIG_ENDIAN)

I think I would rather go with the existing run-time detection of the endianness of the machine. This way, a single binary can run on both types of machine – as well as not requiring makefile changes or compile-time option setup. Also, requiring a compile-time option will force all of our users to modify their per-host include files.

Bob

ahs3:
> Couple of things: Did you compile this support in both 32-bit and 64-bit compiler mode? Did you generate the code on MSVC (2017)? 

I have sporadic access to armv7hf machines; I'll see what happens.  The Debian compile on that arch seems to
have a problem I'm hunting down.  I'll see if I can find a couple of other compilers to play with for the 32-bit.

I do not have access to MSVC (other than a Linux version, which is probably not ideal for this question).

ahs3:
> Re: #if defined(ACPI_BIG_ENDIAN) I think I would rather go with the existing run-time detection of the endianness of the machine. This way, a single binary can run on both types of machine – as well as not requiring makefile changes or compile-time option setup. Also, requiring a compile-time option will force all of our users to modify their per-host include files. Bob

I don't understand.  The executable (e.g., iasl) will always be big- or little-endian, yes?  I don't know of any systems that compile both endians into a single executable and determine endianness at run-time.  And the ACPI tables must by definition be little-endian, regardless of what machine is building the tables.  These patches require no change of the ASL files, and the AML will always end up little-endian as it should be.

This is also why ACPI_BIG_ENDIAN is defined in source/include/platform/aclinux.h.  It should only have a
bearing on Linux-based big-endian systems, and will determine that from the system the executable (e.g.,
iasl) is being built on.   As far as I can tell, unless a platform/OS supports big-endian, there should be no changes needed -- everything should default to little-endian.  And as far as I know, only Linux seems to be asking for big-endian support.

So I'm puzzled; I'm probably just being dense, but I don't understand what this comment means.

ahs3:
> > > @ahs3, could you squash the Makefile tab/space commit to your first commit?
> > 
> > 
> > Sure, Erik. I'll see if I can get to it today or tomorrow. Still digging out from the holidays ...
> 
> No problem. I completely understand. I'm doing the same..

Done.

acpibob:
Yes, I’m saying that a single executable can go both ways (chosen at runtime). It simplifies the build – at a small cost of actual code.

ahs3:
> Yes, I’m saying that a single executable can go both ways (chosen at runtime). It simplifies the build – at a small cost of actual code. 

Aha.  Okay, I think I'm following now.  Let me try that out -- do run-time determination of byte ordering vs compile time.

acpibob:
OK. We already have a function that determines endianness at runtime:

In aslmain.c:
if (UtIsBigEndianMachine ())

ahs3:
Just a quick update: I'm currently testing a modified version of the patches to eliminate the #ifdef ACPI_BIG_ENDIAN and do things at run-time.  I've also gotten some feedback from some 32-bit tests to look at.  So, this is not forgotten work, but it has been on the back burner.

acpibob:
That’s ok. We are busy implementing ACPI 6.4 anyway.

ahs3:
Okay.  Rebased to match upstream, updated for ACPI 6.4 (new tables), verified the formatting, and have now tested on s390x (big-endian), 32-bit (armhf7, i386), GCC 10 and 11, but all Fedora.  The test cases cover ~5600 different ASL files.

The testing has also uncovered some disassembly problems (there are a couple of patches in this series for that), but I'm still hunting down some additional ones.  None of these affect big-endian support at this point.

So, all is ready.  What additional info might be needed?

acpibob:
>The testing has also uncovered some disassembly problems

What are these related to?

ahs3:
>> The testing has also uncovered some disassembly problems
> What are these related to? 

I've got a bunch of DSDT and SSDT files (>100) that disassemble in such a way that they cannot be recompiled; I'm still sorting out underlying causes, and trying to determine if there are multiple causes.  Best I can tell so far, one case appears to be where an If() is used as a parameter in a Store() embedded in an If() -- but, I'm still isolating exactly what's happening, though this appears to be the primary problem in most of the files.  Bisecting the ASL/AML down to just a few constructs has been really messy so far.  I'm double checking whether or not I can redistribute the original ASL/AML since some of the machines I work with are not yet shipping.  Note, too, that in some cases the AML was not produced by iASL so I'm also trying to determine if the opcode sequences are 100% correct, and the language construct is even allowed.

That being said, in the remaining 5000+ test cases, compiling correct ASL produces the same results on either endian; similarly, disassembling the original AML produces the same results.  If nothing else, all the test results are consistent on big- or little-endian.

source/common/dmtbdump2.c
```diff
@@ -1681,39 +1681,43 @@ AcpiDmDumpPhat (
...
 
     Subtable = ACPI_ADD_PTR (ACPI_PHAT_HEADER, Table, sizeof (ACPI_TABLE_PHAT));
 
-    while (Offset < Table->Length)
+    while (Offset < TableLength)
     {
         /* Common subtable header */
 
+        SubtableType = AcpiUtReadUint16 (&Subtable->Type);
+	Subtable->Type = SubtableType;
```
ahs3, line 1699: This line causes a segfault with 'iasl -T phat' and needs to be removed.  GCC 11 is the compiler, on Fedora33, and I suspect a compiler problem at this point.  Regardless, this is extraneous code and is not needed so removing the line has no effect.

source/common/dmtbdump2.c
```diff
@@ -1681,39 +1681,43 @@ AcpiDmDumpPhat (
...
+        SubtableLength = AcpiUtReadUint16 (&Subtable->Length);
+	Subtable->Length = SubtableLength;
```
ahs3, line 1701: Same problem here as with line 1699; this one line (1701) needs to be removed

acpibob:
I don’t see that code in dmtbdump2.c (or at least our current head). Also, no segfault.

acpibob:
Likewise, I don’t see this code in dmtbdump2.c

ahs3:
Ah, sorry.  Those were mostly notes to myself to touch up one of the patches.  And since these have not been merged yet, they of course do not appear in dmtbdump2.c yet.  This only segfaults on s390x when these patches (as is) have been applied; I think I'm running into a compiler problem and the comments reflect a temporary workaround while research continues.

I suppose that was a bit confusing -- sorry about that.

rafaeljw:
I think that this PR can be closed with the expectation to continue the discussion after posting a new pull request making the changes in question.

acpibob:

PR #648, opened 2020-11-03
Add non-clang CFLAGS only on non-clang compilers

tbaederr:
When using clang on Linux, the Makefile was passing unsupported flags to
the compiler. Fix this by only passing those unsupported by clang if
$(CC) is not clang.

I was not sure if the FreeBSD/Apple distinction is still wanted after this patch, but I left it in for now.

### DIFF
--- a/generate/unix/Makefile.config
+++ b/generate/unix/Makefile.config
@@ -227,22 +227,39 @@ CWARNINGFLAGS += \
     -Wold-style-definition\
     -Wno-format-nonliteral\
     -Wredundant-decls
-#
-# Per-host flags and exclusions
-#
-ifneq ($(ACPI_HOST), _FreeBSD)
-    CWARNINGFLAGS += \
-        -Wempty-body
 
-    ifneq ($(ACPI_HOST), _APPLE)
-        CWARNINGFLAGS += \
-            -Woverride-init\
-            -Wlogical-op\
-            -Wmissing-parameter-type\
-            -Wold-style-declaration\
-            -Wtype-limits
-    endif
-endif
+# output directory for tests below
+#
+# Taken from the Linux kernel
+TMPOUT = $(if $(KBUILD_EXTMOD),$(firstword $(KBUILD_EXTMOD))/).tmp_$$$$
+
+# try-run
+# Usage: option = $(call try-run, $(CC)...-o "$$TMP",option-ok,otherwise)
+# Exit code chooses option. "$$TMP" serves as a temporary file and is
+# automatically cleaned up.
+#
+# Taken from the Linux kernel
+try-run = $(shell set -e;		\
+	TMP=$(TMPOUT)/tmp;		\
+	TMPO=$(TMPOUT)/tmp.o;		\
+	mkdir -p $(TMPOUT);		\
+	trap "rm -rf $(TMPOUT)" EXIT;	\
+	if ($(1)) >/dev/null 2>&1;	\
+	then echo "$(2)";		\
+	else echo "$(3)";		\
+	fi)
+
+# Returns either the given flag if supported, or nothing
+#
+# Taken from the Linux kernel, but modified to fit our needs
+cc-option = $(call try-run, $(CC) -Werror $(1) -c -x c /dev/null -o "$$TMP",$(1),)
+
+CWARNINGFLAGS += $(call cc-option,-Wempty-body)
+CWARNINGFLAGS += $(call cc-option,-Wlogical-op)
+CWARNINGFLAGS += $(call cc-option,-Wmissing-parameter-type)
+CWARNINGFLAGS += $(call cc-option,-Wold-style-declaration)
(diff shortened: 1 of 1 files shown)


SchmErik:
Did you mean to do something like this instead?

#
# Per-host flags and exclusions
#
ifneq ($(ACPI_HOST), _FreeBSD)
    CWARNINGFLAGS += \
        -Wempty-body

    ifneq ($(ACPI_HOST), _APPLE)
        CWARNINGFLAGS += \
            -Woverride-init\
            -Wtype-limits
        ifneq ($(ACPI_HOST), clang)
                -Wlogical-op\
                -Wmissing-parameter-type\
                -Wold-style-declaration\
        endif
    endif
endif

tbaederr:
Is your question about the branching or about checking `ACPI_HOST` for `clang`?

I simply took all the flags that clang doesn't support and only add them when `CC` is clang. I'm not sure what the intention of the `_APPLE` check is in any case, does that really depend on the host and not on the compiler?

If you think your version is better, I can change it.

SchmErik:
Oh yeah, I meant to say ifneq ($(CC), clang). My question was intended to be about branching.

@tbaederr you bring up a good point. I think we might want to move away from using compiler flags based on ACPI_HOST and use $(CC) or a gcc version instead. @acpibob what do you think?

acpibob:
Yes, if $(CC) will work consistently. No to using a gcc version, as we have customers that don’t use gcc.

tbaederr:
The best option would be to just check if the compiler supports the arguments we pass, but I'm not sure how to properly do that in plain makefiles.

Checking `CC` for clang/gcc should work in the general case, but I know that on arm the executable name can be e.g. armv7hl-redhat-linux-gnueabi-gcc. So we could check for `*gcc*` instead. What do you think?

nickdesaulniers:
> The best option would be to just check if the compiler supports the arguments we pass, but I'm not sure how to properly do that in plain makefiles.

The linux kernel defines a set of macros for doing such feature detection. `cc-option` for instance is how it checks for command line flags.  See:

https://github.com/ClangBuiltLinux/linux/blob/0fa8ee0d9ab95c9350b8b84574824d9a384a9f7d/scripts/Kbuild.include#L115-L124

for the macro definition, and

https://github.com/ClangBuiltLinux/linux/blob/0fa8ee0d9ab95c9350b8b84574824d9a384a9f7d/Makefile#L665

for an example invocation/expansion.  Needless to say, you don't get this for free in GNU `make`; you need to implement the macros yourself; but it's not impossible, and fairly concise.

tbaederr:
New version is up, which works for me.

generate/unix/Makefile.config
```diff
@@ -227,22 +227,40 @@ CWARNINGFLAGS += \
...
-            -Wtype-limits
-    endif
-endif
+CWARNINGFLAGS += $(call cc-option,-Wempty-body)
+CWARNINGFLAGS += $(call cc-option,-Wlogical-op)
+CWARNINGFLAGS += $(call cc-option,-Wmissing-parameter-type)
+CWARNINGFLAGS += $(call cc-option,-Wold-style-declaration)
+CWARNINGFLAGS += $(call cc-option,-Woverride-init)
+CWARNINGFLAGS += $(call cc-option,-Wtype-limits)
+CWARNINGFLAGS += $(call cc-option,-Wtype-clablabaldf)
```
nickdesaulniers, line 263: probably didn't mean to commit this last one, right?

nickdesaulniers (changes requested):

generate/unix/Makefile.config
tbaederr: Heh, no. But shows that it works. Fixed, thanks.

nickdesaulniers (approved):

acpibob:
Ok, I see you have already made the pull request. We will probably take it.

generate/unix/Makefile.config
```diff
@@ -227,22 +227,39 @@ CWARNINGFLAGS += \
     -Wold-style-definition\
     -Wno-format-nonliteral\
     -Wredundant-decls
+
+# output directory for tests below
 #
-# Per-host flags and exclusions
+# Taken from the Linux kernel
```
jrtc27, line 233: The Linux kernel build system is very clearly marked `SPDX-License-Identifier: GPL-2.0-only`

generate/unix/Makefile.config
```diff
@@ -227,22 +227,39 @@ CWARNINGFLAGS += \
...
+TMPOUT = $(if $(KBUILD_EXTMOD),$(firstword $(KBUILD_EXTMOD))/).tmp_$$$$
```
jrtc27, line 234: KBUILD_EXTMOD does not make any sense outside of the Linux kernel's build system

rafaeljw:
Given the most recent comments, I would expect a new version of this PR to be submitted.  Can be closed.

acpibob:

PR #650, opened 2020-11-13
ACPICA: fix -Wfallthrough

nickdesaulniers:
GCC 7.1 gained -Wimplicit-fallthrough to warn on implicit fallthrough,
as well as __attribute__((__fallthrough__)) and comments to explicitly
denote that cases of fallthrough were intentional. Clang also supports
this warning and statement attribute, but not the comment form.

Robert Moore provides additional context about the lint comments being
removed. They were for "an old version of PC-Lint, which we don't use
anymore." Drop those.

This will help us enable -Wimplicit-fallthrough throughout the Linux
kernel.

Suggested-by: Robert Moore <robert.moore@intel.com>
Signed-off-by: Nick Desaulniers <ndesaulniers@google.com>

### DIFF
--- a/source/compiler/aslerror.c
+++ b/source/compiler/aslerror.c
@@ -1456,7 +1456,7 @@ AslIsExceptionDisabled (
         {
             return (TRUE);
         }
-        /* Fall through */
+        ACPI_FALLTHROUGH;
 
     case ASL_WARNING:
     case ASL_REMARK:
--- a/source/compiler/aslwalks.c
+++ b/source/compiler/aslwalks.c
@@ -516,7 +516,7 @@ AnOperandTypecheckWalkEnd (
                     break;
                 }
 
-            /* Fallthrough */
+                ACPI_FALLTHROUGH;
 
             case ARGI_STORE_TARGET:
 
--- a/source/compiler/dtfield.c
+++ b/source/compiler/dtfield.c
@@ -227,7 +227,7 @@ DtCompileOneField (
             break;
         }
 
-        /* Fall through. */
+        ACPI_FALLTHROUGH;
 
     case DT_FIELD_TYPE_BUFFER:
 
--- a/source/components/disassembler/dmwalk.c
+++ b/source/components/disassembler/dmwalk.c
@@ -400,7 +400,7 @@ AcpiDmBlockType (
             return (BLOCK_NONE);
         }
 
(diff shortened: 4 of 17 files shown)


nickdesaulniers:
cc @acpibob 

attaching equivalent kernel patch FWIW.
[0001-ACPICA-fix-Wfallthrough.patch.txt](https://github.com/acpica/acpica/files/5539518/0001-ACPICA-fix-Wfallthrough.patch.txt)

SchmErik:
looks good to me

acpibob:
What is the point of the "do while"?

#define ACPI_FALLTHROUGH do {} while(0)

SchmErik:
It's basically a no-op. Here's a good read: http://www.bruceblinn.com/linuxinfo/DoWhile.html#:~:text=You%20may%20see%20a%20do,single%20statement%20can%20be%20used.

nickdesaulniers:
We don't need `__attribute__((__fallthrough__))` within `if ()` conditionals; I can remove it, but this definition matches the definition within the Linux kernel.

acpibob:
Well, ACPI_FALLTHROUGH is not intended to be used within if() conditionals, so I don't see much point in having the do...while.  However, I can live with it.

acpibob:
Nick,
While we are at it, can you possibly think of a macro solution to this problem:
ACPICA bugzilla #1520:
ACPICA: force compiler error on GCC when return_ACPI_STATUS is not paired with ACPI_FUNCTION_NAME or TRACE.

nickdesaulniers:
https://bugs.acpica.org/show_bug.cgi?id=1520

Sounds like there's macro's the should be paired, and is error prone to forget one when the other is used.

In that case, you can pass a block as a macro parameter. It would be good to make one macro that unconditionally does the initialization and cleanup unconditionally, and force the existing sites to use that.  For example, it's possible to emulate python's `with` statements in C via preprocessor statements (I can't find the blog where I originally saw that).  This issue is closely related to RAII in C++.

https://godbolt.org/z/rj56E5 for example where `WITH` could be `ACPI_FUNCTION_TRACE`.  That would be less prone to human error around forgetting the startup or cleanup.

acpibob:
Sounds interesting. I’ll look at after next week, since I’ll be out until then.
Thanks,
Bob

PR #654, opened 2020-12-07
Fix UB from arithmetic with null pointer

rokob:
If AmlStart is null then AmlStart + AmlLength is technically undefined
behavior. I saw this when running UBSAN with this code. If AmlStart is
null then end should be null as well.


acpibob:
Well, I doubt very much that AmlStart can be NULL, given that all the callers of this function have already validated it.

tamird:
We're seeing this same UB in Fuchsia. I sent #845 which is slightly different.

PR #656, opened 2020-12-22
Possible index out of range when resolving operands

ahs3:
UBSAN in the Linux kernel reports the following:

UBSAN: Undefined behaviour in drivers/acpi/acpica/dswexec.c:394:12

The line number does not match upstream due the age of the version
of ACPICA in use.  However, that turns out to be these lines:

            Status = AcpiExResolveOperands (WalkState->Opcode,
                &(WalkState->Operands [WalkState->NumOperands -1]),
                WalkState);

When WalkState->NumOperands is used, if by any chance it is zero,
we will end up with an index of (-1) into WalkState->Operands, which
is incorrect.  Put a check around this so that if for some reason
we do end up with zero operands, we do not try to resolve things
that do not exist.

Signed-off-by: Al Stone <ahs3@redhat.com>

### DIFF
--- a/source/components/dispatcher/dswexec.c
+++ b/source/components/dispatcher/dswexec.c
@@ -575,9 +575,11 @@ AcpiDsExecEndOp (
         {
             /* Resolve all operands */
 
-            Status = AcpiExResolveOperands (WalkState->Opcode,
-                &(WalkState->Operands [WalkState->NumOperands -1]),
-                WalkState);
+	    Status = AE_OK;
+	    if (WalkState->NumOperands > 0)
+            	Status = AcpiExResolveOperands (WalkState->Opcode,
+                	 &(WalkState->Operands [WalkState->NumOperands -1]),
+                	 WalkState);
         }
 
         if (ACPI_SUCCESS (Status))


acpibob:
I think this may be a case where WalkState->NumOperands can't be zero, as it was checked earlier.

ahs3:
> I think this may be a case where WalkState->NumOperands can't be zero, as it was checked earlier.

I don't see where; do you have a pointer?

What I see is that WalkState->NumOperands is initialized to zero in AcpiDsExecEndOp(), then passed into AcpiDsCreateOperands() that does not appear to change it.

acpibob:
Well, clearly it gets modified since it works (at least most of the time).

ahs3:
ping?

Will this patch be accepted, rejected, or does it need to be rewritten?  Best I can tell from further examination, UBSAN is indeed finding a potential error, and this does not appear to be a false positive.  Is it rare?  Possibly.  Can it occur?  I believe so.

Just let me know what the plan is, please, so I can figure out what I need to do next.

Thanks.

acpibob:
We are busy working on ACPI 6.4 support in ACPICA, so it will be a little while at least.  Sorry for the delay.
Bob

ahs3:
Gentle ping ... do I need to carry this patch separately or will it go into the main branch?

Thanks.  Just need to know what I need to do on my end.

acpibob:
We haven’t had the time to fully evaluate/review this patch. It will probably go into the main branch soon.
Bob

rafaeljw (commented):
IMV, this change would be hiding the problem which should at least be warned about.

Also, should AcpiDsCreateOperand() be called at all if NumOperands is 0?

rafaeljw:
This pull request is questionable and the question has not been responded since Dec 9, 2021.

rafaeljw:
Also it can be regarded as superseded by #745.  I think it can be closed.

acpibob:
This was handled in another pull request, closing

PR #657, opened 2021-01-04
Old tables

ahs3:
The MTMR and VRTC tables are no longer in use and are not defined in the ACPI specification.  Remove them from ACPICA so they can no longer be used.


ahs3:
None that I have found. The code was deprecated in Linux a couple of releases back and there have been no objections that I know of.

acpibob:
OK, I think we can take this.
thanks,
bob

PR #658, opened 2021-01-06
Generic serial bus race fix

jwrdegoede:
The handling of the GenericSerialBus (I2C) and GPIO OpRegions in AcpiEvAddressSpaceDispatch() passes a number of extra parameters to the address-space handler through the address-space Context pointer (instead of using more function parameters).

The Context is shared between threads, so if multiple threads try to call the handler for the same address-space at the same time, then a second thread could change the parameters of a first thread while the handler is running for the first thread.

An example of this race hitting is the Lenovo Yoga Tablet2 1015L, where there are both AttribBytes accesses and AttribByte accesses to the same address-space. The AttribBytes access stores the number of bytes to transfer in Context->AccessLength. Where as for the AttribByte access the number of bytes to transfer is always 1 and FieldObj->Field.AccessLength is unused (so 0). Both types of accesses racing from different threads leads to the following problem:

1. Thread a. starts an AttribBytes access, stores a non 0 value from FieldObj->Field.AccessLength in Context->AccessLength
2. Thread b. starts an AttribByte access, stores 0 in Context->AccessLength
3. Thread a. calls i2c_acpi_space_handler() (under Linux). Which sees that the access-type is ACPI_GSB_ACCESS_ATTRIB_MULTIBYTE and calls acpi_gsb_i2c_read_bytes(..., Context->AccessLength)
4. At this point Context->AccessLength is 0 (set by thread b.) rather then the FieldObj->Field.AccessLength value from thread a. This 0 length reads leads to the following errors being logged:

```
 i2c i2c-0: adapter quirk: no zero length (addr 0x0078, size 0, read)
 i2c i2c-0: i2c read 0 bytes from client@0x78 starting at reg 0x0 failed, error: -95

```
Note this is just an example of the problems which this race can cause. There are likely many more (sporadic) problems caused by this race.

This commit adds a new ContextMutex to acpi_object_addr_handler and makes AcpiEvAddressSpaceDispatch() take that mutex when using the shared Context to pass extra parameters to an address-space handler, fixing this race.

Note the new mutex must be taken *after* exiting the interpreter, therefor the existing AcpiExExitInterpreter() call is moved to above the code which stores the extra parameters in the Context.

### DIFF
--- a/source/components/events/evhandler.c
+++ b/source/components/events/evhandler.c
@@ -677,6 +677,13 @@ AcpiEvInstallSpaceHandler (
 
     /* Init handler obj */
 
+    Status = AcpiOsCreateMutex (&HandlerObj->AddressSpace.ContextMutex);
+    if (ACPI_FAILURE (Status))
+    {
+        AcpiUtRemoveReference (HandlerObj);
+        goto UnlockAndExit;
+    }
+
     HandlerObj->AddressSpace.SpaceId = (UINT8) SpaceId;
     HandlerObj->AddressSpace.HandlerFlags = Flags;
     HandlerObj->AddressSpace.RegionList = NULL;
--- a/source/components/events/evregion.c
+++ b/source/components/events/evregion.c
@@ -270,6 +270,8 @@ AcpiEvAddressSpaceDispatch (
     ACPI_OPERAND_OBJECT     *RegionObj2;
     void                    *RegionContext = NULL;
     ACPI_CONNECTION_INFO    *Context;
+    ACPI_MUTEX              ContextMutex;
+    BOOLEAN                 ContextLocked;
     ACPI_PHYSICAL_ADDRESS   Address;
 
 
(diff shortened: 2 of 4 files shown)


jwrdegoede:
ping? It would be nice to get this merged, and then added to the 5.11-rc# linux kernel series as this fixes a pretty serious bug.

acpibob:
We have to get around to looking at it.
Thanks,
Bob

SchmErik:
@acpibob you had a concern that adding objects to the ACPI_OBJECT union was going to cause it to become larger. I'm just writing this as a note to you (and myself) that @jwrdegoede adding a mutex is not going to exceed the max size of the union. Here's the output from acpiexec with his change:

```
Common            16
Number            24
String            32
Buffer            48
Package           48
BufferField       48
Device            48
Event             24
Method            56
Mutex             72
Region            56
PowerResource     48
Processor         48
ThermalZone       40
RegionField       72
BankField         56
IndexField        56
Reference         64
Notify            64
AddressSpace      72
Extra             56
Data              32
```

acpibob:
Ok, good. Thanks.

SchmErik:
@acpibob this looks good to me

PR #660, opened 2021-01-15
gitignore: split gitignore file for easier maintenance

ElyesH:
Split ".gitignore" by directory for easier maintenance.

Signed-off-by: Elyes HAOUAS <ehaouas@noos.fr>


SchmErik:
we don't see how splitting one file into many separate files would facilitate maintenance so we will not be taking this. Closing...

PR #663, opened 2021-02-02
Fix for github issue 659

SchmErik:
This pull request contains fixes for github issue 659 by changing iASL behavior to keep external methods that are referenced in iASL constructs such as ```if (CondRefOf (EXT0))```

### DIFF
--- a/source/compiler/aslexternal.c
+++ b/source/compiler/aslexternal.c
@@ -324,49 +324,52 @@ ExInsertArgCount (
 
         NameOp = Next->Asl.Child->Asl.Child;
         ExternalName = AcpiNsGetNormalizedPathname (NameOp->Asl.Node, TRUE);
-
         if (strcmp (CallName, ExternalName))
         {
             ACPI_FREE (ExternalName);
             Next = Next->Asl.Next;
             continue;
         }
-
-        Next->Asl.Child->Asl.CompileFlags |= OP_VISITED;
-
-        /*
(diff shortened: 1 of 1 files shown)


1alessandro1:
First, thanks for the PR and for your time and dedication

However, I compiled the latest binaries with this PR merged and the issue persists.
I'll contact foskvs if it's the same on his end

SchmErik:
@1alessandro1 could you post the file that exhibits this issue? I tried to compile the tables posted by @foskvs in the issues page but I'm not seeing the compiler error.

SchmErik:
I was a little confused by what @foskvs said but I see the compiler error now. I've fixed the compiler so that it can allow iASL to compile, disassemble, and recompile the code in question. However, if you're doing the disassembly and re-compilation on tables that have been initially compiled other than this version of the compiler, then you'll still get the error because the external method declarations have been omitted in the initial compilation.

1alessandro1:
@SchmErik Please test this (compile, no errors, then decompile, and recompile)

[SSDT-EC-USBX-DESKTOP.dsl.zip](https://github.com/acpica/acpica/files/5914822/SSDT-EC-USBX-DESKTOP.dsl.zip)

SchmErik:
ah, I made a slight mistake and I fixed it just now. Could you try fetching and re-compiling? I force-pushed so you might need to do something like git reset --hard

1alessandro1:
perfect, the issue is gone

PR #666, opened 2021-02-18
Add VIOT table

jpbrucker:
Add support for the VIOT table to iasl and acpisrc.

The Virtual I/O Translation Table (VIOT) describes the topology of para-virtual I/O translation devices (currently virtio-iommu) and the endpoints they manage. It was recently approved for inclusion into the standard.

A provisional version of the specification can be found [here](https://jpbrucker.net/virtio-iommu/viot/viot-v9.pdf). Most of the code was copied from similar table implementations (IORT and IVRS).

As an example, the VIOT template converted to aml and then dsl looks like: [viot.txt](https://github.com/acpica/acpica/files/6004956/viot.txt)


SchmErik:
Thanks for this pull request. I haven't looked at it yet but it seems very comprehensive! We'll take a look after the next release.. Thanks

PR #669, opened 2021-02-25
Add CEDT support (from CXL 2.0)

bwidawsk:
The last patch would need to be replaced, though I'm not sure how to generate a good template.

I've opted not to include the compilation bits as I'm currently using QEMU to generate the tables and have no need. They can be added as follow on patches.


bwidawsk:
What is needed to get this merged?

acpibob:
Nothing at this time. We have been busy implementing ACPI 6.4 support, so some requests have been put on hold. Sorry for the delay.

acpibob:
Where is   DtCompileCedt?

bwidawsk:
> Where is DtCompileCedt?

We discussed this internally. We do not need it for our needs, so it was left as a TODO for someone else or when we have more time. Would you like me to add an issue for it?

acpibob:
No need to add an issue, we’ll do it.
Where can we find the spec for CEDT?

bwidawsk:
https://www.computeexpresslink.org/download-the-specification

If you don't feel like signing up, I can send you a copy of the spec internally since Intel is a member of the consortium.

acpibob:
Yes, please send the spec to me.
Thx,
Bob

PR #671, opened 2021-03-02
Revert "Interpreter: fix memory leak by using use existing buffer in …

SchmErik:
…_HID repair"

This reverts commit 52d1da5dcbd79a722b70f02a1a83f04088f51ff6.

The 'exisitng buffer' in this case is the firmware provided table, and
we should not modify that in place. This fixes a crash on arm64 with
initrd table overrides, in which case the DSDT is not mapped with
read/write permissions.

Cc: Robert Moore <robert.moore@intel.com>
Cc: Erik Kaneda <erik.kaneda@intel.com>
Cc: "Rafael J. Wysocki" <rafael.j.wysocki@intel.com>
Cc: Len Brown <lenb@kernel.org>
Reported-by: Shawn Guo <shawn.guo@linaro.org>
Signed-off-by: Ard Biesheuvel <ardb@kernel.org>

### DIFF
--- a/source/components/namespace/nsrepair2.c
+++ b/source/components/namespace/nsrepair2.c
@@ -690,8 +690,9 @@ AcpiNsRepair_HID (
     ACPI_OPERAND_OBJECT     **ReturnObjectPtr)
 {
     ACPI_OPERAND_OBJECT     *ReturnObject = *ReturnObjectPtr;
-    char                    *Dest;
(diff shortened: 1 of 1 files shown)


gwr:
BTW, this also reopens the leak bug fixed in:
https://github.com/acpica/acpica/pull/647

Is that being tracked anywhere?  I don't see an issue for it.

PR #682, opened 2021-03-25
Master iort rev e.b

shamiali2008:
Pull requests to revert the Rev E merge and update it with E.b revision changes.

Please find the E.b spec here,
https://developer.arm.com/documentation/den0049/latest/

Thanks,
Shameer


acpibob:
Thanks.

PR #694, opened 2021-05-20
Add more structs to CEDT table  

SchmErik:
This pull request adds a new sub table along with several defines

### DIFF
--- a/source/include/actbl1.h
+++ b/source/include/actbl1.h
@@ -526,6 +526,15 @@ enum AcpiCedtType
     ACPI_CEDT_TYPE_RESERVED             = 1
 };
 
+/* Values for version field above */
+
+#define ACPI_CEDT_CHBS_VERSION_CXL11    (0)
+#define ACPI_CEDT_CHBS_VERSION_CXL20    (1)
+
+/* Values for length field above */
+
+#define ACPI_CEDT_CHBS_LENGTH_CXL11     (0x2000)
+#define ACPI_CEDT_CHBS_LENGTH_CXL20     (0x10000)
 
 /*
  * CEDT subtables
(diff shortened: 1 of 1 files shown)


source/include/actbl1.h
```diff
@@ -523,7 +523,8 @@ typedef struct acpi_cedt_header
 enum AcpiCedtType
 {
     ACPI_CEDT_TYPE_CHBS                 = 0,
-    ACPI_CEDT_TYPE_RESERVED             = 1
```
kkbodke, line 526: Do we need this?
ECN does not define Reserved as 1.

kkbodke (commented):
Is this required?  ECN does not define this.
ACPI_CEDT_TYPE_RESERVED             = 1

source/include/actbl1.h
AlisonSchofield: I'm not aware that it was RESERVED before it became the TYPE_CFMWS in the ECN. So, I don't think we need to add back a 'ACPI_CEDT_TYPE_RESERVED' in the enum.

SchmErik: Oops, I forgot to mention this point in internal during internal review.

We typically use this "reserved" enum to mean that a subtable type greater than or equal to the value of ACPI_xxxx_TYPE_RESERVED is undefined. A lot of ACPI table specifications state that the values for subtable ID's that are yet to be defined are "reserved" for future usage. We use these defines in our table compiler to detect any strange subtables that we do not yet have definitions for or ACPI tables that contained undefined structures.

kkbodke: ACPI_CEDT_TYPE_RESERVED should be "2"
1 is for CFMWS only.

SchmErik: Yes. That's why I added the line ACPI_CEDT_TYPE_RESERVED = 2 on line 527. Before the CFMWS was added, the subtype 1 was not defined so that's why ACPI_CEDT_TYPE_RESERVED was defined as 1. This patch moves the ACPI_CEDT_TYPE_RESERVED from 1 to 2. and defines ACPI_CEDT_TYPE_CFMWS to 1.

PR #699, opened 2021-06-01
Disassembler: add missing strings to decode subtable types

SchmErik:


acpibob:
got it, thanks.

PR #720, opened 2021-10-03
Use original table pointers rather than round-tripping through ACPI_PHYSICAL_ADDRESS

jrtc27:
Currently there are two sets of cases where pointers to tables are cast to ACPI_PHYSICAL_ADDRESS and later cast back to a pointer to be dereferenced. Whether or not this is supported is implementation-defined.

On CHERI, and thus Arm's experimental Morello prototype architecture, pointers are represented as capabilities, which are unforgeable bounded pointers, providing always-on fine-grained spatial memory safety. This means that any pointer cast to a plain integer will lose all its associated metadata, and when cast back to a pointer it will give a null-derived pointer (one that has the same metadata as null but an address equal to the integer) that will trap on any dereference. As a result, this is an implementation where ACPI_PHYSICAL_ADDRESS cannot be used as a hack to store real pointers.

This patch series propagates the original, real pointers throughout the internals to allow ACPICA to support such more-restrictive implementations.


jrtc27:
Ping?

acpibob:
got it, thanks.

jrtc27:
Thanks!

PR #723, opened 2021-10-08
ACPICA: drop unneeded initialization value

smalinux:
Do not initialize statics to NULL

### DIFF
--- a/source/components/debugger/dbcmds.c
+++ b/source/components/debugger/dbcmds.c
@@ -193,7 +193,7 @@ AcpiDbDoOneSleepState (
     UINT8                   SleepState);
 
 
-static char                 *AcpiDbTraceMethodName = NULL;
+static char                 *AcpiDbTraceMethodName;
 
 
 /*******************************************************************************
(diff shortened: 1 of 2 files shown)


rafaeljw (changes requested):
First off, the changelog only mentions NULL, but the patch changes other cases too.

Second, there are more cases when static vars are initialized to 0 / NULL in the code base, so why not to address them all at once?

rafaeljw:
> My concern with this is that the apparent assumption is that data areas are always initialized to zero –
> which may not be the case on all hosted operating systems.

static/extern variables in C must be initialized to 0 by default as per the language definition and this behavior is relied on quite a bit.

rafaeljw:
Another question is whether or not NULL must be zero, though, and quite frankly I'm not sure, so it probably is prudent to initialize things to NULL (in case it is not 0) explicitly.

acpibob:
So, I think we should ask the poster to finish all of the 0/NULL initializations across ACPICA.

rafaeljw:
I have already done that (I think).

acpibob:
Have not received anything more from the O.P., closing.

PR #729, opened 2021-10-28
Fix AEST Processor generic resource substructure data field byte length

ShuuichirouIshii:
The byte length of the Data field in the AEST Processor generic resource
substructure defined in ACPI for the Armv8 RAS Extensions 1.1 is 4Byte.
However, it is defined as a pointer type, and on a 64-bit machine,
it is interpreted as 8 bytes. Therefore, it is changed from a pointer
type unsigned integer 1 byte to an unsigned integer 4 bytes.

Signed-off-by: Shuuichirou Ishii <ishii.shuuichir@fujitsu.com>

### DIFF
--- a/source/include/actbl2.h
+++ b/source/include/actbl2.h
@@ -309,7 +309,7 @@ typedef struct              acpi_aest_processor_tlb
 
 typedef struct              acpi_aest_processor_generic
 {
-    UINT8                   *Resource;
+    UINT32                   Resource;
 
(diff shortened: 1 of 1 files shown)


ShuuichirouIshii:
ping?

If this fix is not applied, the acpidump command will error when parsing AEST.
Thank you for your consideration.

acpibob:
We will get this for the next ACPICA release, probably later this month.

PR #735, opened 2021-11-22
Pcc opregion support

sudeep-holla:
Hi,

The first change is based on the fact that no one has reported any usage of PCC type 1/2 opregion,
One of the platform I use has PCC type 3 Opregion and is currently crashing at boot with log as in the commit.
So it needs to fixed ASAP.


sudeep-holla:
For reference on how these changes will be used in the Linux kernel:
https://lore.kernel.org/lkml/20211102182542.3460787-4-sudeep.holla@arm.com/

rafaeljw (approved):

rafaeljw (approved):

PR #744, opened 2022-01-14
Headers: Use uintptr_t and offsetof() in Linux kernel builds

rafaeljw:
This is based on the Linux kernel patch at

https://patchwork.kernel.org/project/linux-acpi/patch/2614912.mvXUDI8C0e@kreacher/

To avoid "performing pointer subtraction with a null pointer has
undefined behavior" compiler warnings, use uintptr_t and offsetof()
that are always available during Linux kernel builds to define
ACPI_UINTPTR_T and the ACPI_TO_INTEGER() and ACPI_OFFSET() macros.

### DIFF
--- a/source/include/actypes.h
+++ b/source/include/actypes.h
@@ -658,8 +658,12 @@ typedef UINT64                          ACPI_INTEGER;
 /* Pointer/Integer type conversions */
 
 #define ACPI_TO_POINTER(i)              ACPI_CAST_PTR (void, (ACPI_SIZE) (i))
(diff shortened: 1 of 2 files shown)


tamird:
How come this applies only to Linux *kernel* builds?

SaketADumbre:
> How come this applies only to Linux _kernel_ builds?

@tamird this comes from Linux to us originally, so it was named like that by @rafaeljw , who is the Power Management maintainer of the Linux kernel!

PR #751, opened 2022-02-18
Headers: Replace zero-length array with flexible-array member

rafaeljw:
Replace zero-length array with flexible-array member in the struct acpi_table_rgrt definition.


acpibob:
OK, got it. Thanks, Bob

PR #752, opened 2022-02-21
IORT: Updates for revision E.d

shamiali2008:
IORT revision is now updated to E.d (ARM DEN 0049E.d) and
contains a few additions like,
    -Added descriptor in the root complex node for specifying
     PASID width supported by the root complex.
    -Updated RMR node Flags field.
    -Introduced memory access attributes in the RMR node.

Please note that IORT Rev E.c is deprecated and not supported.


shamiali2008:
The IORT Rev E.d doc can be found here,
https://developer.arm.com/documentation/den0049/ed/

Thanks,
Shameer

acpibob:
We would like these defines to be all upper case.

For example:


+#define ACPI_IORT_RMR_ATTR_DEVICE_nGnRnE   0x00

shamiali2008:
This is the way normally device memory is represented in Arm specification
documents and kernel as well.
eg:
https://elixir.bootlin.com/linux/v4.5/source/arch/arm64/include/asm/pgtable.h#L72

So would like to keep that to avoid any confusion, if it is Ok.

Thanks,
Shameer

shamiali2008:
Yes, Ideally would like to keep the lower case. 
But if ACPICA convention is to capitalise #defines, I am not sure what is the best way here.

@rmurphy-arm  Any thoughts? Can we go with ACPICA convention here?

Thanks.

rmurphy-arm:
Sure, there's no need to break convention - ACPICA isn't bound by Arm's technical writing guidelines, after all. I'm just keen that whatever ACPICA's readable identifier for value 0x5 is, it explicitly captures that it's Write-Back Cacheable. We don't really expect to need yet more Normal memory types with further varieties of cacheablility, but we've intentionally left room to define additional values later if people do demand them.

shamiali2008:
Ok. I will close this one and submit another one taking care of all the above comments.

Thanks.

PR #757, opened 2022-03-09
ACPI: Add LoongArch-related definitions

chenhuacai:
LoongArch is a new RISC ISA, which is a bit like MIPS or RISC-V.
LoongArch includes a reduced 32-bit version (LA32R), a standard 32-bit
version (LA32S) and a 64-bit version (LA64). LoongArch use ACPI as its
boot protocol LoongArch-specific interrupt controllers (similar to APIC)
are already added in the next revision of ACPI Specification (current
revision is 6.4).

This patchset are preparing to add LoongArch support in mainline kernel,
we can see a snapshot here:
https://github.com/loongson/linux/tree/loongarch-next

Cross-compile tool chain to build kernel:
https://github.com/loongson/build-tools/releases

Loongson and LoongArch documentations:
https://github.com/loongson/LoongArch-Documentation

ECR for LoongArch-specific interrupt controllers:
https://mantis.uefi.org/mantis/view.php?id=2203
https://mantis.uefi.org/mantis/view.php?id=2313

ACPI changes of LoongArch have been approved in the last year, but the
new version of ACPI SPEC hasn't been made public yet.


acpibob:
You need to add new strings for the new defines in actbl2.h, in dmtable.c, struct named AcpiDmMadtSubnames.

chenhuacai:
> You need to add new strings for the new defines in actbl2.h, in dmtable.c, struct named AcpiDmMadtSubnames.

OK, now it has been updated.

acpibob:
OK, got it, thanks.
bob

source/components/events/evevent.c
```diff
@@ -312,7 +312,8 @@ AcpiEvFixedEventInitialize (
         {
             Status = AcpiWriteBitRegister (
                 AcpiGbl_FixedEventInfo[i].EnableRegisterId,
-                ACPI_DISABLE_EVENT);
+                (i == ACPI_EVENT_PCIE_WAKE) ?
+		ACPI_ENABLE_EVENT : ACPI_DISABLE_EVENT);
```
paulmenzel, line 316: I think spaces instead of tabs should be used. Please create a fix-up patch.

chenhuacai: > I think spaces instead of tabs should be used. Please create a fix-up patch.

OK, https://github.com/acpica/acpica/pull/767 created, thanks.

lvjianmin-loongson:
Please revert the patch("Events: Support fixed pcie wake event") in this PR, the related patch of which in linux kernel broke wake-on-lan for multiple machines, and has been reverted.

Thanks.

PR #765, opened 2022-03-28
IORT: Updates for revision E.d

shamiali2008:
IORT revision is now updated to E.d (ARM DEN 0049E.d) and
contains a few additions like,
    -Added descriptor in the root complex node for specifying
     PASID width supported by the root complex.
    -Updated RMR node Flags field.
    -Introduced memory access attributes in the RMR node.

Please note that IORT Rev E.c is deprecated and not supported.


acpibob:
Ok, got it.
thanks,
bob

PR #767, opened 2022-04-03
Events: Use spaces instead of tabs to fix indentation

chenhuacai:
Signed-off-by: Huacai Chen <chenhuacai@loongson.cn>


paulmenzel (approved):
I think, Signed-off-by line is not needed.

chenhuacai:
> I think, Signed-off-by line is not needed.

Should I update my PR? And I found that there are many Signed-off-by in git log.

paulmenzel:
> > I think, Signed-off-by line is not needed.
> 
> Should I update my PR? And I found that there are many Signed-off-by in git log.

Yes, I saw that too, but the main developer @acpibob, does not sign off their commits, so I think it is not needed. I have no submit rights, so feel free to wait for other reviews.

acpibob:
thanks, got it.

PR #773, opened 2022-05-06
Constify AcpiGetHandle pathname argument

s-ailus:
Hello,

This pull request contains a single patch, constifying the pathname argument to AcpiGetHandle. The objective is to allow passing a const pathname to the function, in order to avoid having to make a copy of it. This isn't needed by acpica per se, but by another caller in Linux.

It's the first time I'm using Github to do this, let's see how this works...

- Sakari

### DIFF
--- a/source/components/namespace/nsxfname.c
+++ b/source/components/namespace/nsxfname.c
@@ -192,7 +192,7 @@ AcpiNsCopyDeviceId (
 ACPI_STATUS
 AcpiGetHandle (
     ACPI_HANDLE             Parent,
-    ACPI_STRING             Pathname,
+    const char              *Pathname,
     ACPI_HANDLE             *RetHandle)
 {
     ACPI_STATUS             Status;
--- a/source/include/acpixf.h
+++ b/source/include/acpixf.h
@@ -770,7 +770,7 @@ ACPI_EXTERNAL_RETURN_STATUS (
 ACPI_STATUS
 AcpiGetHandle (
     ACPI_HANDLE             Parent,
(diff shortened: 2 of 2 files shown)


rafaeljw (approved):

acpibob:
I'm worried that there are lots of these sprinkled throughout ACPICA - and this fix could open the door to lots of changes.

s-ailus:
Hi Robert,

On Thu, Aug 18, 2022 at 09:15:14AM -0700, Robert Moore wrote:
>    I'm worried that there are lots of these sprinkled throughout ACPICA - and
>    this fix could open the door to lots of changes.

Could you elaborate your concern a little? Is it that there are other
functions that do not write to the data behind a pointer taken as an
argument which could be const but is not, or something else? Const is being
used in acpica already, albeit perhaps not everywhere it could be.

The alternatives for this change are making the large data structure (a
single memory allocation) not const or allocating a temporary buffer for
the path name, checking the allocation succeeded and finally copying the
const pathname there before calling the function.

Both are poor workarounds compared to fixing the root issue.

-- 
Regards,

Sakari Ailus

s-ailus:
Ping.

rafaeljw:
> I'm worried that there are lots of these sprinkled throughout ACPICA - and this fix could open the door to lots of changes.

Do you think that people will try to annotate pathname pointers all over with const?  Or something else?

rafaeljw:
I don't see a problem with AcpiGetHandle() in particular TBH and the other cases can be considered individually IMO.

PR #775, opened 2022-05-15
compiler/dttable1.c: Fix dead assignment

ElyesH:
The value stored to 'ParentTable' is never used.

Signed-off-by: Elyes Haouas <ehaouas@noos.fr>


acpibob:
So, should 'Parent Table' be fully removed from the source?

ElyesH:
maybe I'm wrong but looks like it is used elsewhere, isn't it?

PR #779, opened 2022-05-27
iASL: NHLT: Fix compilation of optional undocumented fields

pmaziarz:
Since those fields are optional some combinations of them were causing
unpredicted flow when compiling the table.


acpibob:
OK, this looks good, we will merge this.
Thanks,
bob

PR #782, opened 2022-06-15
Add support for FFH Opregion special context data

sudeep-holla:
FFH(Fixed Function Hardware) Opregion is approved to be added in ACPIC 6.5 via
code first approach[1]. It requires special context data similar to GPIO and
Generic Serial Bus as it needs to know platform specific offset and length.

Add support for the special context data needed by FFH Opregion.

FFH OpRegion enables advanced use of FFH on some architectures. For example,
it could be used to easily proxy AML code to architecture-specific behavior
(to ensure it is OS initiated)

Actual behavior of FFH is ofcourse architecture specific and depends on
the FFH bindings. The offset and length could have arch specific meaning
or usage.

[1] https://bugzilla.tianocore.org/show_bug.cgi?id=3598

Signed-off-by: Sudeep Holla <sudeep.holla@arm.com>

### DIFF
--- a/source/components/events/evregion.c
+++ b/source/components/events/evregion.c
@@ -329,6 +329,14 @@ AcpiEvAddressSpaceDispatch (
             Ctx->SubspaceId = (UINT8) RegionObj->Region.Address;
         }
 
+        if (RegionObj->Region.SpaceId == ACPI_ADR_SPACE_FIXED_HARDWARE)
+        {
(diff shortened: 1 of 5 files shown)


sudeep-holla:
Thanks Rafael, updated the branch with typo fixed.

rafaeljw (approved):

PR #790, opened 2022-08-18
Fix typo as per Issue #783

### LINKED ISSUE #783: Pls fix typo in /source/include/actbl2.h line 1533
OemData[0];  ->  OemData[];

satmandu:
Fixes #783 

I believe this is what @OS4User was talking about...

### DIFF
--- a/source/include/actbl2.h
+++ b/source/include/actbl2.h
@@ -1549,7 +1549,7 @@ enum AcpiMadtLpcPicVersion {
 
 typedef struct acpi_madt_oem_data
 {
-    UINT8                   OemData[0];
+    UINT8                   OemData[];
 } ACPI_MADT_OEM_DATA;
 
 


OS4User:
exactly, tnx a lot

rafaeljw (approved):

acpibob:
This change fails under gcc:
actbl2.h:1553:29: error: flexible array member in a struct with no named members
     UINT8                   OemData[]

Apparently, this change violates this rule (see "-->")

Flexible Array Member(FAM) is a feature introduced in the C99 standard of the C programming language.

For the [structures](https://www.geeksforgeeks.org/structures-c/) in C programming language from C99 standard onwards, we can declare an array without a dimension and whose size is flexible in nature.
Such an array inside the structure should preferably be declared as the last member of structure and its size is variable(can be changed be at runtime).
--> The structure must contain at least one more named member in addition to the flexible array member.

acpibob:
So, I will probably revert this change.

OS4User:
The reason for using ACPI_MADT_OEM_DATA structure is not quite clear. Perhaps it makes sense to completely remove it.
The alternative is to define OemData cardinality as [1] but this heavily depends on intended usage of this structure.

And
https://github.com/acpica/acpica/blob/master/source/include/actbinfo.h#L331
should probably look like this:
#define ACPI_MADT17_OFFSET(f) (UINT16) ACPI_OFFSET (ACPI_MADT_CORE_PIC,f)
at least that's how it looks like when one sees:
https://github.com/acpica/acpica/blob/master/source/include/actbl2.h#1421

Check this out
https://gcc.gnu.org/onlinedocs/gcc/Zero-Length.html
https://stackoverflow.com/questions/8625572/what-is-the-significance-of-a-0-length-array-in-c  (C++11, 8.3.4/1: "[the array size] shall be greater than zero")

acpibob:
I suspect that MADT_OEM_DATA is an optional field, and the length of this field is contained as part of the subtable->Length field.

OS4User:
"I suspect that MADT_OEM_DATA is an optional field, and the length of this field is contained as part of the subtable->Length field."

In this case, the structure should be declared as follows:

typedef struct acpi_madt_oem_data
{
    ACPI_SUBTABLE_HEADER    Header;
    UINT8                   OemData[];
} ACPI_MADT_OEM_DATA;

acpibob:
Can you go ahead and update the various support functions in the DT compiler (Table info, etc.)

Thanks,
Bob

PR #794, opened 2022-09-13
Headers: Delete bogus NodeArray array of pointers from AEST table

jrtc27:
Like many tables, this is a header followed by multiple subtables of
varying self-identifying types, and ACPICA does not normally add a field
for the subtables, instead relying on pointer arithmetic past the end of
the first header struct, since indexing a flexible array member is
meaningless for variable-length entries. If we really wanted a field for
this, we could use a UINT8 flexible array member, but it contradicts the
current style. Using void *, however, is categorically wrong, as ACPI
tables never contain native C-language pointers.

### DIFF
--- a/source/include/actbl2.h
+++ b/source/include/actbl2.h
@@ -229,7 +229,6 @@
 typedef struct acpi_table_aest
 {
     ACPI_TABLE_HEADER       Header;
-    void                    *NodeArray[];
 
 } ACPI_TABLE_AEST;
 


acpibob:
Well, in your patch, I don't see ACPICA (actually iASL) accessing the field past the header (NodeArray).

jrtc27:
> Well, in your patch, I don't see ACPICA (actually iASL) accessing the field past the header (NodeArray).

I don’t get your point? https://github.com/acpica/acpica/commit/e75074d84d1207339a048486c2d06ecb935d0092 added support for the table but included this bogus unused field. I’m just deleting it to correct that.

PR #811, opened 2022-11-11
nsrepair: handle cases without a return value correctly

8infy:
Previously AcpiNsSimpleRepair() would crash if ExpectedBtypes contained any combination of ACPI_RTYPE_NONE with a different type, e.g | ACPI_RTYPE_INTEGER because of slightly incorrect logic in the !ReturnObject branch, which wouldn't return AE_AML_NO_RETURN_VALUE for such cases.

Found by Linux Verification Center (linuxtesting.org) with the SVACE static analysis tool.

Signed-off-by: Daniil Tatianin <d-tatianin@yandex-team.ru>

### DIFF
--- a/source/components/namespace/nsrepair.c
+++ b/source/components/namespace/nsrepair.c
@@ -328,9 +328,10 @@ AcpiNsSimpleRepair (
      */
     if (!ReturnObject)
     {
-        if (ExpectedBtypes && (!(ExpectedBtypes & ACPI_RTYPE_NONE)))
+        if (ExpectedBtypes)
         {
-            if (PackageIndex != ACPI_NOT_PACKAGE_ELEMENT)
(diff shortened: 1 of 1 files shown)


rafaeljw (approved):
This change looks good to me, thanks!

8infy:
Thank you!
Is there anything else I have do before this can be merged?

rafaeljw:
If you want the fix to be present in the Linux kernel, in particular, please send the corresponding Linux patch to linux-acpi@vger.kernel.org with a Link tag pointing to this PR.

8infy:
> If you want the fix to be present in the Linux kernel, in particular, please send the corresponding Linux patch to [linux-acpi@vger.kernel.org](mailto:linux-acpi@vger.kernel.org) with a Link tag pointing to this PR.

Thanks!

Is there an example of what that would look like?
And how do I get this merged in ACPICA specifically as well?

PR #813, opened 2022-11-18
Replace fake flexible arrays with actual flexible array members

kees:
This continues earlier C99 work to modernize the code to work correctly with compile-time and run-time bounds checkers.

### DIFF
--- a/source/include/actbl2.h
+++ b/source/include/actbl2.h
@@ -1550,7 +1550,7 @@ enum AcpiMadtLpcPicVersion {
 
 typedef struct acpi_madt_oem_data
 {
-    UINT8                   OemData[];
+    ACPI_FLEX_ARRAY(UINT8, OemData);
 } ACPI_MADT_OEM_DATA;
 
 
--- a/source/include/acrestyp.h
+++ b/source/include/acrestyp.h
@@ -927,8 +927,10 @@ typedef struct acpi_pci_routing_table
     UINT32                          Pin;
     UINT64                          Address;        /* here for 64-bit alignment */
     UINT32                          SourceIndex;
-    char                            Source[4];      /* pad to 64 bits so sizeof() works in all cases */
-
+    union {
+                                    char Pad[4];    /* pad to 64 bits so sizeof() works in all cases */
+                                    ACPI_FLEX_ARRAY(char, Source);
+    };
 } ACPI_PCI_ROUTING_TABLE;
 
 #endif /* __ACRESTYP_H__ */
--- a/source/include/acrestyp.h
+++ b/source/include/acrestyp.h
@@ -538,7 +538,7 @@ typedef struct acpi_resource_extended_irq
     UINT8                           WakeCapable;
     UINT8                           InterruptCount;
     ACPI_RESOURCE_SOURCE            ResourceSource;
-    UINT32                          Interrupts[1];
+    UINT32                          Interrupts[];
 
 } ACPI_RESOURCE_EXTENDED_IRQ;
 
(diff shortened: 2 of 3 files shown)


acpibob:
OK, but I’m thinking that using these “new” structures will require changes that affect both the Data Table Compiler and the Data Table disassembler.
Bob

From: Kees Cook ***@***.***>
Sent: Thursday, November 17, 2022 8:56 PM
To: acpica/acpica ***@***.***>
Cc: Subscribed ***@***.***>
Subject: [acpica/acpica] Replace fake flexible arrays with actual flexible array members (PR #813)


This continues earlier C99 work to modernize the code to work correctly with compile-time and run-time bounds checkers.

________________________________
You can view, comment on, or merge this pull request online at:

  https://github.com/acpica/acpica/pull/813

Commit Summary

  *   84851f6<https://github.com/acpica/acpica/pull/813/commits/84851f6b3a9448c5f64a42e241302dcf147c6b10> ACPI_RESOURCE_EXTENDED_IRQ: Replace 1-element array with flexible array
  *   c892aca<https://github.com/acpica/acpica/pull/813/commits/c892aca0bf6b4f2a65b12176f1e38614c97313d7> ACPI_PCI_ROUTING_TABLE: Replace fixed-size array with flex array member

File Changes

(2 files<https://github.com/acpica/acpica/pull/813/files>)

  *   M source/include/acrestyp.h<https://github.com/acpica/acpica/pull/813/files#diff-0d6df2e054e0561eede97cac47d0cd3e005f62038a5fcc4f999b38ef4de9b694> (8)
  *   M source/include/platform/acenv.h<https://github.com/acpica/acpica/pull/813/files#diff-4d7183c49a06f644e22fd6921bbbc0282a4c641c7ca60b1d410031f843ac890c> (11)

Patch Links:

  *   https://github.com/acpica/acpica/pull/813.patch
  *   https://github.com/acpica/acpica/pull/813.diff

kees:
> OK, but I’m thinking that using these “new” structures will require changes that affect both the Data Table Compiler and the Data Table disassembler.

I don't know what this means. :) How can I find/test these?

acpibob:
It depends on the actual table(s) that are affected.

For examples, see compiler/dttable*.c, compiler/dtcompile.c, common/dmtbinfo*.c, and common/dmtbdump*.c

rafaeljw:
@acpibob What if this change is made locally in Linux? I don't think it will become unmanageable going forward.

rafaeljw (approved):
It fixes an existing build issue and I don't see problems with it.

source/include/actbl2.h
```diff
@@ -1550,7 +1550,7 @@ enum AcpiMadtLpcPicVersion {
 
 typedef struct acpi_madt_oem_data
 {
-    UINT8                   OemData[];
+    ACPI_FLEX_ARRAY(UINT8, OemData);
```
jrtc27, line 1553: Structs with flexible array members can't be nested inside other structs, only unions, without extensions

kees: Yes, this works around a deficiency in the C spec using GNU extensions which have been supported by GCC and Clang for a very long time.

jrtc27: I realise that, but you're just swapping one extension for another which doesn't achieve anything as far as I can tell.

If your goal is to avoid extensions then just `typedef UINT8 ACPI_MADT_OEM_DATA[]`, which is an API break but just requires changing `Expr->OemData` to `(*Expr)`.

acpibob:
OK, I'll try this PR as it stands.

PR #815, opened 2022-11-24
specify size for OemData array to avoid compile error

jongwu:
specify size for OemData array to avoid compile error

    The sizeof OemData array is not specified which will lead to compiling
    error of "flexible array member in a struct with no named members".
    From ACPI Spec Table 5.19, the largest data struct for Oem is 8. So,
    specify OemData array size to 8.

    Signed-off-by: Jianyong Wu <jianyong.wu@arm.com>


jongwu:
someone is fixing this.

Issue #816, opened 2022-12-01
Are illegal port ranges still valid?


superm1:
Hi,

I've come across an issue where SBIOS is attempting to write 4 bytes to port I/O 0x80 from AML, and this fails.  Only the first byte is written, and with the Linux kernel debug logs turned up to indicate why the following messages come up:

```
[1058741.307688]   hwvalid-0149 hw_validate_io_request: Denied AML access to port 0x0000000000000080/4 (DMA1 0x0081-0x0083)
[1058741.307694]   hwvalid-0149 hw_validate_io_request: Denied AML access to port 0x0000000000000081/1 (DMA1 0x0081-0x0083)
[1058741.307695]   hwvalid-0149 hw_validate_io_request: Denied AML access to port 0x0000000000000082/1 (DMA1 0x0081-0x0083)
[1058741.307696]   hwvalid-0149 hw_validate_io_request: Denied AML access to port 0x0000000000000083/1 (DMA1 0x0081-0x0083)
```

So the kernel tries to do a 4 byte write, then that's rejected so they're done 1 byte at a time.  Only the first byte succeeds.

I was looking more into this, and it seems that it was introduced because Windows had protected port I/O ranges introduced during the Windows XP timeframe.  If the BIOS tried to write to those ranges messages like this would show up in the Windows event viewer (found from a random search):

```
Source: ACPI
Category: None
Type: Error
Event ID: 5
User: N/A
Computer: Machine_Name
Description: AMLI: ACPI BIOS is attempting to write to an illegal IO port address (0xcf8),
             which lies in the 0xcf8 - 0xcff protected address range. This could lead
             to system instability. Please contact your system vendor for technical assistance. 
             For more information, see Help and Support Center at: 
             http://go.microsoft.com/fwlink/events.asp
```

However taking this exact same system over to Windows 11 and trying to execute the same AML no such messages are emitted and checking the other end of the port I/O all 4 bytes were written.

So I wanted to check; is it really still valid to be guarding the port I/O writes for the DMA1 range?  I'm not sure where the original ranges from https://github.com/acpica/acpica/blob/44ab01315dc9b431df6ede6504f9261f706c62ed/source/components/hardware/hwvalid.c#L205 came from, but at least I couldn't find anything similar searching the public Microsoft website myself.

1. Can these ranges checks be loosened on the more modern OSI values?
2. Or perhaps drop the DMA1 range to help this issue I'm reporting?

superm1:
I think these are the original URLs:
[...]

PR #817, opened 2022-12-02
ACPICA: hwvalid: Drop port I/O validation

superm1:
Microsoft introduced support in Windows XP for blocking port I/O to various regions.  For Windows compatibility ACPICA has adopted the same protections and will disallow writes to those (presumably) the same regions.

On some systems the AML included with the firmware will issue 4 byte long writes to 0x80.  These writes aren't making it over because of this blockage. The first 4 byte write attempt is rejected, and then subsequently 1 byte at a time each offset is tried. The first at 0x80 works, but then the next 3 bytes are rejected.

This manifests in bizarre failures for devices that expected the AML to write all 4 bytes.  Trying the same AML on Windows 10 or 11 doesn't hit this failure and all 4 bytes are written.

Either some of these regions were wrong or some point after Windows XP some of these regions blocks have been lifted.

In the last 15 years there doesn't seem to be any reports popping up of this error in the Windows event viewer anymore.  There is no documentation at Microsoft's developer site indicating that Windows ACPI interpreter blocks these regions. Between the lack of documentation and the fact that the writes actually do work in Windows 10 and 11, it's quite likely Windows doesn't actually enforce this anymore.

So to help the issue, drop enforcing protected ranges.

Fixes https://github.com/acpica/acpica/issues/816
Signed-off-by: Mario Limonciello <mario.limonciello@amd.com>

### DIFF
--- a/source/components/hardware/hwvalid.c
+++ b/source/components/hardware/hwvalid.c
@@ -169,8 +169,8 @@ AcpiHwValidateIoRequest (
  *
  * The table is used to implement the Microsoft port access rules that
  * first appeared in Windows XP. Some ports are always illegal, and some
- * ports are only illegal if the BIOS calls _OSI with a WinXP string or
- * later (meaning that the BIOS itelf is post-XP.)
+ * ports are only illegal if the BIOS calls _OSI with nothing newer than
+ * the specific _OSI strings.
  *
  * This provides ACPICA with the desired port protections and
  * Microsoft compatibility.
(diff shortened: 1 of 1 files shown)


rafaeljw (changes requested):
I'm not sure it is safe to drop the validation altogether, although quite likely it is overkill on post-XP systems.

I would rather rework it so that the OSI_WIN_XP entries take effect only if nothing newer than XP has been acknowledged via _OSI, but the ALWAYS_ILLEGAL entries should be retained.

superm1:
> I'm not sure it is safe to drop the validation altogether, although quite likely it is overkill on post-XP systems.
> 
> I would rather rework it so that the OSI_WIN_XP entries take effect only if nothing newer than XP has been acknowledged via _OSI, but the ALWAYS_ILLEGAL entries should be retained.

OK thanks, I've adjusted accordingly.

rafaeljw (approved):
This looks good to me now, thank you!

rafaeljw:
> Thanks! In terms of getting it to Linux, should I port this PR and send something to linux-acpi referencing this PR?

Yes, please.

Issue #820, opened 2022-12-20
Acpisrc creates broken linux versions


heatd:
Hi,

I'm trying to generate a Linux-like formatted ACPICA version.

Using the latest release acpica-unix-20221020, the command:
`acpica-unix-20221020/generate/unix/bin/acpisrc -l acpica-unix-20221020 acpica-unix-20221020-linux`

generates broken headers and source files, that don't compile.

See [this broken actbl2](https://gist.github.com/heatd/49c95c9b53204892b4e503b24234ccbd) for a quick example.

heatd:
I can confirm the problem still reproduces on acpica HEAD (non-UNIX version)

PR #822, opened 2022-12-21
acpisrc Linux generation fix

heatd:
As reported in issue #820, acpisrc -l was broken and generating bad sources. This PR fixes it.

Note that acpisrc -l without -d (to keep debug prints) is still broken and generates a bad utdebug.c.
I may look into it for this PR, depending on how quickly you get back to me.


acpibob:
>I may look into it for this PR, depending on how quickly you get back to me.

Yes, please take a look. I am merging this pull request.

PR #834, opened 2023-02-07
Remove extern prototypes in achaiku.h file

ElyesH:
Signed-off-by: Elyes Haouas <ehaouas@noos.fr>


rafaeljw:
What's wrong with using 'extern' here?  Is it just inconsistent or is there more to it?

acpibob:
I'd be a bit cautious about this, since the achaiku.h file comes directly from the haiku project.

ElyesH:
"extern" keyword is used implicitly for functions.
it is always implied with function declarations.

acpibob:
Is this true for all C compilers?

ElyesH:
it is for gcc, but for the others I'm not sure.

PR #837, opened 2023-02-07
Revert "Replace fake flexible arrays with actual flexible array members"

acpibob:
Reverts acpica/acpica#813

Causes many build errors on MS Visual C:

Severity	Code	Description	Project	File	Line	Suppression State
Error	C2016	C requires that a struct or union has at least one member (compiling source file ..\..\source\COMPONENTS\Disassembler\dmwalk.c)	AcpiExec	c:\acpica\source\include\acrestyp.h	932

### DIFF
--- a/source/include/acrestyp.h
+++ b/source/include/acrestyp.h
@@ -538,7 +538,7 @@ typedef struct acpi_resource_extended_irq
     UINT8                           WakeCapable;
     UINT8                           InterruptCount;
     ACPI_RESOURCE_SOURCE            ResourceSource;
-    UINT32                          Interrupts[];
+    UINT32                          Interrupts[1];
 
 } ACPI_RESOURCE_EXTENDED_IRQ;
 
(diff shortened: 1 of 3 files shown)


andy-shev:
Moreover, I belive that it's not fake VLA as commit states. The ACPI specification mentioned that at least one is required. 
@kees FYI. So, if you want to fix `Interrupt()` data structure, then it's probably should go as `u32 interrupt; u32 interrupts[];`.

acpibob:
OK, please make a new pull request (and verify that it works correctly).

acpibob:
u32 interrupt; u32 interrupts[];

Why not just: u32 interrupts[1];

kees:
If there will ever be more than 1, it needs to be a flexible array, so, roughly:

```c
...
union {
    u32 interrupt; /* minimum number of interrupts is 1 here. */
    __DECLARE_FLEX_ARRAY(u32, interrupts);
};
```

acpibob:
I don’t think that DECLARE_FLEX_ARRAY worked with MSVC

kees:
Just to close the loop on this PR, I've just opened a PR for a new more comprehensive series of flex-array transformations:
https://github.com/acpica/acpica/pull/856

PR #838, opened 2023-02-08
add support for ClockInput resources introduced in ACPI v6.5

nsait-linaro:
This PR adds support for ClockInput resources introduced in ACPI v6.5 specification.

Please check out the following sections from ACPI v6.5 spec for further details

- 6.4.3.14 ( Clock Input Resource Descriptor)
- 19.6.154 (Clock Input Resource Descriptor Macro)

The changes are tested manually with `iasl`, `acpidump`, `acpiexec` and Linux acpica modifications.


source/compiler/aslcompiler.l
```diff
@@ -461,6 +461,7 @@ NamePathTail                [.]{NameSeg}
 "PinGroup"                  { count (1); return (PARSEOP_PINGROUP); }
 "PinGroupConfig"            { count (1); return (PARSEOP_PINGROUPCONFIG); }
 "PinGroupFunction"          { count (1); return (PARSEOP_PINGROUPFUNCTION); }
+"ClockInput"                { count (1); return (PARSEOP_CLOCKINPUT); }
```
andy-shev, line 464: I'm wondering if the (mis)location is deliberate (to split Pin*() resources instead of putting alphabetically).

source/compiler/aslresources.y
```diff
@@ -214,6 +214,7 @@ ResourceMacroTerm
     | Memory32Term                  {}
     | PinConfigTerm                 {}
     | PinFunctionTerm               {}
+    | ClockInputTerm                {}
```
andy-shev, line 217: I'm wondering if the (mis)location is deliberate (to split Pin*() resources instead of putting alphabetically).

source/compiler/aslresources.y
```diff
@@ -666,6 +667,21 @@ PinFunctionTerm
         error PARSEOP_CLOSE_PAREN   {$$ = AslDoError(); yyclearin;}
     ;
 
+ClockInputTerm
```
andy-shev, line 670: I'm wondering if the (mis)location is deliberate (to split Pin*() resources instead of putting as in other places).

source/components/resources/rsdumpinfo.c
```diff
@@ -374,6 +374,17 @@ ACPI_RSDUMP_INFO        AcpiRsDumpPinFunction[10] =
     {ACPI_RSD_SHORTLISTX,ACPI_RSD_OFFSET (PinFunction.VendorData),          "VendorData",               NULL},
 };
 
+ACPI_RSDUMP_INFO        AcpiRsDumpClockInput[7] =
```
andy-shev, line 377: So, it seems in some cases it goes in between of Pin*() in some at the end of the list. Wondering why so...

source/compiler/aslresources.y
nsait-linaro: No, it isn't. I don't think ClockInput has to break Pin* resources. I can tidy it up

sacdintel: Please submit a new PR if and when you tidy it up! Also please do a fresh git pull/fetch to ensure you have the latest ACPICA since there might have been some new changes in some of those files since the last stable release dated 03312023! Thanks a lot for the contributions!

Issue #841, opened 2023-02-18
iasl: Missing `AcpiGbl_ParseOpRoot = NULL` in adisasm.c leads to segfault when many files are provided with -e


gemarcano:
While disassembling the ACPI tables of my laptop (to fix a bug in one of them), I was running into iasl segfaulting. Specifically, this is what my ACPI dump directory looks like:
```
$ ls *.dat
 apic.dat    boot.dat   dsdt.dat   fpdt.dat   mcfg.dat   ssdt10.dat   ssdt13.dat   ssdt16.dat   ssdt19.dat   ssdt21.dat   ssdt4.dat   ssdt7.dat   tpm2.dat   wsmt.dat
'asf!.dat'   cdit.dat   facp.dat   hpet.dat   msdm.dat   ssdt11.dat   ssdt14.dat   ssdt17.dat   ssdt1.dat    ssdt2.dat    ssdt5.dat   ssdt8.dat   uefi.dat
 bgrt.dat    crat.dat   facs.dat   ivrs.dat   slic.dat   ssdt12.dat   ssdt15.dat   ssdt18.dat   ssdt20.dat   ssdt3.dat    ssdt6.dat   ssdt9.dat   vfct.dat
 ```
 
If I executed `iasl -bf -e *.dat -d apic.dat`, the program segfaulted. After building iasl myself, I threw it through Valgrind, and slowly identified that the problem was that `AcpiGbl_ParseOpRoot` in `adisasm.c` was being freed by `AcpiPsDeleteParseTree (AcpiGbl_ParseOpRoot);`, but unlike other places in the code, the global variables was not being set to NULL, leading future code to try to use it (trying to free it again, I think), even though it had already been freed. Eventually this caused the program to segfault.

Adding the NULL assignment fixes this issue for me. I'll post a PR with the one line fix.

PR #843, opened 2023-02-18
iasl: check Offset before Subtable dereference

gemarcano:
In AcpiDmDumpMadt(), compute and check the Offset before attempting to dereference Subtable fields to prevent a read overflow.

I found this issue while debugging #841. Effectively, some debug statements were trying to dereference data from the updated Subtable pointer before checking if the Offset made it past the end of the table. This PR just moves some lines around to compute and check Offset before trying to access data for debug printing.


sacdintel:
Thanks for the PR. Does this get rid of the seg fault as per issue #841 or is it already taken care of by the already merged PR #842 ?

Issue #849, opened 2023-02-27
[bug] error: variable 'Column' set but not used [-Werror,-Wunused-but-set-variable]


agrajag9:
https://github.com/acpica/acpica/blob/28fc163aa29e208678d901d98bb9030b775521b3/source/tools/acpisrc/asconvrt.c#L1233

See build log here: https://bz-attachments.freebsd.org/attachment.cgi?id=240459

agrajag9:
```
obj/acpisrc ../../../source/tools/acpisrc/asconvrt.c
../../../source/tools/acpisrc/asconvrt.c:1233:29: error: variable 'Column' set but not used [-Werror,-Wunused-but-set-variable]
    UINT32                  Column = 0;
                            ^
1 error generated.
gmake[2]: *** [../Makefile.rules:20: obj/asconvrt.o] Error 1
```

agrajag9:
Looks like some others as well:
```
../../../source/compiler/aslrestype2s.c:1553:29: error: variable 'PinCount' set but not used [-Werror,-Wunused-but-set-variable]
    UINT32                  PinCount = 0;
                            ^
../../../source/compiler/aslrestype2s.c:1739:29: error: variable 'PinCount' set but not used [-Werror,-Wunused-but-set-variable]
    UINT32                  PinCount = 0;
                            ^
../../../source/compiler/aslrestype2s.c:1939:29: error: variable 'PinCount' set but not used [-Werror,-Wunused-but-set-variable]
    UINT32                  PinCount = 0;
                            ^
3 errors generated.
```

sacdintel:
I believe this issue has been fixed after merging PR #851 so will go ahead and close it. Thanks!

PR #851, opened 2023-03-01
Remove write-only variables

juikim:
This fixes build with Clang 15 and later.  Please see Issue #849.


acpibob:
Why did you remove these comments:

             * NOTE: there is no "length" field for this, so from ACPI spec:
             *  The number of pins in the table can be calculated from:
             *  PinCount = (Resource Source Name Offset - Pin Table Offset) / 2
             *  (implies resource source must immediately follow the pin list.)

PR #856, opened 2023-03-02
Flex array transformations, take 2

kees:
Hello!

This is a more comprehensive series to address all the technical debt related to 1-element and 0-length arrays in ACPICA code, with the goal of making this behave correctly in the Linux kernel when built with GCC and Clang's new `-fstrict-flex-arrays=3` option.

I did per-patch A/B build output analysis checking for `.text` and `.data` section changes, solving or noting any differences in each patch. Specifically, this series seems to have found possibly 2 bugs in existing size calculations, noted in their individual patches:
- `AcpiDmDumpNfit()`'s calculation of `ACPI_NFIT_INTERLEAVE` appears to have been originally wrong, but is now fixed.
- `AcpiDmDumpDmar()`'s calculation of `ACPI_DMAR_ANDD` seems suspicious, but the existing result appears to be used as a table offset. It's unclear if this is a bug or intentional, so I left padding in the struct to keep the binary output unchanged.

The helper to work around the C spec's limitation for flexible array members has been limited to only GCC (and Clang), until MSVC can fix their handling.

### DIFF
--- a/source/common/dmtbdump2.c
+++ b/source/common/dmtbdump2.c
@@ -1538,7 +1538,7 @@ AcpiDmDumpNfit (
             /* Has a variable number of 64-bit addresses at the end */
 
             InfoTable = AcpiDmTableInfoNfit6;
-            FieldOffset = sizeof (ACPI_NFIT_FLUSH_ADDRESS) - sizeof (UINT64);
+            FieldOffset = sizeof (ACPI_NFIT_FLUSH_ADDRESS);
             break;
 
         case ACPI_NFIT_TYPE_CAPABILITIES:    /* ACPI 6.0A */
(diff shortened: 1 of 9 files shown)


kees:
Hi!

Yes, that's what I did. See https://github.com/acpica/acpica/pull/856/commits/e73b227e8e475c20cc394f237ea35d592fdf9ec3

This series handles the refactoring in 3 ways:
- simple cases of [0] -> [] which are valid for both GCC/Clang and MSVC without a macro
- [1] -> [] where there are no binary differences and no macro needed
- all remaining conversions using the new macro which wraps [] for GCC/Clang or leaves it as [0] for MSVC.

sacdintel:
Alright, sounds good. Bob and I need to run a few more tests and then we will merge it soon. Also, I realized today that the MSVC build failure was a different issue and not related to the flexible array issue, although it gets confusing sometimes since the C/C++ compilers are slightly different between MSVC 2017 (Bob's machine) and MSVC 2019 (my machine) and also before GCC 12 (Bob's Linux/Cygwin) this issue probably doesn't happen as opposed to GCC 12 or 12+ (my Ubuntu/Cygwin).

kees:
Awesome; thanks! Let me know if I need to make any adjustments. I don't have a way to test MSVC, so that part I had to do based on prior MSVC-only changes. :P

sacdintel:
This looks good to me for now, so I will go ahead and merge this PR. Thanks!

PR #862, opened 2023-04-04
Move ACPI_FUNCTION_ENTRY to function entry

tamird:
This changed in f98909b7928cac8c5328e4852ad5a7ea671204fb, seemingly by accident.

### DIFF
--- a/source/components/resources/rsaddr.c
+++ b/source/components/resources/rsaddr.c
@@ -431,12 +431,12 @@ AcpiRsGetAddressCommon (
     ACPI_RESOURCE           *Resource,
     AML_RESOURCE            *Aml)
 {
+    ACPI_FUNCTION_ENTRY();
+
     /* Avoid undefined behavior: member access within misaligned address */
 
     AML_RESOURCE_ADDRESS Address;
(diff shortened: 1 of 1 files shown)


sacdintel:
Hi Tamir,
We have a minor problem with this PR and hopefully you can help test a couple of things before we merge this PR again with a new suggestion. So our GCC compiler also checks the code compatibility with older C language standards (like C89/C90) to maintain portability across all platforms. It breaks the build due to the warning "ISO C89/C90 forbids mixed declaration and code in C" which is caused by declaration of AML_RESOURCE_ADDRESS Address; after the ACPI_FUNCTION_ENTRY(); code invocation. Can you please test and let me know at your convenience if the desired functionality is still maintained if you moved the declaration line right before the function code invocation line and then it is all followed by the memcpy line? If that doesn't work then can the Address instance be made to have more scope (or be changed to an array of structs?) like outside the bigger function in rsaddr.c file? Please let me know your thoughts on this and we want to thank you once again for all the contributions you have made! You can change the GCC C standard option flag to use and test it on C89 or C90. Thanks!

tamird:
No, this is fine. I didn't know you target C89.

PR #864, opened 2023-04-05
Add support for Interrupt in acpiexec

jmarinho:
Acpiexec lacked support for GED devices (relevant for HW-reduced platforms), where an interrupt (of a given GSIV) triggers the execution of the _EVT method of that GED.

This patch set introduces support for interrupt simulation in acpiexec.

### DIFF
--- a/source/tools/acpiexec/aeinstall.c
+++ b/source/tools/acpiexec/aeinstall.c
@@ -176,6 +176,13 @@ AeInstallPciHandler (
     void                    *Context,
     void                    **ReturnValue);
 
+static ACPI_STATUS
+AeInstallGedHandler (
+    ACPI_HANDLE             ObjHandle,
+    UINT32                  Level,
+    void                    *Context,
+    void                    **ReturnValue);
+
 
 BOOLEAN                     AcpiGbl_DisplayRegionAccess = FALSE;
 ACPI_CONNECTION_INFO        AeMyContext;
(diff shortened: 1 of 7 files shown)


sacdintel (changes requested):
Hi Jose,
Thanks for the PR which can add a major feature. I do get a few errors under MSVC that need to be fixed though before I can merge it. 
- Saket Dumbre

source/tools/acpiexec/aeinstall.c
```diff
@@ -343,6 +350,108 @@ AeInstallRegionHandlers (
...
+
+			/*
+			 * Found an Interrupt resource. Link the interrupt resource
+			 * and the _EVT method of this GED device in the GED event handler list.
+			 */
+				ACPI_GED_HANDLER_INFO *GedHandler =
+				ACPI_ALLOCATE (sizeof (ACPI_SCI_HANDLER_INFO));
+			if (!GedHandler)
+			{
+				return_ACPI_STATUS (AE_NO_MEMORY);
```
sacdintel, line 430: 3 errors total
430:'_AcpiFunctionName': undeclared identifier - I think you first need to use ACPI_FUNCTION_ENTRY() at the beginning of the function 
    ACPI_FUNCTION_ENTRY(AeInstallGedHandler); (ln 380)
this error is part of although I am not 100% sure right now and just this might not fix this error here

438 and 441:'Interrupts':is not a member of 'acpi_resource_extended irq' - you need to use u.Interrupts[0] both places to get rid of this error

You also might want to do an include for rsmisc.c file

Also 2 warnings on line 430:
'function': 'const char*' differs in levels of indirection from 'int' - you can try type casting the return value
'AcpiUtStatusExit': different types for formal and actual parameter 2 - which is the function name that needs to be fixed too

jmarinho: Thank you for the review,

- I've added the ACPI_FUNCTION_ENTRY() to the top of the AeInstallGedHandler function.
- I'm returning AE_NO_MEMORY; if ACPI_ALLOCATE fails (I believe this solves the 2 additional warnings).

So far I have tested the build with GCC.
I will build this branch with MSVC too and confirm that all errors and warnings are removed -- I'll comment once I've performed the verification with MSVC.

sacdintel:
Hi Jose,
I apologize since I have reverted a commit to match the Linux kernel ACPICA, so now you don't need to access the union members with a u.variable (like I had suggested). Also, there is a conflict in rsaddr.c but I will take your patch and merge it. Also if the builds are successful, I will go ahead and merge your PR. Thanks!

sacdintel:
Jose,
Can you please do me a favor by doing a fresh git pull and then check if your PR build works successfully with both GCC and MSVC and update your PR accordingly and make changes here as well, if needed? I still see the same build errors in MSVC for some reason, but also there have been some changes in some files which your commits touch in ACPICA (eg. rsaddr.c and not to mention some conflicts too) from previous PRs or changes I (or Bob) must have made. I am sorry, your PR might have been working correctly when you did the development and testing, but our ACPICA code keeps on changing every few days due to new features (by us & contributors), bug fixes, merged PRs etc. and it doesn't take much to lose compatibility. I apologize once again for the inconvenience, hope you understand! Thanks! As soon as this is resolved, I will merge your PR as soon as I can and then for your further development, a fresh git pull/fetch again will keep your local branch/master copy up to date as well. 
- Saket Dumbre

jmarinho:
We've rebased the changes to the latest master baranch.
Building with MSVC 2017 yields warning C4201, but that does not seem related to the changes in this PR.
No other errors or warnings are being identified by MSVC 2017 (msbuild generate\msvc2017\AcpiComponents.sln -t:Rebuild -p:Configuration=Release).

PR #868, opened 2023-04-12
build: add module name for build system

najumon1980:
add module name for Zephyr build system

### DIFF
--- a/zephyr/module.yml
+++ b/zephyr/module.yml
@@ -1,0 +1,4 @@
+name: acpica
+build:
+  cmake-ext: True
+  kconfig-ext: True


jhedberg:
Some background:

The `zephyr/module.yml` file is needed to integrate the ACPICA tree as a submodule into the Zephyr build system. Zephyr has lots of submodules (you can see them listed here: https://github.com/zephyrproject-rtos/zephyr/blob/afaf4cddd24240307dd823cfcc1848e6d9440451/west.yml#L27 ) and they take varying approaches to this. Some upstreams of Zephyr submodules have accepted the zephyr-specific files (e.g. OpenThread: https://github.com/openthread/openthread ), whereas others have forked the upstream tree and added the Zephyr integration there.

We understand if ACPICA upstream doesn't want to pollute its top-level directory with Zephyr-specific integration files, so if this PR isn't acceptable, we'll simply add the files to the ACPICA fork that we've created within the Zephyr GitHub project: https://github.com/zephyrproject-rtos/acpica ). However, from a logistics perspective (in particular when it comes to keeping up with ACPICA upstream changes), it'd be simpler for us if these were part of the upstream tree rather than the Zephyr fork of the tree.

zephyr/module.yml
```diff
@@ -0,0 +1,4 @@
+name: acpi
```
jhedberg, line 1: @najumon1980 shouldn't this be `acpica`? I thought we agreed that the ACPI APIs in Zephyr will use the "acpi" namespace, but any references to ACPICA should use "acpica".

najumon1980: sure I will update

sacdintel:
Hi Zephyr folks,
We would like to know how many files and the total size it would add to the ACPICA project. We all would like to work with you all on this, but a lot of our customers work on embedded systems with memory/resource constraints, so it might not be fair to them to make a new release of ACPICA "very bulky" but we would be OK with adding a few files in the upstream/master to tackle any arising logistical challenges. We might need to discuss it with a few important stakeholders as well and then come to a conclusion. Please keep us updated. I only see 1 file changed/added so far with only 4-5 lines of code in this PR. Thanks!
- Saket Dumbre

najumon1980:
hi Saket,
for Zephyr RTOS we need only this PR and the one we early given for OS service layer (https://github.com/acpica/acpica/pull/861) which already got merged. So there are total 1342  line of code ( This PR have 4 line & 1338 for previous PR861).

sacdintel:
I don't see a problem why we can't add this .yml file in that case. I just hope it is only included in the build when the OS is detected as Zephyr RTOS (in the aczephyr.h file) and not for any other OS, else I am afraid it might break the build since other C compilers that run ACPICA might not be able to execute a .yml file without any extensions added (eg. for GCC/Linux).

acpibob:
Actually, the OS should be detected in acenv.h, something similar to this;

#elif defined(__ZEPHYR__)
#include "aczephyr.h"

This way, aczephyr.h is only pulled in for ZEPHYR. (which the current patch seems to do).

jhedberg:
The `zephyr/module.yml` should end up getting completely ignored by other OSes. It's not actually used by any GCC or similar build tools, rather Zephyr has a higher level "meta tool" called [west](https://github.com/zephyrproject-rtos/west) that requires this yaml file in each submodule.

jhedberg:
@acpibob @sacdintel could we try to reach some conclusion here? As I described, the `zephyr/module.yml` file is not something that gets executed or interpreted by gcc or similar tools. It's simply describes for the Zephyr build system how this submodule should get integrated as part of the larger Zephyr build process. I don't think there's any risk of other builds or OSes getting confused by this, since it's purely a Zephyr-specific file.

sacdintel:
Hi Johan,
I will test and build ACPICA under different environments with this PR this week and if the builds are good then I will approve and merge this PR, but in case it breaks something, I will let you know here so that you can fix it! Regarding this PR, it seems alright and we won't have any trouble including an extra file for Zephyr. Thanks for your understanding and patience!

jhedberg:
@sacdintel thanks a lot!

sacdintel:
You are welcome!

Issue #867, opened 2023-04-07
Various build issues w/ GCC 12


pprindeville:
The file `source/include/platform/aclinux.h` defines `ACPI_UINTPTR_T` as `uintptr_t` even though it doesn't include `<stdint.h>` from whence that definition comes.

The address of `CurrentSp` (a local variable) is being stored into `AcpiGbl_EntryStackPointer` which is, as the name implies, a global and the compiler considers a persistent reference to an ephemeral variable a "dangling pointer":

```
obj/acpiexec ../../../source/components/utilities/utdebug.c
../../../source/components/utilities/utdebug.c: In function 'AcpiUtInitStackPtrTrace':
../../../source/components/utilities/utdebug.c:188:31: error: storing the address of local variable 'CurrentSp' in 'AcpiGbl_EntryStackPointer' [-Werror=dangling-pointer=]
  188 |     AcpiGbl_EntryStackPointer = &CurrentSp;
      |     ~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~
../../../source/components/utilities/utdebug.c:185:29: note: 'CurrentSp' declared here
  185 |     ACPI_SIZE               CurrentSp;
      |                             ^~~~~~~~~
In file included from ../../../source/include/acpi.h:173,
                 from ../../../source/components/utilities/utdebug.c:154:
../../../source/include/acglobal.h:335:41: note: 'AcpiGbl_EntryStackPointer' declared here
  335 | ACPI_GLOBAL (ACPI_SIZE *,               AcpiGbl_EntryStackPointer);
      |                                         ^~~~~~~~~~~~~~~~~~~~~~~~~
../../../source/include/acpixf.h:188:17: note: in definition of macro 'ACPI_GLOBAL'
  188 |     extern type name
      |                 ^~~~
cc1: all warnings being treated as errors
make[4]: *** [../Makefile.rules:20: obj/utdebug.o] Error 1
```

Either a pragma to suppress the warning, or else casting the pointer to a `uintptr_t` [laundering] so it loses its pointer-ness, is needed.

acpibob:
What version of ACPICA are you running. I ask because I thought the dangling pointer issue was solved.
Bob


From: Philip Prindeville ***@***.***>
Sent: Friday, April 07, 2023 12:16 PMWhat
To: acpica/acpica ***@***.***>
Cc: Subscribed ***@***.***>
[...]

PR #869, opened 2023-04-12
Fix GCC 12 dangling-pointer warning

pprindeville:
We're storing a persistent pointer to an ephemeral local variable which technically is a dangling pointer and the compiler is correct. However, since we never indirect the pointer, this is a safe operation and we can suppress the warning.

Also, some C run-times (like MUSL) aren't including <stdint.h> indirectly so we must include it explicitly or we won't have the type definition for uintptr_t.

Fixes issue #867.

### DIFF
--- a/source/components/utilities/utdebug.c
+++ b/source/components/utilities/utdebug.c
@@ -185,7 +185,12 @@ AcpiUtInitStackPtrTrace (
     ACPI_SIZE               CurrentSp;
 
 
+#pragma GCC diagnostic push
+#if defined(__GNUC__) && __GNUC__ >= 12
+#pragma GCC diagnostic ignored "-Wdangling-pointer="
+#endif
     AcpiGbl_EntryStackPointer = &CurrentSp;
+#pragma GCC diagnostic pop
 }
 
 
--- a/source/include/platform/aclinux.h
+++ b/source/include/platform/aclinux.h
@@ -312,6 +312,7 @@
 #ifdef ACPI_USE_STANDARD_HEADERS
 #include <stddef.h>
 #include <unistd.h>
+#include <stdint.h>
 
 #define ACPI_OFFSET(d, f)   offsetof(d, f)
 #endif


sacdintel:
Hi Philip,
Thanks for your bug report and this PR which hopefully fixes it! Let me review and test it and get back to you!
- Saket Dumbre

sacdintel:
Philip,
I also had a question/suggestion which you could help us test on GCC 12 and that way we might not need to suppress any warnings using #pragma. 
What if in dbstats.c file towards the end around line number 654 and 655 (new lines) we simply reassign the two pointers to NULL  (AcpiGbl_EntryStackPointer = NULL; and also AcpiGbl_LowestStackPointer = NULL;)? I am not 100% sure if it will resolve the issue but we could give it a fair try I think and also check for functionality/operation retention. Some other pointers which could possibly be dangling can also be tackled in a similar manner. What do you think about this possible solution? I might not be able to test with GCC 12 since my older Ubuntu version might not be able to upgrade beyond GCC 9 or 10 I think. Thanks a lot!

pprindeville:
I tried your suggestion and it makes no difference, as I anticipated.  The pointer technically begins dangling as soon as `AcpiUtInitStackPtrTrace()` returns and pops its stack frame.

sacdintel:
OK got it, this is what I had feared would happen or how things worked once we return from the function. So I will go ahead and merge your patch this/next week after confirming things don't break elsewhere in Windows/MSVC env. Thanks for the contribution and please keep us in the loop about any (if) further problems with GCC 12+ versions.

pprindeville:
> Thanks for the contribution and please keep us in the loop about any (if) further problems with GCC 12+ versions.

You got it!

sacdintel (changes requested):
Hi Philip,
Unfortunately, I get an error while "making" acpiexec on Cygwin with GCC version 11.3.0

../../../source/components/utilities/utdebug.c:189:32: error: unknown option after ‘#pragma GCC diagnostic’ kind [-Werror=pragmas]
  189 | #pragma GCC diagnostic ignored "-Wdangling-pointer="
      |                                ^~~~~~~~~~~~~~~~~~~~~
cc1: all warnings being treated as errors
make[1]: *** [../Makefile.rules:20: obj/utdebug.o] Error 1

Any idea how to resolve this? Is it a good idea to use an #ifdef for GCC versions 12+ since it looks like the lower versions don't recognize the "-Wdangling-pointer=" warning turned error message?

Update: I managed to seemingly fix it and I am gonna go ahead and add some suggestions to this PR to be included, so that we maintain compatibility with non-GCC systems.   

#if __GNUC__ >= 12
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Wdangling-pointer="
    AcpiGbl_EntryStackPointer = &CurrentSp;
#pragma GCC diagnostic pop
#else
    AcpiGbl_EntryStackPointer = &CurrentSp;
#endif

pprindeville:
> Hi Philip, Unfortunately, I get an error while "making" acpiexec on Cygwin with GCC version 11.3.0

Fixed the issue.  Built with 11.3.0 with no warnings.

pprindeville:
Anything needed from me at this time?

sacdintel:
Not really, but I will keep you updated if anything is needed. I need to test the updated PR some time this week under various environments and then I can approve and merge this PR. Thanks a lot!

sacdintel (approved):

PR #870, opened 2023-04-20
ACPICA: Modify ACPI_STATE_COMMON

georgejguo:
Avoid trailing semicolons in macro, and it's not readable to put macro ACPI_STATE_COMMON and other variables in the same line.
So modify the macro and just put it in a single line.


georgejguo:
hi，I wonder why here's code is different with kernel's.
And this results  gives 1 pr here and 1 patch to kernel.

PR #875, opened 2023-05-29
Slightly simplify an error message in AcpiDsResultPush()

tititiou36:
'object' is known to be NULL at this point. There is little value to log it twice in the error message.


sacdintel:
Looks good. Thanks for the contribution!

PR #879, opened 2023-06-29
fix coverity issue

lieric15:
dswstate.c
'State' is deleted before DEBUG_PRINT()

nsalloc.c
'Node' is delete before DEBUG_PRINT()

nsdump.c
the default block will not be excuted.

psargs.c
when hit the 'Field' is null, the 'Arg' will not be free which will lead memory leak.

### DIFF
--- a/source/components/namespace/nsdump.c
+++ b/source/components/namespace/nsdump.c
@@ -635,10 +635,6 @@ AcpiNsDumpOneObject (
             break;
         }
         break;
-
-    default:
-        AcpiOsPrintf ("\n");
-        break;
     }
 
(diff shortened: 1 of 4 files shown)


source/components/namespace/nsdump.c
```diff
@@ -635,10 +635,6 @@ AcpiNsDumpOneObject (
             break;
         }
         break;
-
```
sacdintel, line 638: Getting rid off of this default case under display type in nsdump.c is not necessary and it also throws a warning which we treat as error in Unix environment.

PR #882, opened 2023-07-06
os_specific: zephyr: add generic platform support

najumon1980:
add support for uefi as well as non uefi platform for zephyr rtos.
Also add support for 32bit platforms.

### DIFF
--- a/source/os_specific/service_layers/oszephyr.c
+++ b/source/os_specific/service_layers/oszephyr.c
@@ -153,7 +153,7 @@
 #include "accommon.h"
 #include "acapps.h"
 #include "aslcompiler.h"
-#include <zephyr/arch/x86/efi.h>
+#include <zephyr/arch/x86/x86_acpi_osal.h>
 #include <zephyr/drivers/pcie/pcie.h>
 #include <zephyr/dt-bindings/interrupt-controller/intel-ioapic.h>
 #include <zephyr/sys/__assert.h>
@@ -163,15 +163,13 @@ LOG_MODULE_DECLARE(acpica, LOG_LEVEL_ERR);
 
 typedef void (*zephyr_irq_t)(const void *);
 
-#define ASL_MSG_BUFFER_SIZE (1024 * 128)
-
 /* Global varibles use from acpica lib. */
 BOOLEAN                     AslGbl_DoTemplates = FALSE;
 BOOLEAN                     AslGbl_VerboseTemplates = FALSE;
 
 char                        AslGbl_MsgBuffer[ASL_MSG_BUFFER_SIZE];
 static BOOLEAN              EnDbgPrint;
-
+static ACPI_PHYSICAL_ADDRESS RsdpPhyAdd;
 
 /******************************************************************************
  *
@@ -384,12 +382,12 @@ AcpiOsReadMemory (
 
         *((UINT32 *) Value) = sys_read32 (Address);
         break;
-
+#if defined(__x86_64__)
     case 64:
 
         *((UINT64 *) Value) = sys_read64 (Address);
         break;
-
+#endif
     default:
 
         return (AE_BAD_PARAMETER);
@@ -435,12 +433,12 @@ AcpiOsWriteMemory (
 
         sys_write32 ((UINT32) Value, Address);
         break;
-
+#if defined(__x86_64__)
     case 64:
 
         sys_write64 ((UINT64) Value, Address);
         break;
-
+#endif
     default:
 
         return (AE_BAD_PARAMETER);
@@ -773,11 +771,17 @@ ACPI_PHYSICAL_ADDRESS
 AcpiOsGetRootPointer (
     void)
 {
+	 LOG_DBG ("");
 
-    LOG_DBG ("");
-    return ((ACPI_PHYSICAL_ADDRESS) efi_get_acpi_rsdp ());
+	if(RsdpPhyAdd)
+	{
+		return RsdpPhyAdd;
+	}
+
+	RsdpPhyAdd = (ACPI_PHYSICAL_ADDRESS)acpi_rsdp_get();
+
+	return RsdpPhyAdd;
 }
-
 
 #ifndef ACPI_USE_NATIVE_MEMORY_MAPPING
 /******************************************************************************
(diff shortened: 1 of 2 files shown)


najumon1980:
@ acpibob could you please review and merge the PR if ok ?

tbursztyka (changes requested):
I would revisit this entirely.

Have a function in zephyr side (in include/zephyr/arch/x86/acpi.h) called z_acpi_get_rsdp_phys_addr(), which one will internally call the existing code that you are basically duplicating here: less code to maintain in both projects then. (just store the physical address as well instead of the mapped one only, in zephyr side I mean). 
The code there as been working for years, it is less prone to bug and does not require to bloat the adaptation code in acpica part (which should be as minimal as possible).

najumon1980:
> I would revisit this entirely.
> 
> Have a function in zephyr side (in include/zephyr/arch/x86/acpi.h) called z_acpi_get_rsdp_phys_addr(), which one will internally call the existing code that you are basically duplicating here: less code to maintain in both projects then. (just store the physical address as well instead of the mapped one only, in zephyr side I mean). The code there as been working for years, it is less prone to bug and does not require to bloat the adaptation code in acpica part (which should be as minimal as possible).

We are replacing the legacy acpi x86 implementation with acpica (see the draft Zephyr PR: https://github.com/zephyrproject-rtos/zephyr/pull/59917 where the x86/../acpi.c removed). With minimum acpica table initialization (AcpiInitializeTables), we can retrieve most of required tables which is design for pre-kernel initialization. 

ACPICA is already using on multiple OS and hence I believe it is more stable and also need not want worry much about acpi version as platform FW updated to newer versions.

najumon1980:
@tbursztyka would like to know whether you got time to review my response ?

sacdintel:
Is there any other way around to avoid hardcoding the RSDP address and can instead be passed from the Zephyr side? Much of the suggestion from @tbursztyka might help address it but I am not sure of the motive firstly behind moving on from the legacy x86 acpi implementation. Thanks!

sacdintel:
Let me know what you would like to do. From my end, everything looks good and is ready to be merged. Since it is a major implementation change from the Zephyr side, I hope it is approved by the community and other architects also want to move ahead with this! Thanks!

jhedberg:
> Let me know what you would like to do. From my end, everything looks good and is ready to be merged. Since it is a major implementation change from the Zephyr side, I hope it is approved by the community and other architects also want to move ahead with this! Thanks!

Since this is not very time critical, I'd suggest to wait until August, when @tbursztyka is back from vacation (I'll also be off for the next two weeks). Then we can make sure that we have proper consensus among the upstream Zephyr maintainers before proceeding with the changes.

najumon1980:
> Let me know what you would like to do. From my end, everything looks good and is ready to be merged. Since it is a major implementation change from the Zephyr side, I hope it is approved by the community and other architects also want to move ahead with this! Thanks!

@sacdintel I will discuss Zephyr arch change in Zephyr Architecture Working Group and will update you once it got approval (although this PR will not impact/break existing functionalities).

sacdintel:
Sounds good guys! Take your time, no rush! Thanks for the updates!

tbursztyka:
The problem I see with this is that it bring even more zephyr API usage into an external code. So more maintenance hassle. 
Thus why I think it's better to keep it as minimal as possible. Get the rsdp pointer function from zephyr and call it here, instead of coding it here: less exposure of zephyr API, less risk of api changes affecting this code then, etc etc... not to mention the configuration option from zephyr, as well as the x86 bios stuff btw: which makes this code not zephyr arch-agnostic anymore.

najumon1980:
> The problem I see with this is that it bring even more zephyr API usage into an external code. So more maintenance hassle. Thus why I think it's better to keep it as minimal as possible. Get the rsdp pointer function from zephyr and call it here, instead of coding it here: less exposure of zephyr API, less risk of api changes affecting this code then, etc etc... not to mention the configuration option from zephyr, as well as the x86 bios stuff btw: which makes this code not zephyr arch-agnostic anymore.

Then the entire OS abstraction layer need to move to Zephyr repo. I can move  if you think source/os_specific/service_layers/oszephyr.c is better to be part of Zephyr.

jhedberg:
> Then the entire OS abstraction layer need to move to Zephyr repo. I can move if you think source/os_specific/service_layers/oszephyr.c is better to be part of Zephyr.

Could you elaborate a bit on that logic? I think Tomasz is just suggesting to keep the `oszephyr.c` code and set of Zephyr APIs we depend on minimal, but not getting rid of it entirely. I believe ACPICA is designed so that each supported OS has an adaption layer as part of  the ACPICA tree - not outside of it.

tbursztyka:
Indeed, I am not proposing to ditch oszephyr.c at all. What I want is:
- a public z_arch_acpi_rsdp_lookup() function on zephyr side (it's really easy to implement since the code already exists for x86 in arch/x86/core/acpi.c) 
- then in present oszephyr.c, use that function to get rsdp

result: No x86, no EFI, no BIOS, not too much zephyr API etc... exposure into acpica.

source/os_specific/service_layers/oszephyr.c
```diff
@@ -773,11 +772,21 @@ ACPI_PHYSICAL_ADDRESS
...
 
-    LOG_DBG ("");
-    return ((ACPI_PHYSICAL_ADDRESS) efi_get_acpi_rsdp ());
-}
+	if(RsdpPhyAdd)
+	{
+		return RsdpPhyAdd;
+	}
 
+#if defined(CONFIG_X86_EFI)
```
tbursztyka, line 782: No you are still bringing CONFIG_X86_EFI from zephyr. Just get a unique function from zephyr to call, as proposed

tbursztyka (changes requested):

tbursztyka (changes requested):
ok now it's fine.

sacdintel:
Is this PR ready for our review and testing? Please let me know, thanks!

jhedberg (approved):

tbursztyka (approved):

jhedberg:
@sacdintel from our side the PR ready now i.e. feel free to do a final review & merge it if it looks good.

najumon1980:
@sacdintel could you please review and merge if OK

sacdintel:
Looks good, thanks guys!

PR #885, opened 2023-07-21
fix the phat getting wrong string length and offset

Ivanhu5866:
Unlike Windows UTF-16 2-byte wide characters, wchar_t on Linux is 4 bytes UTF-32. So it will get unexpected string length.
  ```  
    /* 2196, Subtable->Type 1 */
    /* 2291, PathLength 20, Offset 50, Table->Length A2 */
    2307, Subtable->Length 2E, VendorLength FFFFFFF2, Offset 70 PathLength: 20
    
    [0002]                      Subtable Type : 0001 [Firmware Health Data]
    [0002]                             Length : 002E
    [0001]                           Revision : 00
    [0002]                           Reserved : 0000
    [0001]                             Health : 00
    [0016]                        Device GUID : 91AF0530-5D86-470E-A6B0-0A2DB9408249
    [0004]             Device-Specific Offset : 0000002A
    [0032]                        Device Path : 41 00 42 00 43 00 44 00 45 00 46 00 00 00 01 02 /* A.B.C.D.E.F..... */\
    /* 010h 0016  16 */                            03 04 00 00 28 00 00 00 00 00 01 00 00 00 30 05 /* ....(.........0. */\
```

And use the correct ACPI_PHAT_HEALTH_DATA struct to calculate offset.

### DIFF
--- a/source/common/dmtbdump2.c
+++ b/source/common/dmtbdump2.c
@@ -149,7 +149,6 @@
  *
  *****************************************************************************/
 
-#include <wchar.h>
 #include "acpi.h"
 #include "accommon.h"
 #include "acdisasm.h"
@@ -2283,7 +2282,8 @@ AcpiDmDumpPhat (
     UINT32                  PathLength;
     UINT32                  VendorLength;
     UINT16                  RecordType;
-    const wchar_t           *WideString;
+    const UINT16            *WideString;
+    UINT32                  i;
 
 
     Subtable = ACPI_ADD_PTR (ACPI_PHAT_HEADER, Table, sizeof (ACPI_TABLE_PHAT));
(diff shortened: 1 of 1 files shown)


sacdintel:
Hi Ivan, 

We will need some more time to review it, test it and verify it because @acpibob has a future agenda which is something very similar in fixing the PHAT table! Thanks!

sacdintel:
@Wer-Wolf  I just realized (after merging your PHAT fixes PR) that this PR exists, although I have concerns regarding the code quality and proper functionality. What would you recommend we do with this? Thanks!

Wer-Wolf:
The second commit is already included in my PR, and the first commit is obsolete since now we use the device-specific data offset to calculate the path length.

So i think this PR can be dropped.

sacdintel:
Thanks for the quick response, that is the same line I was going as well. I concur that this PR has already served its intended purpose and can be closed safely! Thanks @Ivanhu5866 for the PR!

PR #894, opened 2023-09-22
dmtbdump: remove duplicated code for both paths of an if statement

ColinIanKing:
There is an if statement where both paths print the same output, hence the if is redundant and can be replace with just the print.


sacdintel:
Thanks for the PR, looks good and merging it now!

PR #895, opened 2023-09-22
dmtbdump2: remove redundant failure check on Status

ColinIanKing:
The failure check on Status is redundant as it is being checked in the previous if statement and before the if statement too. It is impossible for ACPI_FAILURE(Status) to be true because the previous checks have already returned out of the function at this point. Remove it.


sacdintel:
Looks good again, thanks for the PR contributions and making ACPICA better!

PR #897, opened 2023-09-22
[RFC] nssearch: Improve performance of AcpiNsSearchOneScope

stanspas:
As documented, the linear search performed by AcpiNsSearchOneScope is
good enough for small namespace trees and is generally dominated by
other overheads anyway.

However, there are environments where improved performance during boot
is highly desirable.

This pull request is a proposed path forward towards improving performance by
using a hash table to look up nodes. The hash table is keyed by a
combination of node name and parent node pointer.

Instead of a fixed number of hash table buckets, it might be preferable
to pick the size dynamically: for example, allocation can be shifted
until after the total size of ACPI tables is known, and that can be used
for a heuristic to pick the size and avoid extraneous memory overhead.
The specifics around the best approach here are up for discussion.

Another point for discussion is whether the rather Linux-inspired set of macros
that platforms need to define to make use of the acceleration is appropriate, or
whether more of the core logic to compute hashes and handle simple hashtables
should reside in ACPICA itself.

Signed-off-by: Peter Szuecs <pszucs@amazon.de>
Co-authored-by: Stanislav Spassov <stanspas@amazon.de>

### DIFF
--- a/source/components/namespace/nsalloc.c
+++ b/source/components/namespace/nsalloc.c
@@ -237,6 +237,10 @@ AcpiNsDeleteNode (
         return_VOID;
     }
 
+#ifdef ACPI_USE_ACCELERATED_NS_SEARCH
+    AcpiNsSearchAccelerationDeleteNode(Node);
+#endif
+
     /* Detach an object if there is one */
 
     AcpiNsDetachObject (Node);
@@ -421,6 +425,10 @@ AcpiNsInstallNode (
 
     Node->OwnerId = OwnerId;
     Node->Type = (UINT8) Type;
+
+#ifdef ACPI_USE_NS_SEARCH_ACCELERATION
+    AcpiNsSearchAccelerationAddNode (Node);
+#endif
 
     ACPI_DEBUG_PRINT ((ACPI_DB_NAMES,
         "%4.4s (%s) [Node %p Owner %3.3X] added to %4.4s (%s) [Node %p]\n",
--- a/source/components/namespace/nsaccess.c
+++ b/source/components/namespace/nsaccess.c
@@ -208,6 +208,14 @@ AcpiNsRootInitialize (
         goto UnlockAndExit;
     }
 
+#ifdef ACPI_USE_NS_SEARCH_ACCELERATION
+    Status = AcpiNsSearchAccelerationInit ();
+    if (ACPI_FAILURE (Status))
+    {
+        return_ACPI_STATUS (Status);
+    }
+#endif
+
     /*
      * Tell the rest of the subsystem that the root is initialized
      * (This is OK because the namespace is locked)
@@ -272,6 +280,10 @@ AcpiNsRootInitialize (
 
         NewNode->Parent = &AcpiGbl_RootNodeStruct;
         PrevNode = NewNode;
+
+#ifdef ACPI_USE_NS_SEARCH_ACCELERATION
+        AcpiNsSearchAccelerationAddNode (NewNode);
+#endif
 
         /*
          * Name entered successfully. If entry in PreDefinedNames[] specifies
--- a/source/components/namespace/nssearch.c
+++ b/source/components/namespace/nssearch.c
@@ -172,6 +172,46 @@ AcpiNsSearchParentTree (
 
 /*******************************************************************************
  *
+ * FUNCTION:    AcpiNsSearchOneScopeCore
+ *
+ * PARAMETERS:  TargetName      - Ascii ACPI name to search for
+ *              ParentNode      - Node to search under
+ *
+ * RETURN:      The matching node or NULL, when a node with the provided name,
+ *              parented by the provided node doesn't exist.
+ *
+ * DESCRIPTION: This function iterates over the list of child nodes in O(n)
+ *
+ ******************************************************************************/
+
+static ACPI_NAMESPACE_NODE *
+AcpiNsSearchOneScopeCore(
+    UINT32                  TargetName,
+    ACPI_NAMESPACE_NODE     *ParentNode)
+{
+    ACPI_NAMESPACE_NODE     *Node;
+
+    ACPI_FUNCTION_TRACE (AcpiNsSearchOneScopeCore);
+
+    Node = ParentNode->Child;
+    while (Node)
+    {
+        /* Check for match against the name */
+        if (Node->Name.Integer == TargetName)
+        {
+            return Node;
+        }
+
+        /* Didn't match name, move on to the next peer object */
+        Node = Node->Peer;
+    }
+
+    return NULL;
+}
+
+
+/*******************************************************************************
+ *
  * FUNCTION:    AcpiNsSearchOneScope
  *
  * PARAMETERS:  TargetName      - Ascii ACPI name to search for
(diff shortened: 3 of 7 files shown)


sacdintel:
Hi Stanislav,

Thank you very much for this PR contribution. On the surface, it looks like a great suggestion. However, I will need to do some homework on this and discuss the future vision with a few key stakeholders and come back with some suggestions and probably even more questions, but it might take a little time and we would also want to ensure the universal compatibility of ACPICA (across most OS, platforms, architectures, C compilers etc.) is retained and moving forward how we can possibly look at speeding up other bottlenecks in ACPICA. Thanks! 

- Saket Dumbre

stanspas:
Thank you, Saket! This version of the pull request is meant as a conversation starter more than anything, so I am looking forward to your suggestions and/or questions. There is no rush from my side (and my co-author, Peter, 's side).

We are aware of the need to keep ACPICA universally compatible. The currently posted approach introduces some macros that are hopefully easy for any platform to fulfill, and a platform that does not will observe zero difference when compiling the code with or without this change.

However, Peter suggests it might be even better to implement the (very basic for the needs here) hashtable logic itself within ACPICA rather than having platforms "dependency-inject" it via the macros.

acpibob:
I’m not convinced that we need a hash table for the namespace:


  1.  We have never had any complaints about the performance of the namespace.
  2.  The Scopes that are searched are typically rather small (a few entries in the namespace).

Do you have any numbers concerning the relative  performance increase?

Bob

stanspas:
> 1.  We have never had any complaints about the performance of the namespace.

You can consider this pull request the first such "complaint" then :)
At Amazon EC2 (where I am at), updating the hypervisor involves booting a new kernel (and taking CPU time away from customer workloads to do that), often times on servers with quite complex (in terms of size/number of namespace nodes) ACPI tables. To minimize impact, we are shaving off as much of the overhead as we can to the point where even a few dozen milliseconds worth of savings do matter. This proposed ACPICA change is a part of that optimization effort.

>   2.  The Scopes that are searched are typically rather small (a few entries in the namespace).

I agree this should be the case most of the time.

> Do you have any numbers concerning the relative  performance increase?

Yes. Internally, we have a variant of this with instrumentation (not included in the pull request for now, as there are some more fundamental things to clarify first about the general approach) to run both existing search path and accelerated one, verify the results match, and gather some basic stats for the number of list element visits/iterations performed by the code.

Peter got the following data on one of our more complex machines:

Number of namespace nodes: ~34000
Number of searches performed during boot: ~407000
Iterations without the acceleration: ~27.5 million
Iterations with the acceleration (hash table size of 512 KiB): ~0.6 million
The total savings in boot time when going from existing search to accelerated search in this specific case are in the order of ~100ms

acpibob:
>>The total savings in boot time when going from existing search to accelerated search in this specific case are in the order of ~100ms
I would hardly call a savings of 1 tenth of a second to be a major performance gain.

Anyway, there is certainly some risk in replacing the existing code with a hash table. Also, we want to minimize the use of kernel memory.

Bob

stanspas:
> I would hardly call a savings of 1 tenth of a second to be a major performance gain.

I do believe there are very few use cases where saving those 100ms matters. Our use case happens to be one of them, though.

Perhaps it helps to look at the numbers in relative terms: Those ~100ms savings thanks to this optimization on that particular complex platform are out of what's currently ~1000ms of time where no customer code is able to execute on the server. There are various highly latency-sensitive workloads that care about this kind of interruption, and a 10% improvement is significant.

> Anyway, there is certainly some risk in replacing the existing code with a hash table.

Agreed. As of right now, we have sufficient confidence in the correctness of our internal variant of this pull request after internal testing of this to deploy it to production at a next opportunity. However, lack of negative side effects on our systems of course does not imply lack of negative side effects elsewhere, so we welcome feedback and concerns.

> Also, we want to minimize the use of kernel memory.

In the data sample I shared, our total memory overhead is about 1 MiB:
- 512 KiB extra memory for the table itself (array of list-heads), and
- a per-node overhead of two pointers (16 bytes on 64-bit systems) for the `SearchListNode` added to `struct acpi_namespace_node` to keep a list of nodes whose hashes collide. For our system's ~34000 nodes, that's another half-megabyte of overhead

Some ides for minimizing kernel memory usage:
- Pick the table size based on total ACPI table size, or better yet based on namespace node count (assuming we can determine the node count before needing to do namespace searches in the first place). On systems with a small namespace, allocating the table (and later using it to accelerate things) could be skipped entirely
- The hash-table could be deallocated after kernel boot completes, as ACPI namespace searches happening after that are much more rare (and less likely to be happening in hot paths).

Stanislav

sacdintel:
Hi Stanislav,

Thanks for the responses, I did have a chat with @acpibob and a few other key stakeholders who might be affected due to such a major change in ACPICA. Here are some conclusions mixed up with questions below:

We would be ready to approve and merge this PR and also hopefully talk about any other ideas for future implementations on optimizing or re-doing certain things (at least in ACPICA) for the better, but it is also important to address some concerns before some global changes to ACPICA could possibly outrage some other customers. 

We would like you and Peter to allow for an option (or a flag) so that this optimization only happens or does not happen in certain circumstances. For example, certain embedded systems and maybe even some smaller client systems would not benefit much from a hash table logic and rather the overhead of 512 KiB (+2 ptrs for each node thereafter) might be too much for such systems. So like you have mentioned before, a heuristic based algorithm that can determine smaller hash table allocations (if any) makes sense in that case which can take into account some critical factors such as total memory availability on that system, number of nodes per scope (or simply total nodes), number of ACPI tables weighted with their total size based on no. of avg entries, etc. At what point can we consider it breakeven that the advantages of a hash table start coming into play?

The other thing is those allocated hash tables need to be deallocated after a while (maybe after booting?) to the point that storing them in memory isn't worth it anymore if only a handful more searches are gonna be performed after that. Also it would be beneficial/preferable for all users of ACPICA to be able to use the faster search path, not only Linux users based on the macro usages (although I believe most servers utilize some sort of Linux-based OS along with ACPICA). So in conclusion, if you and Peter can implement an ACPICA native hash table logic with almost the same level of optimizations as the Linux hash table, it would make this PR a really significant contribution to our ACPICA project. You see there are licensing issues with using the Linux implementation of hash table in ACPICA when used with other (non-GPL) licenses of ACPICA itself and/or also the host OS, so we need to tackle that as well. Overall, since ACPICA is such a universal project, it is a strong preference and almost a requirement that whatever contributions are made to ACPICA, they should be able to benefit everyone regardless of the platform, OS, Architecture, vendor etc.

Sorry for the long comment, please take your time and I will be happy to discuss things further and/or answer any questions. Thank you very much!

stanspas:
Hello,

I just wanted to let you know that Peter is internally engaged with any additional work necessary to make this a worthy contribution, but unfortunately not in a dedicated or highly prioritized manner, so expect delays.

Stanislav

sacdintel:
Hi Stanislav,

Thanks for the update! We are in no hurry either, so please feel free to continue the great work! In the meantime, please let us know if any unforeseen difficulties are encountered as I don't expect implementing the hash table logic from scratch would be an easy task! Thanks!

-Saket

paulmenzel:
@stanpas, nice work. Do you have public branch, from which I can easily fetch this for the Linux kernel?

source/components/namespace/nsalloc.c
```diff
@@ -237,6 +237,10 @@ AcpiNsDeleteNode (
         return_VOID;
     }
 
+#ifdef ACPI_USE_ACCELERATED_NS_SEARCH
```
d-tatianin, line 240: This should probably be `ACPI_USE_NS_SEARCH_ACCELERATION` :)

SaketADumbre: Good catch.

SaketADumbre:
I think this PR can move forward to be merged with a couple of changes to be made first: By default, we keep linear search due to the overhead of 0.5 MB for the hashmap, whose allocation will succeed on client systems, but won't be justifiable because our release files themselves are a few MB (<3-4 MB). I understand on the server side, the 10% acceleration is a great idea, but on client systems it is not necessary for a few nodes (a couple hundred if even that) and I am opposed to making our releases "bloatware". The other change would be testing its correctness as it looks like that is not the case which @d-tatianin points out above in the code review! I have changed my mind on the topic of "blanket fairness" which means not leaving anyone behind as having 0.5 MB in-built ACPICA hashmap would make us "bloatware" and a Linux-based API solution is now preferable. Other OS users can later on add support with their own hash API implementations, if they also feel the advertised acceleration would benefit their server side OS! Thanks a lot!

paulmenzel:
Review by Claude Opus 4.8:

> Review of stanspas/accelerate-nssearch
>
> The branch adds an optional hash-table fast path to AcpiNsSearchOneScope, keyed on parent_ptr ^ name, gated behind ACPI_USE_NS_SEARCH_ACCELERATION (Linux-kernel-only). The design is sound; findings, most severe first:    
>
> 1. CRITICAL — delete hook compiled out → use-after-free. nsalloc.c guards the removal with #ifdef ACPI_USE_ACCELERATED_NS_SEARCH, but the feature macro is ACPI_USE_NS_SEARCH_ACCELERATION. So AcpiNsSearchAccelerationDeleteNode is never called — freed nodes stay linked in their bucket, and the next search hashing there walks freed memory. AML method execution churns temporary nodes constantly, so this triggers fast. This is the ship-blocker.                           
>
> 2. Logic/perf — fallback fires on every miss. if (!Node) Node = ...Core(...) re-runs the full linear scan on every genuine "not found." Since namespace load searches-before-insert, every insert pays both a hash miss and an O(n) scan — negating the acceleration on the hot path it targets. Should branch on AcpiNsSearchHashTable != NULL, not on the result. This also currently masks correctness gaps (e.g. #1's stale entries), so #1 and #2 interact — you can't tighten #2 without full add/delete coverage.
>
> 3. Minor. AcpiNsSearchAccelerationInit always returns AE_OK, so the ACPI_FAILURE check in AcpiNsRootInitialize is dead — and worse, its return_ACPI_STATUS on failure would leak the namespace mutex (I used goto unlock_and_exit in the port). "%ud bytes" format typo. Fixed HashBits = 10 with a TODO. Trace macros (ACPI_FUNCTION_TRACE) used in helpers that return a pointer directly would imbalance debug nesting.

PR #901, opened 2023-10-23
Add option to make output deterministic

jjd27:
The output from iasl -d included a timestamp. This meant that each execution returned different output. To provide a means of making the output deterministic, use the value of the SOURCE_DATE_EPOCH environment variable in preference to the current time.

See https://reproducible-builds.org/specs/source-date-epoch/ for background on the environment variable.

### DIFF
--- a/source/common/dmtables.c
+++ b/source/common/dmtables.c
@@ -190,7 +190,7 @@ extern ACPI_PARSE_OBJECT    *AcpiGbl_ParseOpRoot;
  * RETURN:      None
  *
  * DESCRIPTION: Create the disassembler header, including ACPICA signon with
- *              current time and date.
+ *              optional current time and date.
  *
  *****************************************************************************/
 
@@ -201,8 +201,6 @@ AdDisassemblerHeader (
 {
     time_t                  Timer;
 
-
-    time (&Timer);
 
     /* Header and input table info */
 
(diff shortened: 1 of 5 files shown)


aveselyy:
I am doing a similar thing in #902, but instead of an env. variable I add a flag to completely suppress date printing. I'm not sure which one is preferable from the code maintainers' perspective.

acpibob:
Yes, a command-line flag would be preferable to an environment variable.

jlevon:
A flag means that every single project that needs to be built reproducibly has to have code changes. SOURCE_DATE_EPOCH can be set across a build globally without needing to fork upstream repositories and so on.

If the original suggestion (the flag actually enables writing out the date) was implemented, that'd be fine though.

acpibob:
This sounds ok to me (a command-line flag to enable the timestamping.  Please use -v (verbosity), and  the "d" (date) subcommand to enable timestamping.

acpibob:
So, what is it exactly that you are doing that you need identical compiler output (I presume it is a listing file)?
This feature has been requested before, but we've never done it.

jlevon:
If you mean why is it useful at all, then please see the explanation on https://reproducible-builds.org/ - which explains it way better than I could

If you mean specifically our immediate need: we have some regression tests that ACPI content doesn't change, and are currently having to work around the date part here.

aveselyy:
I'm also ok with either solution. Variable or flag, on by default or off by default.

jjd27:
Thanks all for the input and suggestions, @acpibob, @sacdintel, @jlevon, @aveselyy.

> Please use -v (verbosity), and the "d" (date) subcommand to enable timestamping.

That one's already taken (for printing the timestamp at which the tool was compiled). In fact all the best ones in the `-v` namespace seem to be taken...

I like the idea of combining forces with https://github.com/acpica/acpica/pull/902; so I've rebased this PR on a lightly modified version of that patch, which used the `-ld` option.

aveselyy:
Thanks! This looks good.

aveselyy:
@acpibob Is there a timeline when this PR can be expected to get merged?

sacdintel:
Looks good to me, sorry for the delay, we have been incredibly busy with other things and also covering for each other's leave periods! Great teamwork on this PR BTW!

PR #902, opened 2023-10-24
Fix build indeterminism for asm compiler header.

aveselyy:
Existing code puts a timestamp in the .lst files, e.g.: 
Compilation of ".../src/fw/ssdt-misc.i" - Thu Oct 22 13:38:41 2023 
This makes the build non-deterministic.

### DIFF
--- a/source/compiler/aslcompile.c
+++ b/source/compiler/aslcompile.c
@@ -655,6 +655,8 @@ void
 AslCompilerFileHeader (
     UINT32                  FileId)
 {
+    char                    *NewTime;
+    time_t                  Aclock;
     char                    *Prefix = "";
 
 
@@ -695,11 +697,25 @@ AslCompilerFileHeader (
         break;
     }
 
-    /* Compilation header */
+    /* Compilation header (with timestamp) */
 
     FlPrintFile (FileId,
         "%sCompilation of \"%s\"",
         Prefix, AslGbl_Files[ASL_FILE_INPUT].Filename);
+
+    if (!AslGbl_Deterministic) 
+    {
+        Aclock = time (NULL);
+        NewTime = ctime (&Aclock);
+        if (NewTime)
+        {
+            FlPrintFile (FileId, " - %s%s\n", NewTime, Prefix);
+        }
+    }
+    else 
+    {
+        FlPrintFile (FileId, "\n");
+    }
 
     switch (FileId)
     {
--- a/source/compiler/aslcompile.c
+++ b/source/compiler/aslcompile.c
@@ -655,8 +655,6 @@ void
 AslCompilerFileHeader (
     UINT32                  FileId)
 {
-    char                    *NewTime;
-    time_t                  Aclock;
     char                    *Prefix = "";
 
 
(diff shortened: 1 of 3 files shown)


acpibob:
Could you add a new flag to the compiler to force the output of the time/date information?

Thanks,
Bob


From: aveselyy ***@***.***>
Sent: Tuesday, October 24, 2023 12:47 PM
To: acpica/acpica ***@***.***>
Cc: Subscribed ***@***.***>
Subject: [acpica/acpica] Fix build indeterminism for asm compiler header. (PR #902)


Existing code puts a timestamp in the .lst files, e.g.:
Compilation of ".../src/fw/ssdt-misc.i" - Thu Oct 22 13:38:41 2023
This makes the build non-deterministic.

________________________________
You can view, comment on, or merge this pull request online at:

  https://github.com/acpica/acpica/pull/902

Commit Summary

  *   f82e83d<https://github.com/acpica/acpica/pull/902/commits/f82e83d24cfa46d199f47715e62845559d6c70c7> Fix build indeterminism for asm compiler header.

File Changes

(1 file<https://github.com/acpica/acpica/pull/902/files>)

  *   M source/compiler/aslcompile.c<https://github.com/acpica/acpica/pull/902/files#diff-98b3bf0f3050d372c3f4052c3f0d2e633a753c44958b5cd20003886f473b38e9> (14)

Patch Links:

  *   https://github.com/acpica/acpica/pull/902.patch
  *   https://github.com/acpica/acpica/pull/902.diff

aveselyy:
Updated with a flag inside the "-l" block, old behavior is active by default.

aveselyy:
@acpibob Anything else I should fix?

sacdintel:
Not really at this time, but we might need a little more time to review this PR and merge it! I had a question in the meantime if the timer implementation in PR #901 tries to do the same thing as this PR (but for the disassembler instead of the compiler) and if possible can you please check it and comment there to see if a collaboration with the contributor is possible so that the flag is used not by default for the disassembler too? It wouldn't be ideal to have duplicated code essentially fixing the same functionality. Thanks a lot!
- Saket

sacdintel:
I don't know which method does @acpibob prefer but I would presume that setting a flag is preferable over an environment variable. Thanks!

sacdintel:
So yeah, a command line flag is preferable to an env. var. according to Bob as well! Thanks!

aveselyy:
So what would the next steps be? Can this PR be merged? Should I invert the logic of the flag? 
Also @jjd27 @jlevon

jjd27:
Hi @aveselyy, it sounds like the consensus is to use a command-line flag, and that there's a slight preference for determinism by default.

The suggestion in https://github.com/acpica/acpica/pull/901#issuecomment-1789681841 was to use command-line option `-vd` but that's already taken. I feel like something in the `-v` namespace would work better than `-ld` but it doesn't bother me.

I've taken the liberty of updating https://github.com/acpica/acpica/pull/901 to take what you propose in this PR, with three changes:
* Added help text to source/compiler/aslhelp.c.
* Inverted the sense of the flag (deterministic output on by default; `-ld` disables it).
* Your four commits squashed into one.

aveselyy:
Thanks! Sounds good.

Issue #827, opened 2023-01-13
Handling of Integers in ToHexString() problematic


Wer-Wolf:
Sadly, the ACPI specification does only vaguely define the format of the string produced by ToHexString().
As for the handling of integers with ToHexString(), i observed that the ACPI implementation of Microsoft Windows does implement ToHexString() differently than ACPICA:

Windows:
byte 0x0f =>"0xf"
word 0x0f0f => "0xf0f"
dword 0x0f0f0f0f => "0xf0f0f0f"
qword 0x0f0f0f0f0f0f0f0f => "0xf0f0f0f0f0f0f0f"

ACPICA:
byte 0x0f =>"0000000f"
word 0x0f0f => "00000f0f"
dword 0x0f0f0f0f => "0f0f0f0f"
qword 0x0f0f0f0f0f0f0f0f => "0f0f0f0f" (32-bit table, but qword is 64-bit?)

This causes problems on my Inspiron 3505, as the battery serial number generated by using ToHexString(<16 bit field>, Local1) and Mid(Local1, 0x02, 0x04, Local0), causing a wrong serial number to be displayed.

I already wrote a email to the ASWG suggesting clarifying the string format of ToHexString() in a future revision of the ACPI standard,
but i am still waiting for a reply.

Could the behavior of ToHexString() for integers be changed to match the Windows behavior? Something similar was already done in version 20181213.

sacdintel:
When your PR 903 merges, does it resolve this issue 827? Thanks.

Wer-Wolf:
Yes

sacdintel:
Completed via PR 903 merge.

PR #903, opened 2023-10-28
Modify Integer handling in ToHexString()

Wer-Wolf:
This series changes the handling of integers in ToHexString(), see Issue #827 for details.
The following ASL code was used to test the behavior of ToHexString() with integer arguments under both Windows and ACPICA: [ssdt.txt](https://github.com/acpica/acpica/files/13194389/ssdt.txt).

The output of \_SB.TE08, \_SB.TE16, \_SB.TE32 and \_SB.TE64 was verified to be the same under Windows and ACPICA
except when encountering 64 bit constants in 32 bit tables. In such a case, Windows would not truncate the output of \_SB.TE64,
but ACPICA did.

Since using 64 bit integers in a 32 bit table is unsupported anyway and the chance of encountering this in the wild are
rather low, i decided to ignore this special case.

As a side note, something similar was already done in commit f9efaa885dc3 ("ToHexString: Add "0x" prefix to the converted string
"), which was later reverted for different reasons.

### DIFF
--- a/source/components/executer/exconvrt.c
+++ b/source/components/executer/exconvrt.c
@@ -437,6 +437,7 @@ AcpiExConvertToAscii (
     UINT32                  DecimalLength;
     UINT32                  Remainder;
     BOOLEAN                 SupressZeros = !LeadingZeros;
+    UINT8                   HexChar;
 
 
     ACPI_FUNCTION_ENTRY ();
@@ -503,8 +504,18 @@ AcpiExConvertToAscii (
         {
             /* Get one hex digit, most significant digits first */
 
-            String[k] = (UINT8)
+            HexChar = (UINT8)
                 AcpiUtHexToAsciiChar (Integer, ACPI_MUL_4 (j));
+
+            /* Supress leading zeros until the first non-zero character */
+
+            if (HexChar == ACPI_ASCII_ZERO && SupressZeros)
+            {
+                continue;
+            }
+
+            SupressZeros = FALSE;
+            String[k] = HexChar;
             k++;
         }
         break;
--- a/source/components/executer/exconvrt.c
+++ b/source/components/executer/exconvrt.c
@@ -165,7 +165,8 @@ AcpiExConvertToAscii (
     UINT64                  Integer,
     UINT16                  Base,
     UINT8                   *String,
-    UINT8                   MaxLength);
+    UINT8                   MaxLength,
+    BOOLEAN                 LeadingZeros);
 
 
 /*******************************************************************************
@@ -412,6 +413,7 @@ AcpiExConvertToBuffer (
  *              Base            - ACPI_STRING_DECIMAL or ACPI_STRING_HEX
  *              String          - Where the string is returned
  *              DataWidth       - Size of data item to be converted, in bytes
+ *              LeadingZeros    - Allow leading zeros
  *
  * RETURN:      Actual string length
  *
@@ -424,7 +426,8 @@ AcpiExConvertToAscii (
     UINT64                  Integer,
     UINT16                  Base,
     UINT8                   *String,
-    UINT8                   DataWidth)
+    UINT8                   DataWidth,
+    BOOLEAN                 LeadingZeros)
 {
     UINT64                  Digit;
     UINT32                  i;
@@ -433,7 +436,7 @@ AcpiExConvertToAscii (
     UINT32                  HexLength;
     UINT32                  DecimalLength;
     UINT32                  Remainder;
-    BOOLEAN                 SupressZeros;
+    BOOLEAN                 SupressZeros = !LeadingZeros;
 
 
     ACPI_FUNCTION_ENTRY ();
@@ -464,7 +467,6 @@ AcpiExConvertToAscii (
             break;
         }
 
-        SupressZeros = TRUE;     /* No leading zeros */
         Remainder = 0;
 
         for (i = DecimalLength; i > 0; i--)
@@ -556,6 +558,7 @@ AcpiExConvertToString (
     UINT32                  StringLength = 0;
     UINT16                  Base = 16;
     UINT8                   Separator = ',';
+    BOOLEAN                 LeadingZeros;
 
 
     ACPI_FUNCTION_TRACE_PTR (ExConvertToString, ObjDesc);
@@ -581,6 +584,7 @@ AcpiExConvertToString (
              * Make room for the maximum decimal number size
              */
             StringLength = ACPI_MAX_DECIMAL_DIGITS;
+            LeadingZeros = FALSE;
             Base = 10;
             break;
 
@@ -589,6 +593,7 @@ AcpiExConvertToString (
             /* Two hex string characters for each integer byte */
 
             StringLength = ACPI_MUL_2 (AcpiGbl_IntegerByteWidth);
+            LeadingZeros = TRUE;
             break;
         }
 
@@ -607,7 +612,7 @@ AcpiExConvertToString (
         /* Convert integer to string */
 
         StringLength = AcpiExConvertToAscii (
-            ObjDesc->Integer.Value, Base, NewBuf, AcpiGbl_IntegerByteWidth);
+            ObjDesc->Integer.Value, Base, NewBuf, AcpiGbl_IntegerByteWidth, LeadingZeros);
 
         /* Null terminate at the correct place */
 
@@ -628,6 +633,7 @@ AcpiExConvertToString (
              * From ACPI: "If the input is a buffer, it is converted to a
              * a string of decimal values separated by commas."
              */
+            LeadingZeros = FALSE;
             Base = 10;
 
             /*
@@ -661,6 +667,7 @@ AcpiExConvertToString (
              *
              * Each hex number is prefixed with 0x (11/2018)
              */
+            LeadingZeros = TRUE;
             Separator = ' ';
             StringLength = (ObjDesc->Buffer.Length * 5);
             break;
@@ -674,6 +681,7 @@ AcpiExConvertToString (
              *
              * Each hex number is prefixed with 0x (11/2018)
              */
+            LeadingZeros = TRUE;
             Separator = ',';
             StringLength = (ObjDesc->Buffer.Length * 5);
             break;
@@ -715,7 +723,7 @@ AcpiExConvertToString (
             }
 
             NewBuf += AcpiExConvertToAscii (
-                (UINT64) ObjDesc->Buffer.Pointer[i], Base, NewBuf, 1);
+                (UINT64) ObjDesc->Buffer.Pointer[i], Base, NewBuf, 1, LeadingZeros);
 
             /* Each digit is separated by either a comma or space */
 
--- a/source/components/executer/exconvrt.c
+++ b/source/components/executer/exconvrt.c
@@ -599,6 +599,15 @@ AcpiExConvertToString (
             Base = 10;
             break;
 
+        case ACPI_EXPLICIT_CONVERT_HEX:
+            /*
+             * From ToHexString.
+             *
+             * Supress leading zeros and append "0x"
+             */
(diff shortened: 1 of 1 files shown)


Wer-Wolf:
Any feedback on this patch series?

sacdintel:
Hi Armin,

This same topic is an ongoing discussion as per my emails and it looks like Windows truncates the preceding 0's while Linux keeps the leading zeros. Do you know which way does the implementation go under OpenBSD? We might not be able to do anything until we are clear on some things and which behavior would be accepted. Thanks for your patience and contribution!

-Saket

Wer-Wolf:
AFAIK, OpenBSD uses ACPICA too, so the behavior should be the same as Linux.

Since ACPICA aims for compatibility with other ACPI implementations, i advocate to change the behavior of ToHexString().

After all, something similar was done in 2016 already, and it only got reverted because the patch accidently affected conversations other than ToHexString().

Wer-Wolf:
Ok, it seems i was wrong about OpenBSD and ACPICA. I will check.

Wer-Wolf:
Ok, according to the code inside the function `aml_convert()`, they use `snprintf(c->v_string, c->length, (ctype == AML_OBJTYPE_HEXSTRING) ? "0x%llx" : "%lld", a->v_integer);`, which effectively also strips leading zeros when using ToHexString() and appends "0x".

sacdintel:
Thanks a lot for pointing that out! Right now, the ongoing discussion revolves around which behavior should be the default regardless of the underlying OS implementations and other ACPI implementations (notably Windows) and possible codifying that language in the ACPI specification itself. We might need to take further steps depending on the outcome of the ongoing public discussions, but we will get back to this PR as soon as we hear about the decisions made. Thanks for your patience in the meantime!

Wer-Wolf:
Ok, i assume the discussion happens inside the ACPI Working Group?
Is there any way to participate in this discussion?

sacdintel:
Yes at ASWG and also discussions internally at Intel but I highly doubt anyone else can participate in ASWG meetings other than the representatives or full time employees of the representing member companies of UEFI.

Wer-Wolf:
I already send an email to the chair of the ASWG, asking for the string format of ToHexString() to be codified inside the ACPI spec a long time ago.

I suspect it got ignored...

acpibob:
Our bottom line is that we (ACPICA) must be Windows compatible – and the ACPI spec can be changed to match the Windows behavior. In the beginning, we said that ACPICA was a “reference” ACPI implementation. We were forced to make it a “Windows Compatible” implementation.

Wer-Wolf:
I know, and this patch series allows ToHexString() to be just that: being compatible with the Windows implementation.

I agree that documenting the string format of ToHexString() inside the ACPI spec would be great, since this would prevent such issues in the future.

sacdintel:
I agree with what Bob and you just said, but we just need to clarify a few things to go ahead and merge this PR so that we aren't seen, from our end, as the sole decision makers (breaking alignments with OS and Firmware teams industry wide) in this regard as such a change could possibly break any Linux drivers currently reading the existing behavior of any Hex Strings. We try our best to make things experience a smoother transition so that everyone who uses that functionality are made aware beforehand and adjust the expected behavior of their products/services too and don't get spooked out! Thanks!

acpibob:
While we are at it, could you re-post a description of the original problem?
Thanks,
Bob

Wer-Wolf:
Sadly, the ACPI specification does only vaguely define the format of the string produced by ToHexString().
As for the handling of integers with ToHexString(), i observed that the ACPI implementation of Microsoft Windows does implement ToHexString() differently than ACPICA:

Windows:
byte 0x0f =>"0xf"
word 0x0f0f => "0xf0f"
dword 0x0f0f0f0f => "0xf0f0f0f"
qword 0x0f0f0f0f0f0f0f0f => "0xf0f0f0f0f0f0f0f"

ACPICA:
byte 0x0f =>"0000000f"
word 0x0f0f => "00000f0f"
dword 0x0f0f0f0f => "0f0f0f0f"
qword 0x0f0f0f0f0f0f0f0f => "0f0f0f0f" (32-bit table, but qword is 64-bit?)

This causes problems on my Inspiron 3505, as the battery serial number generated by using ToHexString(<16 bit field>, Local1) and Mid(Local1, 0x02, 0x04, Local0), causing a wrong serial number to be displayed.

I already wrote a email to the ASWG suggesting clarifying the string format of ToHexString() in a future revision of the ACPI standard,
but i am still waiting for a reply.

Could the behavior of ToHexString() for integers be changed to match the Windows behavior? Something similar was already done in version 20181213.

acpibob:
We can possibly change the behavior to match windows. However, what does the original ASL look like?

The behavior you are seeing looks like the automatic truncation (by the AML interpreter) of 64-bit numbers to 32-bits in a 32-bit table. In other words, all 64-bit numbers are disallowed in a 32-bit table. If windows is allowing 64-bit values in a 32-bit table, then I am afraid this goes against the ACPI specification

Wer-Wolf:
[acpidump.log](https://github.com/acpica/acpica/files/13254066/acpidump.log)

The case with the 64 bit constant inside a 32 bit table only exists inside my custom SSDT for testing, the DSDT on the real machine calls ToHexString() on a 16 bit integer, so this should not be an issue.

The patch series also does not follow the Windows behavior regarding 64 bit integers in 32 bit tables for exactly this reason. The scenario of encountering an 64 bit value inside an 32 bit table is unsupported, the output of ToHexString() will get truncated in such a case.

acpibob:
>encountering an 64 bit value inside an 32 bit table is unsupported, the output of ToHexString() will get truncated in such a case.
So, what is the problem that you are seeing (other than the ACPI spec being vague?)

Wer-Wolf:
The problem is that the battery serial number is displayed wrong. On Windows, the result of `ToHexString()` is `0x20cd`, meaning that after executing `Mid(Local1, 0x02, 0x04, Local0`, the resulting battery serial number is `20cd`.

On Linux however, the result of `ToHexString()` is `000020cd`, which means that the resulting serial number is `0020`.

Wer-Wolf:
As a fun side note: the firmware on this machine has some bugs because the Windows ACPI implementation seems to be susceptible to buffer overflows when converting buffers into strings. If the buffer is not NUL-terminated, the Windows implementation will read past the end of the buffer.

ACPICA however is not affected by this, so it truncates the result because the buffer size was specified two byte too short inside the DSDT of the machine. I wont suggest adding this behavior to ACPICA for obvious reasons.

acpibob:
> On Linux however, the result of ToHexString() is 000020cd, which means that the resulting serial number is 0020.
It sounds like ToHexString is working correctly (except for a missing 0x prefix).

Wer-Wolf:
Not exactly, the leading zeros need to be striped too.

acpibob:
>Not exactly, the leading zeros need to be striped too

Why?

Also, what are you using to output the result of ToHexString?

sacdintel:
Should the final result after mid be 20cd in your example under Linux rather than 0020? The 0x can be ignored since it just means that the format is Hex, not decimal

Wer-Wolf:
The final result should be `20cd`.

Wer-Wolf:
And the leading zeros need to be striped because otherwise the result after `Mid()` would still be `0000`, which is wrong.

I tested this by modifying the ACPI method in question to operate on constant integer values instead of the integer value from the battery controller.

Wer-Wolf:
On Linux, i verified the output of `ToHexString()` directly using `acpiexec`. On Windows, i again patched the DSDT to directly display the output of `ToHexString()` as the battery serial number.

acpibob:
>On Linux however, the result of ToHexString() is 000020cd, which means that the resulting serial number is 0020.
This sounds like the value is being treated as a decimal number (because there is no 0x prefix).

Wer-Wolf:
I dont think so, the `cd` in `000020cd` shows that it is treated as a hex number.

Wer-Wolf:
Also appending `0x` to hex numbers what added in commit f9efaa885dc3fbdd73307e3748c096ec0849f7a5 ("ToHexString: Add "0x" prefix to the converted string"), which was later reverted.

acpibob:
The revert would indicate to me that adding 0x broke something.

Wer-Wolf:
Yes, because it affected other conversations. This patch series avoids that so that only the output of `ToHexString()` with an integer argument is changed.

Wer-Wolf:
See ec434a8 ("ToHexString: Add "0x" prefix to the converted string") for details.

acpibob:
>This patch series avoids that so that only the output of ToHexString() with an integer argument is changed.
Do you mean an integer constant, or a named object containing an integer?

Wer-Wolf:
Everything which does `ACPI_EXPLICIT_CONVERT_HEX`, which is the handler for `ToHexString()`. Wether the integer is a constant or a named object or a reference is irrelevant, it is resolved to an `ACPI_OPERAND_OBJECT` before the code in question handles it.

acpibob:
We will need to figure out why the change was reverted. It may have broken our test suite, or it may have broken some machines (worse).

Wer-Wolf:
I think to was meant to only affect the results of `ToHexString()` with an integer argument, but it accidentally also affected the results of `ToHexString()` with an buffer argument, because both use `AcpiExConvertToAscii()`, where the change was made.

This patch series avoids that by appending the `0x` outside of `AcpiExConvertToAscii()`, so that other users are not affected.
Additionally, the stripping of leading zeros when converting base 16 integers with `AcpiExConvertToAscii()` can be controlled with a boolean flag, so current users aside from `ACPI_EXPLICIT_CONVERT_HEX` are not affected.

Wer-Wolf:
Any progress with the discussion inside the ACPI work group?

Wer-Wolf:
Ok, have a nice vacation!

Wer-Wolf:
I hope i do not annoy, but how is the status of this ECR?

Wer-Wolf:
In the meantime i found yet another machine which is affected by this.

Can i do anything to accelerate this discussion inside intel?

sacdintel:
Hi Armin,

Sorry for the delayed response, I am just back to work last week after 6 weeks of bonding leave to care for my newborn. You are good, haven't annoyed me or anything. While I was on my leave, I internally presented almost 13 possible ECRs including all of @d-tatianin 's issues reported regarding differences in behavior between ACPICA, the ACPI spec and acpi.sys and yours was part of it! Some folks need more time to review all my suggestions but I am hoping in the coming weeks I can move ahead presenting them at ASWG and hopefully getting approval for them all! From my end, I will definitely be merging this PR whether or not the ACPI spec is updated. Thanks for your patience!

Wer-Wolf:
Thank you so much, i am overjoyed to hear that the discussion is moving forward.

sacdintel:
I did manage to get this PR (fix for ToHexString() in ASL) approved, I will merge it soon when I am back in the office on Monday and I am also in the process of drafting several ECRs as we speak (one for this PR which clarifies and restricts the format of ToHexString() in ASL) which will be presented to the ASWG meeting in the next few weeks and will update you and @d-tatianin of the status of each one, thanks a lot!

Wer-Wolf:
I couldn't be happier, thank you for pursuing this issue! :)

source/components/executer/exconvrt.c
```diff
@@ -501,8 +504,17 @@ AcpiExConvertToAscii (
         {
             /* Get one hex digit, most significant digits first */
 
-            String[k] = (UINT8)
+            HexChar = (UINT8)
                 AcpiUtHexToAsciiChar (Integer, ACPI_MUL_4 (j));
+
+            /* Supress leading zeros until the first non-zero character */
+
+            if (HexChar == ACPI_ASCII_ZERO && SupressZeros) {
```
sacdintel, line 512: Could you just move the opening brace to the next blank line in order to be compatible with the ACPICA coding style, thanks!

Wer-Wolf: Done

PR #905, opened 2023-10-29
os_specific: zephyr: update AcpiOsGetTimer with arch or platform specific implementation

najumon1980:
update AcpiOsGetTimer implementation with arch/platform specific implementation (acpi_timer_get) instead of using sys_clock_cycle_get_64 which might internally use hpet timer etc. if it enabled as system timer and can cause raise condition if AcpiOsGetTimer invoked before those timer driver instantiate.

### DIFF
--- a/source/os_specific/service_layers/oszephyr.c
+++ b/source/os_specific/service_layers/oszephyr.c
@@ -967,9 +967,8 @@ UINT64
 AcpiOsGetTimer (
     void)
 {
-    return (k_cycle_get_64 ());
+    return acpi_timer_get();
 }
-
 
 /******************************************************************************
  *


source/os_specific/service_layers/oszephyr.c
```diff
@@ -967,7 +967,10 @@ UINT64
 AcpiOsGetTimer (
     void)
 {
-    return (k_cycle_get_64 ());
+    uint32_t TimerHi, TimerLo;
+
+    __asm__ volatile("rdtsc" : "=d"(TimerHi), "=a"(TimerLo));
```
tbursztyka, line 972: again you are making this x86 specific. 

Let's keep k_cycle_get_64().
And if we have such race condition, it means we are not initializing hpet (or any other timer actually) early enough. So it's a zephyr side issue, not an acpica one.

tbursztyka (changes requested):

source/os_specific/service_layers/oszephyr.c
```diff
@@ -967,10 +967,9 @@ UINT64
 AcpiOsGetTimer (
     void)
 {
-    return (k_cycle_get_64 ());
+    return acpi_timer_get();
```
jhedberg, line 970: Where is `acpi_timer_get()` defined? I can't find it anywhere (neither in the Zephyr tree nor in the acpica tree).

jhedberg: > Where is `acpi_timer_get()` defined? I can't find it anywhere (neither in the Zephyr tree nor in the acpica tree).


Also, the commit message talks about tsc timer and pet timer, but this is inconsistent with the content of the commit. I also agree with Tomasz that we shouldn't make assumptions about which timer implementation is available in Zephyr (or what architecture Zephyr is using)

najumon1980: > Where is `acpi_timer_get()` defined? I can't find it anywhere (neither in the Zephyr tree nor in the acpica tree).

This Zephyr PR is : https://github.com/zephyrproject-rtos/zephyr/pull/64570

najumon1980: > > Where is `acpi_timer_get()` defined? I can't find it anywhere (neither in the Zephyr tree nor in the acpica tree).
> 
> Also, the commit message talks about tsc timer and pet timer, but this is inconsistent with the content of the commit. I also agree with Tomasz that we shouldn't make assumptions about which timer implementation is available in Zephyr (or what architecture Zephyr is using)

acpi_timer_get() is included via acpi_osal.h (which expected to implement per Arch) and currently it enabled/implemented only for x86 (zephyr/acpi/x86_acpi_osal.h)

Updated commit message/heading.

jhedberg (approved):
This looks fine to me, and I've also seen in practice that `k_cycle_get_64()` simply doesn't work with certain system timer drivers like HPET.

@tbursztyka @finikorg please take a look as well.

finikorg (approved):

jhedberg:
@sacdintel this should be ready to be merged, thanks

tbursztyka (approved):

jhedberg:
> @sacdintel this should be ready to be merged, thanks

ping. We'd like to sync our downstream fork of acpica once this has been merged.

PR #911, opened 2023-11-15
iASL: Add support for the AEST V2 table

winterddd:
AEST V2 was published [1], add V2 support based on AEST V1.

[1]: https://developer.arm.com/documentation/den0085/latest/


sacdintel:
Please also don't forget to update the table template under dttemplate.h and please address my concern regarding the coding style mismatch in the code review, thanks a lot!

sacdintel:
Updating the template is a big hint (although not a hard evidence) to me that the PR for another (non-x86 i.e. not my system) architecture works fine, meaning the advertised table changes are compiled and disassembled correctly as intended since we don't have the means to test a system with every possible permutation and combination of OS, platform, architecture, C compiler etc. that is supported by ACPI.

winterddd:
update:
- update AEST template
- follow ACPICA coding style

sacdintel:
Looks good now, thanks a lot! Ready to merge it!

PR #890, opened 2023-08-28
NHLT: Drop preliminary support

crojewsk-intel:
Continuation of #881.

Goal of the patchset is to provide clear, readable and actual representation of Non HDAudio Link Table within ACPICA. Specification on which current implementation is based on, does not list all important types and does not inform a reader how things changed throughout the years, unfortunately. Because of numerous conflicts, first patch drops the iASL support entirely for the table ~~and the follow up just re-introduces the types and constants~~. As the work to re-add iASL support is non-trivial and will take more time, decided do this in two-step fashion - add types first, then provide NHLT support in iASL in incremental fashion. This will allow for linux-kernel related changes to be unblocked sooner than later.

Non HDAudio Link Table (NHLT) is designed to separate hardware-related description (registers) from AudioDSP firmware-related one i.e.: pipelines and modules that together make up the audio stream on Intel DSPs. This task is important as same set of hardware registers can be used with different topologies and vice versa, same topology could be utilized with different set of hardware. As the hardware registers description is directly tied to specific platform, intention is to have such description part of low-level firmware e.g.: BIOS.

The initial design has been provided in early Skylake (SKL) days. The audio architecture goes by the name cAVS. SKL is a representative of cAVS 1.5. The table helps describe endpoint capabilities ever since. While Raptorlake (RPL) is the last of cAVS architecture - cAVS 2.5 to be precise - its successor, the ACE architecture which begun with Meteorlake (MTL) inherited the design for all I2S and PDM configurations. These two configurations are the primary targets for NHLT table.

~~As the details changed throughout the years, the re-introduction is stripped off members which are OS-specific or were introduced temporarily.~~

PR #881, opened 2023-07-04
NHLT: Declare device configuration types

crojewsk-intel:
Device configuration structures are plenty so declare a struct for each known variant. While these kind of duplicate few types present in actbl2.h already, change is motivated by usage improvements - simplicity and shorten wording. Intention is to have them replacing the existing members in the future.

PR #912, opened 2023-11-20
NHLT: Reintroduce types the table consists of

crojewsk-intel:
Followup to PR #890. Paired with [Add NHLT table specification](https://github.com/UEFI/ACPI_spec/pull/150) for ACPI_spec.

Non HDAudio Link Table (NHLT) is designed to separate hardware-related description (registers) from AudioDSP firmware-related one i.e.: pipelines and modules that together make up the audio stream on Intel DSPs. This task is important as same set of hardware registers can be used with different topologies and vice versa, same topology could be utilized with different set of hardware. As the hardware registers description is directly tied to specific platform, intention is to have such description part of low-level firmware e.g.: BIOS.

The initial design has been provided in early Sky Lake (SKL) days. The audio architecture goes by the name cAVS. SKL is a representative of cAVS 1.5. The table helps describe endpoint capabilities ever since. While Raptor Lake (RPL) is the last of cAVS architecture - cAVS 2.5 to be precise - its successor, the ACE architecture which begun with Meteor Lake (MTL) inherited the design for all I2S and PDM configurations. These two configurations are the primary targets for NHLT table.

The second patch updates acpisrc utility to account for all the NHLT types.


crojewsk-intel:
After re-evaluating the spec, the definition of "device configs" structures could indeed be simplified. That's what I did here. I believe the suggestion came from @andy-shev suggested earlier, during the review of PR #881.
```
struct acpi_nhlt_config {
    u32 capabilities_size;
    u8 capabilities[];
};

/* Below types only expand the 'capabilities' field above. */
struct acpi_nhlt_gendevice_config {
    /* no capabilites_size */
    (...)
};
struct acpi_nhlt_micdevice_config {
    /* no capabilites_size */
    (...)
};
struct acpi_nhlt_vendor_micdevice_config {
    /* no capabilites_size */
    (...)
};

union acpi_nhlt_device_config {
    u8 virtual_slot;
    struct acpi_nhlt_gendevice_config gen;
    struct acpi_nhlt_micdevice_config mic;
    struct acpi_nhlt_vendor_micdevice_config vendor_mic;
};
```
The previous implementation was similar, but all the structs included `capabilities_size` into their definition.

crojewsk-intel:
Changes in v2:
- reworded "acpi_wave_formatext" to "acpi_nhlt_wave_formatext" i.e.: missing "nhlt" prefix
- reworded "SubFormatGuid" to "Subformat" to match its equivalent in ACPI_spec

source/include/actbl2.h
```diff
@@ -2265,6 +2265,215 @@ typedef struct nfit_device_handle
...
+} ACPI_NHLT_GENDEVICE_CONFIG;
+
+/* Values for ConfigType field above */
+
+#define ACPI_NHLT_CONFIGTYPE_GENERIC         0
+#define ACPI_NHLT_CONFIGTYPE_MICARRAY        1
+
+typedef struct acpi_nhlt_micdevice_config
+{
+    UINT8                          VirtualSlot;
```
andy-shev, line 2359: So, this is basically an embedded `struct acpi_nhlt_gendevice_config`, correct? I dunno if we can use it inline (and it will need packed attribute I believe) per ACPICA code style, but at least a comment maybe added as you done elsewhere like

    /* ACPI_NHLT_GENDEVICE_CONFIG */
    ... VirtualSlot;
    ...

source/include/actbl2.h
```diff
@@ -2265,6 +2265,215 @@ typedef struct nfit_device_handle
...
+#define ACPI_NHLT_MICLOCATION_BOTTOM          1
+#define ACPI_NHLT_MICLOCATION_LEFT            2
+#define ACPI_NHLT_MICLOCATION_RIGHT           3
+#define ACPI_NHLT_MICLOCATION_FRONT           4
+#define ACPI_NHLT_MICLOCATION_REAR            5
+
+typedef struct acpi_nhlt_vendor_micdevice_config
+{
+    UINT8                          VirtualSlot;
+    UINT8                          ConfigType;
```
andy-shev, line 2415: As per above.

    /* ACPI_NHLT_MICDEVICE_CONFIG */
    ...

source/include/actbl2.h
```diff
@@ -2265,6 +2265,215 @@ typedef struct nfit_device_handle
...
+    UINT8                          ArrayType;
+    UINT8                          MicsCount;
+    ACPI_NHLT_VENDOR_MIC_CONFIG    Mics[];
+
+} ACPI_NHLT_VENDOR_MICDEVICE_CONFIG;
+
+typedef union acpi_nhlt_device_config
```
andy-shev, line 2422: Yep, cool!

andy-shev (changes requested):
Way better than previous attempt, thank you!
Few minor things I have commented on, though.

source/include/actbl2.h
crojewsk-intel: Yes, you're correct.

My vote goes for the comment option. I believe that these structs are so thin that adding a "base" element would complicate accessing their fields unnecessarily.

For example, obtaining `ConfigType` for instance of `struct acpi_nhlt_vendor_micdevice_config` would take the form of: `vendor_mic.Base.Base.ConfigType`. As there is a friendly union already (below), I don't think inlining is a bad take.

crojewsk-intel: Hmm.. now when I think about it, reader after seeing the comment, may think there are two sets of fields rather than one. Comments tailing description of several structs above make it clear that given descriptor is followed by other data types. Here it's not the case. Nothing "hidden" is following or prepending the device_configs.

I see that there are other examples of such constructs in `actbl2.h`: `struct acpi_ivrs_hardware_10` and `struct acpi_ivrs_hardware_11`.

andy-shev: May be elaborated comment? Oh, well, I leave it to ACPICA maintainers and you. The whole thing looks already good enough to me!

sacdintel:
Thanks for the review @andy-shev and I agree with you that everything looks nice in this PR! @crojewsk-intel I keep getting this one type of error repeated a ton of times under Windows using MSVC compiler:
https://learn.microsoft.com/en-us/cpp/error-messages/compiler-errors-1/compiler-error-c2233?view=msvc-170

So I have no clue how and why it is supported in GCC (although I have seen Linux folks use this before), but having a flexible array of any object with a flexible array element itself is theoretically unbounded, memory unsafe and not guaranteed that we can return from. Microsoft disallows that, meaning the said innermost element needs to be bounded (even if it is an arbitrarily large number that can hold all the flexible entities). I will suggest a fix in the code review, but for now I can go ahead with merging this PR and adding a commit myself with that fix applied and that number in this fix can be then changed later on if it is too small of a number that I recommended. Thanks a lot for your patience, I have been just back this week from a 6 week long leave for caring for my new baby!

source/include/actbl2.h
```diff
@@ -2265,6 +2265,215 @@ typedef struct nfit_device_handle
...
+
+/* Values for Direction field above */
+
+#define ACPI_NHLT_DIR_RENDER                 0
+#define ACPI_NHLT_DIR_CAPTURE                1
+
+typedef struct acpi_nhlt_config
+{
+    UINT32                         CapabilitiesSize;
+    UINT8                          Capabilities[];
```
sacdintel, line 2341: We need a bound on Capabilities[] flexible array since the struct object is in itself made into another flexible array from inside another struct! So I recommend something like Capabilities[1024] here which should ensure that the 1024 is logically a big enough number for any practical applications/purposes. Thanks!

andy-shev: This feels simply wrong. I believe you have to have it as flexible array. Can't you use `ACPI_FLEX_ARRAY()`?

sacdintel: It would look more organized for sure, but then we are opening a can of worms since there are dozens of such usages all over in ACPICA and maybe one day we can clean them up! AFAIK ACPI_FLEX_ARRAY() was added fairly recently because GCC 12(+) or Clang would not allow (without a warning and thus breaking our build) a flexible array as the sole element of a struct/union that is nested inside another struct/union. As a side effect, we have to disable a specific MSVC warning C4201 on nameless structs/unions usage too. This merged PR added the ACPI_FLEX_ARRAY() feature:
https://github.com/acpica/acpica/pull/856

However, in this scenario, even using ACPI_FLEX_ARRAY() does not work because under the hood it is still expanded to [] or [0] depending on the environment and MSVC simply does not allow any type of usage of nested flexible arrays which results in that error C2233. I recommend Cezary updates the bound on Capabilities[] both in code and documentation and maybe uses a macro instead of a magic number like I used to get past the hurdle, but I am still afraid he won't be able to use any kind of flexible and nested arrays! It feels like a hard job sometimes for us balancing both MSVC and GCC/LLVM/Clang compilers and keeping all the environments stable and warning/error-free!

crojewsk-intel:
I agree with Andy, this does not look good. The problem is a known one and on Windows side the driver code is using macros to calculate the size rather than directly calling `sizeof()` and leaves the array as `Capabilities[1]`. `Capabilities[1024]` is ambiguous from sound-driver point of view, regardless if we're talking about Windows or Linux equivalent. Payloads targeting PDM endpoints often span more than 4KB.

As this table is not an upcoming feature but rather something that's with us for several years, my wiggle room is very limited.

source/include/actbl2.h
crojewsk-intel: I didn't drink my coffee and replied in general thread rather than this conversion. Sorry for the mess.

andy-shev: > It would look more organized for sure, but then we are opening a can of worms since there are dozens of such usages all over in ACPICA and maybe one day we can clean them up! AFAIK ACPI_FLEX_ARRAY() was added fairly recently because GCC 12(+) or Clang would not allow (without a warning and thus breaking our build) a flexible array as the sole element of a struct/union that is nested inside another struct/union. As a side effect, we have to disable a specific MSVC warning C4201 on nameless structs/unions usage too. This merged PR added the ACPI_FLEX_ARRAY() feature: #856
> 
> However, in this scenario, even using ACPI_FLEX_ARRAY() does not work because under the hood it is still expanded to [] or [0] depending on the environment and MSVC simply does not allow any type of usage of nested flexible arrays which results in that error C2233. I recommend Cezary updates the bound on Capabilities[] both in code and documentation and maybe uses a macro instead of a magic number like I used to get past the hurdle, but I am still afraid he won't be able to use any kind of flexible and nested arrays! It feels like a hard job sometimes for us balancing both MSVC and GCC/LLVM/Clang compilers and keeping all the environments stable and warning/error-free!

I would rather say that MSVC is simply broken. And maybe exclude compilation of NHLT from the ACPICA on Windows (bugs or misfeatures in the compilers are not so rare).

sacdintel: I might have been able to fix it using Capabilities[1] on MSVC after updating the toolset on MSVC which includes the C/C++ compiler. The most recent commit is addressed to that effect and I am hoping that with GCC version >12 it won't throw a warning thus breaking the build since I found similar usages (1 in the same file) and some in other parts of ACPICA (I can't test GCC>12 on my old NUC with Ubuntu).

PR #918, opened 2023-12-12
Upstream changes in Haiku support

pulkomandy:
These fixes were done in Haiku's copy of the ACPICA sourcecode, but not sent to ACPICA yet.

It would ease our maintenance process if they could be upstreamed. Thanks!

The first two commits are changes only in Haiku specific code, and should cause no problems.

The third one was already discussed for upstreaming, but was never merged. Let me know if I implemented correctly what was described.

### DIFF
--- a/source/components/utilities/utdelete.c
+++ b/source/components/utilities/utdelete.c
@@ -293,7 +293,7 @@ AcpiUtDeleteInternalObj (
             /* Global Lock has extra semaphore */
 
             (void) AcpiOsDeleteSemaphore (AcpiGbl_GlobalLockSemaphore);
-            AcpiGbl_GlobalLockSemaphore = NULL;
+            AcpiGbl_GlobalLockSemaphore = ACPI_SEMAPHORE_NULL;
 
             AcpiOsDeleteMutex (Object->Mutex.OsMutex);
             AcpiGbl_GlobalLockMutex = NULL;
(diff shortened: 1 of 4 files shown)


sacdintel:
Could you please provide a brief description of what is designed to achieve, what is the motivation behind that or some other things I need to know to review and merge it? For example, one big hint is that it adds semaphores in Haiku (I guess?) which didn't exist before from the looks of it. I am not aware of the 2 commits that are apparently stable under the local ACPICA codebase of Haiku and neither am I aware of what was discussed for the 3rd commit. Thanks a lot!

pulkomandy:
Hello,

Semaphores in Haiku are already implemented using our native semaphore code (using sem_id as an identifier for semaphores). The problem is that semaphore IDs are integers starting at 0 (they are not pointers). And, since ACPI is initialized pretty early in the boot process, it is in fact likely to get the first semaphore used, that will have id 0.

The fix is to define ACPI_SEMAPHORE_NULL to be NULL on all other platforms, and -1 on Haiku, so that the code can make a difference between an uninitialized semaphore and one that happens to have the ID 0.

The fix to ACPI_FLUSH_CPU_CACHE is simply synchronizing Haiku with similar changes made on other platforms. ACPI_CPU_FLUSH_CACHE had been missed because until now we didn't have any non-x86 ports. We have Haiku on RISC-V now and I think ARM or ARM64 is also not very far from being completed. Now the implementation is similar to acwin.h.

For AcpiOsAcquireGlobalLock, the missing volatile resulted in the compiler optimizing the function implementation too much, and removing the actual check of the lock value. We found this after a compiler update, and we had to change the function prototype to fix it. The implementation is in Haiku specific files that are not upstreamed in ACPICA, so here only the prototype change is visible.

sacdintel:
Thanks for the very detailed answer, that is very helpful! LGTM, so I am merging it now! Thanks!

PR #928, opened 2024-02-19
Fix CXL 3.0 structure (RDPAS) field in the CEDT table

HojinNamm:
Fix CXL 3.0 structure (RDPAS) in the CEDT table

struct acpi_cedt_rdpas does not match with CXL r3.0 9.17.1.5 Table 9-24. 
reserved1 and length fields are already added by struct acpi_cedt_header.

### DIFF
--- a/source/include/actbl1.h
+++ b/source/include/actbl1.h
@@ -834,8 +834,6 @@ struct acpi_cedt_cxims {
 
 struct acpi_cedt_rdpas {
     ACPI_CEDT_HEADER        Header;
-    UINT8                   Reserved1;
-    UINT16                  Length;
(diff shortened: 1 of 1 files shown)


sacdintel:
CEDT_CXL_RDPAS
[CXL-3.1-Specification.pdf](https://github.com/acpica/acpica/files/14516699/CXL-3.1-Specification.pdf)
Thanks for catching this! Attaching the PDF and screenshot if anyone needs to verify this!

sacdintel:
[CXL-3.1-Specification.pdf](https://github.com/acpica/acpica/files/14516704/CXL-3.1-Specification.pdf)

PR #933, opened 2024-03-12
Detect FACS even for reduced hardware platforms

dwmw2:
The FACS is optional even on hardware reduced platforms, and may exist for the purpose of communicating the hardware_signature field to provke a clean reboot instead of a resume from hibernation.

### DIFF
--- a/source/components/tables/tbfadt.c
+++ b/source/components/tables/tbfadt.c
@@ -489,24 +489,19 @@ AcpiTbParseFadt (
         ACPI_TABLE_ORIGIN_INTERNAL_PHYSICAL, NULL, FALSE, TRUE,
         &AcpiGbl_DsdtIndex);
 
-    /* If Hardware Reduced flag is set, there is no FACS */
-
-    if (!AcpiGbl_ReducedHardware)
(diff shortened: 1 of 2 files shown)


metin-arm:
Nit: `s/provke/provoke/` in commit message? :)

sacdintel:
Hi David,

I will need to check and confirm if this PR can possibly go against the ACPI (6.5) Specification or not since usually there is a good reason for prior design decisions made by previous (or even current) ACPICA maintainers! Also please confirm in the meantime that whatever you tried works fine on reduced hardware or if it still is under test, thanks a lot!

dwmw2:
Hi Saket,

Yes, it's tested on Arm64 in conjunction with https://lore.kernel.org/lkml/20240312134148.727454-2-dwmw2@infradead.org/#t (and https://github.com/tianocore/edk2/commit/ba96acd963e794e86633ffd5e7f96cbc03db673c to make UEFI actually expose the FACS).

rafaeljw (approved):
Makes total sense to me.

sacdintel:
Thanks a lot for this contribution, LGTM now!

PR #934, opened 2024-03-16
events/evxfgpe: correctly clear GPE_AUTO_ENABLED in AcpiSetupGpeForWake

d-tatianin:
Because of a typo we would inverse the mask twice, thus producing a bogus result.


sacdintel:
Looks good to me, thanks! One of your first changes/suggestions that is being officially added so far!

PR #938, opened 2024-03-21
events/evgpeinit: don't forget to increment registered GPE count

d-tatianin:
This was used to log the number of newly discovered GPEs post table load in AcpiEvUpdateGpes, but we never incremented the number inside AcpiEvMatchGpeMethod so that was never logged.


sacdintel:
Hi Daniil,

Thanks for this PR, merging it now! Your attention to detail is helping us out a lot especially in edge and corner cases which previously have gone undetected/unreported! Please continue the great work!

- Saket

PR #941, opened 2024-03-22
Implement the DWordPCC Resource Descriptor Macro

jmarinho:
The DWordPcc (alongside WordPcc and QWordPcc) were accepted to ACPI inclusion via a code-first process.

This patch implements support in iASL for DWordPcc.

ASWG ticket: https://mantis.uefi.org/mantis/view.php?id=2429
Bugzilla ticket: https://bugzilla.tianocore.org/show_bug.cgi?id=4594

Change-Id: I06d61644b428bed2802a76abff59ceb1c79930da

### DIFF
--- a/source/compiler/aslmap.c
+++ b/source/compiler/aslmap.c
@@ -319,6 +319,7 @@ const ASL_MAPPING_ENTRY     AslKeywordMapping [] =
 /* DWORDCONST */                OP_TABLE_ENTRY (AML_RAW_DATA_DWORD,         0,                              0,                  ACPI_BTYPE_INTEGER),
 /* DWORDIO */                   OP_TABLE_ENTRY (AML_DEFAULT_ARG_OP,         0,                              0,                  0),
 /* DWORDMEMORY */               OP_TABLE_ENTRY (AML_DEFAULT_ARG_OP,         0,                              0,                  0),
+/* DWORDPCC */                  OP_TABLE_ENTRY (AML_DEFAULT_ARG_OP,         0,                              0,                  0),
 /* DWORDSPACE */                OP_TABLE_ENTRY (AML_BYTE_OP,                0,                              0,                  0),
 /* EISAID */                    OP_TABLE_ENTRY (AML_DWORD_OP,               0,                              0,                  ACPI_BTYPE_INTEGER),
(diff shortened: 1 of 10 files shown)


source/compiler/aslmap.c
```diff
@@ -339,6 +339,7 @@ const ASL_MAPPING_ENTRY     AslKeywordMapping [] =
 /* FINDSETRIGHTBIT */           OP_TABLE_ENTRY (AML_FIND_SET_RIGHT_BIT_OP,  0,                              0,                  ACPI_BTYPE_INTEGER),
 /* FIXEDDMA */                  OP_TABLE_ENTRY (AML_DEFAULT_ARG_OP,         0,                              0,                  0),
 /* FIXEDIO */                   OP_TABLE_ENTRY (AML_DEFAULT_ARG_OP,         0,                              0,                  0),
+/* DWORDPCC */                  OP_TABLE_ENTRY (AML_DEFAULT_ARG_OP,         0,                              0,                  0),
```
rgough-ampere, line 342: It looks like entries in this file are alphabetical- so DWORDPCC would go between DWORDMEMORY and DWORDSPACE.
This seems to be true for other places where code is inserted in this PR also.

source/compiler/aslrestype2d.c
```diff
@@ -158,11 +158,168 @@
...
+
+/*******************************************************************************
+ *
+ * FUNCTION:    RsDoDwordPccDescriptor
+ *
+ * PARAMETERS:  Info                - Parse Op and resource template offset
+ *
+ * RETURN:      Completed resource node
+ *
+ * DESCRIPTION: Construct a long "DwordIO" descriptor
```
rgough-ampere, line 176: I think this should say DwordPcc instead of DwordIO

source/compiler/aslmap.c
jmarinho: Fixed

source/compiler/aslrestype2d.c
jmarinho: Done

sacdintel: Thanks for the review Rob!

sacdintel:
OK now it looks much better @jmarinho than when I peeked at this PR on Friday. You seem to have fixed a couple of ill-indentations which were bothering me when it comes to coding style whereas @rgough-ampere provided you with some good content feedback! Thanks a lot, I will merge this PR now!

PR #942, opened 2024-03-28
Allow PCC Data Type in MCTP resource.

admiyo:
explicitly allow QWord address space description type to be 0xA to indicate it refers to a Platform
Communication channel
IAW  ACPI_ECR_PCC_DESCRIPTORS_CF_V2

https://bugzilla.tianocore.org/show_bug.cgi?id=4594

An entity in a DSDT or in SSDTs that requires
a platform communication channel table (PCCT) entry to communicate with a remote service will have a single value that is the index of the entry in the PCCT.
This data type with have a ResourceType value of 0xA.

This value indicates that the type shares the same footprint as a DWordSpace section.

### DIFF
--- a/source/components/resources/rsaddr.c
+++ b/source/components/resources/rsaddr.c
@@ -441,7 +441,8 @@ AcpiRsGetAddressCommon (
     /* Validate the Resource Type */
 
     if ((Address.ResourceType > 2) &&
-        (Address.ResourceType < 0xC0))
+        (Address.ResourceType < 0xC0) &&
+        (Address.ResourceType != 0x0A))
     {
         return (FALSE);
     }


source/components/resources/rsaddr.c
```diff
@@ -441,7 +441,8 @@ AcpiRsGetAddressCommon (
     /* Validate the Resource Type */
 
     if ((Address.ResourceType > 2) &&
-        (Address.ResourceType < 0xC0))
+        (Address.ResourceType < 0xC0) &&
+        (address.resource_type != 0xA))
```
sacdintel, line 445: You might have left this line (445) from your Linux implementation! This needs to be corrected to (Address.ResourceType != 0xA) ) just like the couple of lines above that inside the same 'if' statement predicate. The hex value might look even better if it is 0x0A to maintain the consistency in the same usage. Please fix this and then it will be all good I think, thanks.

admiyo: Thank you.  You are correct, the change was from Linux, and I missed the capitalization and naming convention differences.  I believe I have them correct in the updated version.

sacdintel:
LGTM now, thanks for fixing this PR! Merging it now!

PR #943, opened 2024-04-03
Implement ACPI_WARNING_ONCE and ACPI_ERROR_ONCE

anarsoul:
Implement ACPI_WARNING_ONCE and ACPI_ERROR_ONCE and use them to stop nagging the user about every Stall() call that violates the spec

### DIFF
--- a/source/components/executer/exsystem.c
+++ b/source/components/executer/exsystem.c
@@ -296,7 +296,7 @@ AcpiExSystemDoStall (
          * (ACPI specifies 100 usec as max, but this gives some slack in
          * order to support existing BIOSs)
          */
-        ACPI_ERROR ((AE_INFO,
+        ACPI_ERROR_ONCE ((AE_INFO,
             "Time parameter is too large (%u)", HowLongUs));
         Status = AE_AML_OPERAND_VALUE;
     }
@@ -304,7 +304,7 @@ AcpiExSystemDoStall (
     {
         if (HowLongUs > 100)
         {
-            ACPI_WARNING ((AE_INFO,
+            ACPI_WARNING_ONCE ((AE_INFO,
                 "Time parameter %u us > 100 us violating ACPI spec, please fix the firmware.", HowLongUs));
         }
         AcpiOsStall (HowLongUs);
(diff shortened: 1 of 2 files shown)


anarsoul:
@sacdintel here is the change that we discussed yesterday in https://github.com/acpica/acpica/commit/05ba545ce7859392250b18c10081db25c90ed8d7

sacdintel:
Hi Vasily,

Thanks for opening this PR very promptly, I really appreciate it! It's a clever macro usage to implement the 'once' functionality that can be extended to warnings and errors. I just wanted to ask if you think it would make any sense or provide us any meaningful insight (as to the severity of the error/warning msg) to keep a count on how many times that particular error/warning is repeated? In your case it is in stall() but can be extended similarly to get rid of annoying messages repeated several times in other places, perhaps. Let me know what you think! Looks like a good PR regardless, will test it out and provide any feedback by next week and also hopefully merge it soon! I would just like to keep it open for a few more days to see if someone else has anything more to add or any questions, thanks!

anarsoul:
> I just wanted to ask if you think it would make any sense or provide us any meaningful insight (as to the severity of the error/warning msg) to keep a count on how many times that particular error/warning is repeated

I think it's orthogonal to the issue that this PR resolves. You are talking about rate limiting certain messages. Rate limiting can be useful in some cases, but, in my opinion, not here, since it doesn't increase or decrease severity of the issue. I.e. if the firmware does out-of-spec Stall() call, does it really matter if it calls it once or 100 times?

In this particular case interpreter warns the user about buggy firmware. I believe it is enough to warn the user once per boot, so they are aware. It doesn't make any sense to print the warning every time when out-of-spec Stall() is called

> In your case it is in stall() but can be extended similarly to get rid of annoying messages repeated several times in other places, perhaps. Let me know what you think

Sorry, but I am not really familiar with the codebase, so I am not sure if there are other places that can benefit from it.

> Looks like a good PR regardless, will test it out and provide any feedback by next week and also hopefully merge it soon! I would just like to keep it open for a few more days to see if someone else has anything more to add or any questions, thanks!

Thanks a lot for a prompt reply! No need to hurry, feel free to keep it open as long as you need.

sacdintel:
Thanks for the analysis Vasily! We won't bother with the message frequency then, unless someone requests for that piece of information in the future, for whatever reason they deem it worthy.

sacdintel:
LGTM so I am merging this PR. Thanks for the very prompt contribution!

PR #944, opened 2024-04-03
Fix memory leak if AcpiPsGetNextNamepath() fails

Wer-Wolf:
If AcpiPsGetNextNamepath() fails, the previously allocated ACPI_PARSE_OBJECT needs to be freed before returning the status code.

The issue was first being reported on the Linux ACPI mailing list:

https://lore.kernel.org/linux-acpi/56f94776-484f-48c0-8855-dba8e6a7793b@yandex.ru/T/

### DIFF
--- a/source/components/parser/psargs.c
+++ b/source/components/parser/psargs.c
@@ -1016,6 +1016,11 @@ AcpiPsGetNextArg (
 
             Status = AcpiPsGetNextNamepath (WalkState, ParserState,
                 Arg, ACPI_NOT_METHOD_CALL);
+            if (ACPI_FAILURE(Status))
+            {
+                AcpiPsFreeOp (Arg);
+                return_ACPI_STATUS (Status);
+            }
         }
         else
         {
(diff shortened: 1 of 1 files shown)


Wer-Wolf:
As a side note: another possible memory leak might happen in psargs.c in line 946 (if AcpiPsGetNextField() fails).

Wer-Wolf:
I just fixed a issue with the coding style.

sacdintel:
> As a side note: another possible memory leak might happen in psargs.c in line 946 (if AcpiPsGetNextField() fails).
 
Do we need to address that too and if yes, can you do so with perhaps another commit to this PR? Thank you very much Armin for contributing a lot to our project recently and I hope to merge this PR by next week after testing it out! Please feel free to get in touch via email!

Wer-Wolf:
I already have a prototype, however i do not know how to properly undo `AcpiPsAppendArg` in case of an error.

Wer-Wolf:
Ok, it seems that i found the solution. However i would prefer to submit this as a separate PR.

sacdintel:
Not a problem Armin, a separate PR would be fine! Thanks a lot!

d-tatianin:
Would it perhaps be better to add a `goto` here instead of duplicating this cleanup code multiple times? This seems more error prone than just doing it in one place

Wer-Wolf:
Since AcpiPsFreeOp() does not have to be called in all cases, i would not use a `goto` here.

sacdintel:
A 'goto' statement can be really useful in certain circumstances but it is not very favorable since it is an 'unconditional jump' and needs a few more CPU cycles than any other alternative. In this scenario, only for 2 usages, I prefer not to use it and also every kind of 'failure in ACPI' needs to be dealt with by its own specific 'fix and cleanup & exit'. This PR looks good to me, so I will go ahead and merge it now! Thanks a lot Armin!

PR #946, opened 2024-04-09
check null return of ACPI_ALLOCATE_ZEROED in AcpiDbConvertToPackage

TheIntouchables:
ACPI_ALLOCATE_ZEROED may fails, Elements might be null and will cause null pointer dereference later.

### DIFF
--- a/source/components/debugger/dbconvert.c
+++ b/source/components/debugger/dbconvert.c
@@ -354,6 +354,8 @@ AcpiDbConvertToPackage (
 
     Elements = ACPI_ALLOCATE_ZEROED (
         DB_DEFAULT_PKG_ELEMENTS * sizeof (ACPI_OBJECT));
+    if (!Elements)
+        return (AE_NO_MEMORY);
 
     This = String;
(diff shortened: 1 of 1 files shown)


source/components/debugger/dbconvert.c
```diff
@@ -354,6 +354,8 @@ AcpiDbConvertToPackage (
 
     Elements = ACPI_ALLOCATE_ZEROED (
         DB_DEFAULT_PKG_ELEMENTS * sizeof (ACPI_OBJECT));
+	if (!Elements)
```
d-tatianin, line 357: i think the indentation is a bit too deep 😅

TheIntouchables: Sorry about that, but I didn't make such an ugly indentation when modifying the code. I believe this is caused by a web page rendering issue.

d-tatianin: LGTM

sacdintel:
Thanks @TheIntouchables for this PR and also thanks to @d-tatianin for catching the tab usage instead of spaces, we do need to clean up some tabs and replace them with spaces using acpisrc hopefully before the next release! LGTM now so I am merging it!

PR #948, opened 2024-04-17
Do not strip debug info when debugging is enabled

Wer-Wolf:
Then DEBUG is set to TRUE, the resulting binaries contain debugging information. Since those binaries are however always stripped during installation, this debug information is effectively unusable.

Fix this by not stripping the resulting binaries during installation in such a case.

Fixes: a7a2de9b80be ("Makefile: add debug flag")

### DIFF
--- a/generate/unix/Makefile.config
+++ b/generate/unix/Makefile.config
@@ -80,7 +80,14 @@ INSTALL  =   cp
 INSTALLFLAGS ?= -f
 else
 INSTALL =    install
+
+# Do not strip debug info when in debug mode
+ifeq ($(DEBUG),TRUE)
+INSTALLFLAGS ?= -m 555
+else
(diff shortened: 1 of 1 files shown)


sacdintel:
Hi Armin,

Is this debug information reflected in the ADBG object (for Linux) which is the ACPI Debug and can be logged in dmsg for example? I will need more time and maybe some community feedback as well on whether anything sensitive/confidential isn't leaked through the debug data. Thanks a lot for all your contributions!

Wer-Wolf:
I dont think so, the debug info in question in generated by the compiler and AFAIK should not end up in any AML code.
It also only affect the installed utilities (iasl, acpidump, acpiexec, ...), so no changes to the kernel subsystem.

sacdintel:
LGTM! Thanks!

PR #951, opened 2024-05-07
Avoid warning for Dump Functions

admlck:
Only include the functions AcpiRsDumpResourceList and AcpiRsDumpIrqList if ACPI_DEBUGGER is defined, as specified in the header. This avoid the compiler warning by adding the ifdef for ACPI_DEBUGGER.

The header: https://github.com/acpica/acpica/blob/master/source/include/acresrc.h#L470


sacdintel:
LGTM, thanks for your contribution!

PR #952, opened 2024-05-10
Get rid of a useless UseCount field in ACPI_MUTEX_INFO

d-tatianin:
This could only ever have a value of 1 or 0 since AcpiOSAcquireMutex is not recursive, so having this counter just wastes extra space.

See check above (triggered if ACPI_MUTEX_DEBUG is enabled):
```c
if (AcpiGbl_MutexInfo[i].ThreadId == ThisThreadId)
{
    if (i == MutexId)
    {
        ACPI_ERROR ((AE_INFO,
            "Mutex [%s] already acquired by this thread [%u]",
            AcpiUtGetMutexName (MutexId),
            (UINT32) ThisThreadId));

        return (AE_ALREADY_ACQUIRED);
    }
```
And also a call to non-recursive `AcpiOSAcquireMutex`, which will hang forever if UseCount is 1:
```c
    Status = AcpiOsAcquireMutex (
        AcpiGbl_MutexInfo[MutexId].Mutex, ACPI_WAIT_FOREVER);
```

### DIFF
--- a/source/components/debugger/dbstats.c
+++ b/source/components/debugger/dbstats.c
@@ -585,8 +585,9 @@ AcpiDbDisplayStatistics (
         AcpiOsPrintf ("\nMutex usage:\n\n");
         for (i = 0; i < ACPI_NUM_MUTEX; i++)
         {
-            AcpiOsPrintf ("%-28s:       %7u\n",
-                AcpiUtGetMutexName (i), AcpiGbl_MutexInfo[i].UseCount);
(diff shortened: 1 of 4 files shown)


d-tatianin:
Actually never mind, I think the idea here is to just display how many times it was _used_, not its recursion depth (?)

sacdintel:
Yes, I think it displays the use and not the depth!

PR #954, opened 2024-05-22
acpica: Revert "Do not cross page boundaries when mapping operation regions."

RajuRangoju:
Revert changes made in commit 381f8561ee4e ("Do not cross page boundaries when mapping operation regions."). The original intention of that commit was to prevent memory mappings for operation regions from crossing page boundaries, as it can cause warnings in case of different page attributes.

However, it was discovered that when this condition occurs, mapping is done until the end of the boundary, but still attempts to read/write the entire length of the map, resulting in a NULL pointer deference. For instance, if there is a request to map four bytes but only one byte is mapped because it reaches the end of the current page boundary, there is still an attempt to read/write four bytes, causing a NULL pointer dereference.

Instead, map the complete length, as there is no restriction in the ACPI specification that requires it to be in the same page boundary. It is acceptable for it to be mapped in different regions.

Cc: Mario Limonciello <mario.limonciello@amd.com>
Link: https://bugzilla.kernel.org/show_bug.cgi?id=14445
Fixes: 381f8561ee4e ("Do not cross page boundaries when mapping operation regions.")
Co-developed-by: Sanath S <Sanath.S@amd.com>
Signed-off-by: Sanath S <Sanath.S@amd.com>
Signed-off-by: Raju Rangoju <Raju.Rangoju@amd.com>

### DIFF
--- a/source/components/executer/exregion.c
+++ b/source/components/executer/exregion.c
@@ -191,7 +191,6 @@ AcpiExSystemMemorySpaceHandler (
     ACPI_MEM_MAPPING        *Mm = MemInfo->CurMm;
     UINT32                  Length;
     ACPI_SIZE               MapLength;
-    ACPI_SIZE               PageBoundaryMapLength;
 #ifdef ACPI_MISALIGNMENT_NOT_SUPPORTED
     UINT32                  Remainder;
 #endif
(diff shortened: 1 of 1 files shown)


superm1:
The correct issue to link is:
 https://bugzilla.kernel.org/show_bug.cgi?id=218849

source/components/executer/exregion.c
```diff
@@ -298,28 +297,10 @@ AcpiExSystemMemorySpaceHandler (
...
-        }
-
-        if (MapLength > PageBoundaryMapLength)
-        {
-            MapLength = PageBoundaryMapLength;
-        }
+	if (MapLength > ACPI_DEFAULT_PAGE_SIZE)
+	{
+		MapLength = ACPI_DEFAULT_PAGE_SIZE;
+	}
```
superm1, line 303: The whitespace is wrong for the style in the file.

```suggestion
        if (MapLength > ACPI_DEFAULT_PAGE_SIZE)
        {
                MapLength = ACPI_DEFAULT_PAGE_SIZE;
        }
```

superm1:
@rafaeljw can you take a look at this?

sacdintel:
Hi @RajuRangoju and @superm1 

I can review it tomorrow and then merge it if it LGTM! Sorry I am still on bonding leave until almost the end of this month, but I am planning on doing the next ACPICA release before this Q2 ends! Thanks and appreciate your patience!

sacdintel:
LGTM! Thanks for being patient, it is a good PR!

PR #959, opened 2024-07-07
A few fixes for acpixtract

d-tatianin:


d-tatianin:
> Maybe it would be better to initialize `UpperSignature` with zeros instead?
> 
> This way, we are still safe even when `Signature` is smaller than 4 characters.

Currently there's only two callers, [both of which use a hardcoded 4 character signature](https://github.com/acpica/acpica/blob/a8b050df914875839b717a51e1b32d6c6b86ae96/source/tools/acpixtract/axmain.c#L319).  

All ACPI object names are 4 bytes long exactly, so providing a smaller string is a caller error IMO, and since this is an internal function I wouldn't bother with additional safety measures, unless maintainers think it's needed in this case.

Wer-Wolf:
I think zeroing would be enough.

sacdintel:
Zeroing should be enough validation, although I believe the ACPI spec forces all signatures to be exactly 4 characters long (if not, at least the ACPI spec needs to explicitly mandate that standard according to me). The only remotely possible drawback of zeroing  would be a resulting signature mismatch of a buggy signature (not 4 chars long) although no output should be generated in that case. Let me know when I can move forward with this PR, thanks a lot @d-tatianin for your contributions!

d-tatianin:
Thanks, in that case I'll add zeroing later today and then we can get this merged.

d-tatianin:
@sacdintel I think this should be good to merge now.

Before:
image

After:
image

sacdintel:
Thanks, LGTM!

paulmenzel:
This caused a regression, cf. issue https://github.com/acpica/acpica/issues/978.

paulmenzel:
Just a small request to always copy-and-paste text, for example from terminals, instead of using screenshots for usability to be able to select text, and also smaller data size.

PR #960, opened 2024-08-07
S3 wakeup support for hardware reduced model

jiaqingz-intel:
S3 wakeup requires waking vector in FACS table, which is typically set
by OS. This patchset enables FACS table support and allows setting
waking vector in "Hardware-reduced ACPI support only" builds
(CONFIG_ACPI_REDUCED_HARDWARE_ONLY=y).


jiaqingz-intel:
Maling list thread: https://lore.kernel.org/acpica-devel/20240805060812.1002004-2-jiaqing.zhao@linux.intel.com/

sacdintel:
If you can, then please update any documentation especially in the Linux kernel that would shed more light on this new way to wakeup from S3 state in the case of hardware reduced ACPI model. I am not sure if anything needs to be updated in the ACPI specification, but I can bring up that question in one of my regular meetings. Thanks a lot!

jiaqingz-intel:
ACPI specification doesn't explicitly say that FACS table is required for passing OS waking vector to firmware.

https://uefi.org/htmlspecs/ACPI_Spec_6_4_html/16_Waking_and_Sleeping/sleeping-states.html?highlight=waking%20vector#example-s3-sleeping-state-implementation
> From S3 Sleeping State, the wake event repowers the system and resets most devices (depending on the implementation). Execution control starts from the CPU’s boot vector. The platform boot firmware is required to:
> 1. Program the initial boot configuration of the CPU (such as the MSR and MTRR registers).
> 2. Initialize the cache controller to its initial boot size and configuration.
> 3. Enable the memory controller to accept memory accesses.
> 4. Jump to the waking vector.

But in real cases, having waking vector in firmware is required for S3 resume, and almost all bootloader using FACS to provide the interface. It seems having FACS table is required for implementing S3 even in hardware-reduced model, and in hw-reduced model, FACS table only serves for S3 wakeup purpose (no global lock, S4bios). Since S3 is a optional feature, I guess that's why FACS is optional in reduced model.

Probably there is nothing need to be updated, but this is something point that not so clear in specifiation.

jiaqingz-intel:
> If you can, then please update any documentation especially in the Linux kernel that would shed more light on this new way to wakeup from S3 state in the case of hardware reduced ACPI model.

I searched kernel doc, it seems not mentioning ACPI S3 flow at all. And this is not a new way, it's just the same as non-reduced model. Maybe supporting S3 state on hw-reduced platforms is so rare that everyone ignores it.

jiaqingz-intel:
Patch v2 submitted to mailing list
https://lore.kernel.org/linux-acpi/20240812054647.1194716-1-jiaqing.zhao@linux.intel.com/T/#t

sacdintel:
Thank you for the updates @jiaqingz-intel 
The ACPI Specification is very vague when it comes to practical implementations, so I wouldn't be too concerned with updating it. The kernel might also not need any updates if this is not new (with the full HW model as opposed to this case of HW-reduced) but not sure if folks need to be aware of this new capability (especially the ARM or RISC-V or maybe even the non-Linux Zephyr RTOS folks who use the ACPI HW-reduced model more).

PR #961, opened 2024-08-08
HMAT: Add extended linear address mode to MSCIS

davejiang:
Redefine the 2 reserved bytes at offset 28 of Memory Side Cache Information Structure as "Address Mode" and add defines of the new value.

    * 0 - Reserved (Unkown Address Mode)
    * 1 - Extended-linear (N direct-map aliases linearly mapped)
    * 2..65535 - Reserved (Unknown Address Mode)

Reference: Mantis ID: 2459


sacdintel:
Thanks for the PR. I see that Mantis 2459 is still in the 'assigned' stage and it does not yet show approved and/or closed. However, it was approved in the ASWG meeting dated 8/1/2024 as per my meeting minutes records. I can review it and if it LGTM, I will go ahead and merge it.

Issue #963, opened 2024-08-11
iasl: decompiling DSDT results in segfault


alexVinarskis:
Seems to be a corner case, perhaps due to buggy ACPI tables. Tables dumped with `acpidump`, and attempted to be decompiled via `iasl *aml`. All tables are decompiled just fine besides DSDT (attached [DSDT.txt](https://github.com/user-attachments/files/16576170/DSDT.txt) ) which results in segfault.

Tested on:
* x64 Windows, version `20230628`
* arm64 Windows, version `20230628`
* arm64 Linux (wsl), version `20230628`, `master`

Segfault dump (wsl):
```

Intel ACPI Component Architecture
ASL+ Optimizing Compiler/Disassembler version 20240322
Copyright (c) 2000 - 2023 Intel Corporation

File appears to be binary: found 61249 non-ASCII characters, disassembling
Binary file appears to be a valid ACPI table, disassembling
Input file /mnt/c/Users/alexv/Desktop/iasl-win-20240321/DSDT.aml, Length 0x3E413 (254995) bytes
ACPI: DSDT 0x0000000000000000 03E413 (v02 QCOMM  SDM8380  00000003 MSFT 05000000)
Pass 1 parse of [DSDT]
Pass 2 parse of [DSDT]
Segmentation fault
*** signal 11

Backtrace:
./generate/unix/bin/iasl(+0x930b0)[0xaac0b36530b0]
./generate/unix/bin/iasl(+0x7bf64)[0xaac0b363bf64]
./generate/unix/bin/iasl(+0x7c3f4)[0xaac0b363c3f4]
./generate/unix/bin/iasl(+0x81d2c)[0xaac0b3641d2c]
./generate/unix/bin/iasl(+0xa1908)[0xaac0b3661908]
./generate/unix/bin/iasl(+0xa3668)[0xaac0b3663668]
./generate/unix/bin/iasl(+0x9bb40)[0xaac0b365bb40]
./generate/unix/bin/iasl(+0x74258)[0xaac0b3634258]
./generate/unix/bin/iasl(+0x450f0)[0xaac0b36050f0]
./generate/unix/bin/iasl(+0x63870)[0xaac0b3623870]
./generate/unix/bin/iasl(+0x297f8)[0xaac0b35e97f8]
/lib/aarch64-linux-gnu/libc.so.6(+0x284c4)[0xff375d7484c4]
[...]

PR #964, opened 2024-08-11
iasl: handle empty ConnectionNode

alexVinarskis:
Solves https://github.com/acpica/acpica/issues/963

Not sure whether this is most clean/correct way to exit. Attempt to return error status or delete object description results in further failures. With just `break` decompiling finally completes:

```
Intel ACPI Component Architecture
ASL+ Optimizing Compiler/Disassembler version 20240322
Copyright (c) 2000 - 2023 Intel Corporation

File appears to be binary: found 61249 non-ASCII characters, disassembling
Binary file appears to be a valid ACPI table, disassembling
Input file /mnt/c/Users/alexv/Desktop/iasl-win-20240321/DSDT.aml, Length 0x3E413 (254995) bytes
ACPI: DSDT 0x0000000000000000 03E413 (v02 QCOMM  SDM8380  00000003 MSFT 05000000)
Pass 1 parse of [DSDT]
Pass 2 parse of [DSDT]
Parsing Deferred Opcodes (Methods/Buffers/Packages/Regions)

Parsing completed
Disassembly completed
ASL Output:    /mnt/c/Users/alexv/Desktop/iasl-win-20240321/DSDT.dsl - 2251342 bytes
```

### DIFF
--- a/source/components/executer/exprep.c
+++ b/source/components/executer/exprep.c
@@ -611,6 +611,10 @@ AcpiExPrepFieldValue (
         if (Info->ConnectionNode)
         {
             SecondDesc = Info->ConnectionNode->Object;
+            if (SecondDesc == NULL)
+            {
+                break;
+            }
             if (!(SecondDesc->Common.Flags & AOPOBJ_DATA_VALID))
             {
                 Status = AcpiDsGetBufferArguments (SecondDesc);


sacdintel:
Thanks a lot for this PR which fixes the issue you opened! Just a break statement should be enough as per your observations since it is an error anyway (with undefined handling) if the second connection node is not found or is null. LGTM!

PR #965, opened 2024-08-13
Fixes 20240322

ahs3:
Some fixes that were in my tree that touch on table dumping.  In particular for DBG2, SLIC and PHAT tables, subtable offsets were either not always calculated properly or used problematic arithmetic with unsigned variables.

Test cases for these can sometimes be very simple; for PHAT, for example, all you have to do is run "iasl -T phat" to get a segfault.

### DIFF
--- a/source/common/dmtbdump1.c
+++ b/source/common/dmtbdump1.c
@@ -1371,8 +1371,8 @@ AcpiDmDumpDbg2 (
 
         if (Subtable->OemDataOffset)
         {
-            Status = AcpiDmDumpTable (Length, Offset + Subtable->OemDataOffset,
-                Table, Subtable->OemDataLength,
+            Status = AcpiDmDumpTable (Length, Subtable->OemDataOffset,
+                Subtable, Subtable->OemDataLength,
                 AcpiDmTableInfoDbg2OemData);
             if (ACPI_FAILURE (Status))
             {
--- a/source/common/dmtbdump2.c
+++ b/source/common/dmtbdump2.c
@@ -2028,8 +2028,10 @@ AcpiDmDumpPhat (
 
             /* Get Device-Specific Data - length of which is the remaining subtable length. */
 
-            VendorLength =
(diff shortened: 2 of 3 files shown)


Wer-Wolf:
I think the underlying problem inside the PHAT table is that the offset calculation is wrong, so the string-based calculation of PathLength fails.

I opened a separate PR #966 which should fix those issues. The other patches look good though.

ahs3:
On 13 Aug 2024 13:21, Armin W. wrote:
> I think the underlying problem inside the PHAT table is that the offset calculation is wrong, so the string-based calculation of PathLength fails.
> 
> I opened a separate PR #966 which should fix those issues. The other patches look good though.

Aha, I suspected PHAT was more problematic than that.  I just don't
have a decent sample of tables, so I wasn't completely sure I hit on
the root cause.

Thanks for doing that!

> -- 
> Reply to this email directly or view it on GitHub:
> https://github.com/acpica/acpica/pull/965#issuecomment-2287058672
> You are receiving this because you authored the thread.
> 
> Message ID: ***@***.***>

-- 
Ciao,
al
----------------------------------------------------------------------
Al Stone 
E-mail: ***@***.***
----------------------------------------------------------------------

sacdintel:
@Wer-Wolf Can you please rebase your PR on top of this PR from @ahs3 ? That way I can merge them both sometime this week without seeing any merge conflicts. Thanks a lot to the both of you!

Wer-Wolf:
I do not know how to rebase PRs onto other PRs, so i would prefer to just rebase my PR when you merge this PR.

sacdintel:
@Wer-Wolf Not a problem and yes that would be a more straightforward way to work, I understand. Bringing commits from this PR to your environment and then applying your commits might need you to resolve those merge conflicts yourself and then to update your PR sounds unnecessarily complicated. Just wait until next week to rework it and let me know when PR #966 is ready for review. Thanks a lot for all your outstanding contributions.

sacdintel:
Thanks @ahs3 for this PR, LGTM!

PR #967, opened 2024-08-14
Remove extraneous spaces before \n character

ColinIanKing:
There are a handful of output strings that have a trailing space before a \n character. The spaces seem to be extraneous and can be removed.


sacdintel:
LGTM! Thanks a lot @ColinIanKing also for your recent Linux contribution based on enabling the SLAB_HWCACHE_ALIGN from ACPICA side!

PR #976, opened 2024-10-11
CDAT/DSMAS: Add shareable and read only flag bit definitions

weiny2:
Add shareable and read only flag bit definitions to decode DSMAS flags field.


sacdintel:
LGTM! I will merge it now, but please leave the link to the updated documentation for reference! Thanks!

weiny2:
The latest 1.04 spec is here:

https://uefi.org/sites/default/files/resources/Coherent%20Device%20Attribute%20Table_1.04%20published_0.pdf

sacdintel:
Note: Support for bits 4 and 5 will be done in the future when the functionalities are enabled and validated, thanks!

PR #977, opened 2024-10-17
Add EINJV2_GET_ERROR_TYPE

zaidal-ampere:
Add EINJV2_GET_ERROR_TYPE as defined in the new specs(1)(2).

    Approved ACPI spces for EINJv2:
    (1) https://bugzilla.tianocore.org/show_bug.cgi?id=4615
    (2) https://bugzilla.tianocore.org/attachment.cgi?id=1446

### DIFF
--- a/source/include/actbl1.h
+++ b/source/include/actbl1.h
@@ -1369,17 +1369,17 @@ typedef struct acpi_einj_entry
 
 enum AcpiEinjActions
 {
-    ACPI_EINJ_BEGIN_OPERATION               = 0,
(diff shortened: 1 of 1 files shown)


sacdintel:
This LGTM, I am going to merge it now, but I had a question about the design: The decimal numbers are now hex values, so is there a reason 0xA through 0xF and even 0x10 were left out (the previous version had them all continuous and only after the end was "reserved") or is it done on purpose for future use? Looks like the doc in the link was edited in place, so I wonder if Samer momentarily forgot about the existence of alphabetical hex values? Thanks a lot!

PR #990, opened 2024-12-20
Interpret DMAR changes from revisions 3.4/4.0 of the VT-d spec.

stilor:
- DRHD may describe a register set of more than 4k using a new Size field.
- New structure in DMAR: SIDP
- To support SIDP, device scope structure changed to add Flags field.


SaketADumbre:
Hi @stilor has this been tested? If so, then could you please update the DMAR template which theoretically confirms that the compilation and disassemble works correctly? I would really appreciate that! Thanks a lot!

SaketADumbre:
Alright, thanks for testing it out! The DMAR template is here https://github.com/acpica/acpica/blob/master/source/compiler/dttemplate.h
and you can use the command "iasl -T DMAR" to generate the template. It will basically update the existing template with the addition of the updated SATC and SIDP sub-structures and it just fills out mostly meaningless data except for a few fields like vendor etc. (which it fills those out with generic data like INTL showing Intel as the vendor for example) so you won't need to worry about any proprietary data from your system leaking out here! I say that templates theoretically shows the well functioning of those tables, but every once in a while we do get reports into issues due to hardware misalignments/misconfigurations! Let me know if there is anything I can do to help, thanks a lot!

stilor:
SATC was already in the template (my commit only updated the incorrect comment in the code); added an SIDP record.

SaketADumbre:
Thanks a lot @stilor for the contribution, LGTM! Have a great weekend!

PR #991, opened 2024-12-20
Drop stale comment about the header file content

sudeep-holla:
Commit a025731aec31 ("Restructure ACPI table files") restructed the header files and moved lots of tables that are in the ACPI specification to actbl2.h(e.g.: APIC/MADT, PCCT, PPTT, ..etc).

This restructure made the comment stating this header file contains tables that are not in the ACPI specification incorrect. From that commit onwards it has remained as stale. Let us get rid of it as it might be misleading.

Reported-by: Sahil Kaushal <sahil.kaushal@arm.com>


SaketADumbre:
Thanks @sudeep-holla and Sahil Kaushal for noticing this tiny detail and keeping ACPICA accurate! LGTM!

sudeep-holla:
Thanks for the quick response and getting it merged.

Issue #781, opened 2022-06-11
Fail to build


hyperswine:
Im on a macos monterey 12.4 (mac mini m1) and running `make` on `generate/unix` gives me this error:

```
ld: unaligned pointer(s) for architecture arm64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make[1]: *** [obj/acpiexec] Error 1
make: *** [acpiexec] Error 2
```

hyperswine:
The full linker message:

```
- Link obj/acpiexec
ld: warning: pointer not aligned at address 0x100075EDE (_AcpiExDumpNode + 62 from obj/exdump.o)
ld: warning: pointer not aligned at address 0x100075ECA (_AcpiExDumpNode + 42 from obj/exdump.o)
ld: warning: pointer not aligned at address 0x100075EB6 (_AcpiExDumpNode + 22 from obj/exdump.o)
ld: warning: pointer not aligned at address 0x100075F06 (_AcpiExDumpCommon + 32 from obj/exdump.o)
ld: warning: pointer not aligned at address 0x100076046 (_AcpiExDumpFieldCommon + 62 from obj/exdump.o)
ld: warning: pointer not aligned at address 0x100076032 (_AcpiExDumpFieldCommon + 42 from obj/exdump.o)
ld: warning: pointer not aligned at address 0x10007601E (_AcpiExDumpFieldCommon + 22 from obj/exdump.o)
ld: warning: pointer not aligned at address 0x10007605A (_AcpiExDumpInteger + 12 from obj/exdump.o)
ld: warning: pointer not aligned at address 0x10007606E (_AcpiExDumpString + 12 from obj/exdump.o)
ld: warning: pointer not aligned at address 0x1000760AA (_AcpiExDumpBuffer + 32 from obj/exdump.o)
ld: warning: pointer not aligned at address 0x100076096 (_AcpiExDumpBuffer + 12 from obj/exdump.o)
ld: warning: pointer not aligned at address 0x1000760E6 (_AcpiExDumpPackage + 42 from obj/exdump.o)
ld: warning: pointer not aligned at address 0x1000760D2 (_AcpiExDumpPackage + 22 from obj/exdump.o)
ld: warning: pointer not aligned at address 0x10007610E (_AcpiExDumpDevice + 22 from obj/exdump.o)
ld: warning: pointer not aligned at address 0x100076186 (_AcpiExDumpMethod + 82 from obj/exdump.o)
[...]

PR #995, opened 2025-01-07
Apply pack(1) to AML_RESOURCE

### LINKED ISSUE #446: UB Misalignment reports even when building with -DACPI_MISALIGNMENT_NOT_SUPPORTED in NetBSD
Hi
I'm trying to fix Undefined Behaviour reports by UBSan in NetBSD. I have built with the ACPI_MISALIGNMENT_NOT_SUPPORTED macro, this has taken care of most of the reports but a few still remain.

Can these be fixed?

```
UBSan: Undefined Behavior in components/events/evregion.c:719:14, load of misaligned address 0xffffffff82e27b5b for type 'const ACPI_NAME' which requires 4 byte alignment
UBSan: Undefined Behavior in components/resources/rsaddr.c:439:22, member access within misaligned address 0xffff922460d45f62 for type 'union AML_RESOURCE' which requires 4 byte alignment
UBSan: Undefined Behavior in components/resources/rsutils.c:407:37, member access within misaligned address 0xffff922460d76419 for type 'union AML_RESOURCE' which requires 4 byte alignment
UBSan: Undefined Behavior in components/resources/rsutils.c:358:25, member access within misaligned address 0xffff922460d76419 for type 'union AML_RESOURCE' which requires 4 byte alignment
UBSan: Undefined Behavior in components/resources/rsutils.c:372:31, member access within misaligned address 0xffff922460d76419 for type 'union AML_RESOURCE' which requires 4 byte alignment
UBSan: Undefined Behavior in components/resources/rsutils.c:371:41, member access within misaligned address 0xffff922460d76419 for type 'union AML_RESOURCE' which requires 4 byte alignment
```

tamird:
Please see commit message.

Closes #446.

### DIFF
--- a/source/components/resources/rsaddr.c
+++ b/source/components/resources/rsaddr.c
@@ -431,18 +431,13 @@ AcpiRsGetAddressCommon (
     ACPI_RESOURCE           *Resource,
     AML_RESOURCE            *Aml)
 {
-
-    /* Avoid undefined behavior: member access within misaligned address */
-
-    AML_RESOURCE_ADDRESS Address;
-    memcpy(&Address, Aml, sizeof(Address));
     ACPI_FUNCTION_ENTRY();
 
     /* Validate the Resource Type */
 
-    if ((Address.ResourceType > 2) &&
-        (Address.ResourceType < 0xC0) &&
-        (Address.ResourceType != 0x0A))
+    if ((Aml->Address.ResourceType > 2) &&
+        (Aml->Address.ResourceType < 0xC0) &&
+        (Aml->Address.ResourceType != 0x0A))
     {
         return (FALSE);
     }
@@ -469,7 +464,7 @@ AcpiRsGetAddressCommon (
         /* Generic resource type, just grab the TypeSpecific byte */
 
         Resource->Data.Address.Info.TypeSpecific =
-            Address.SpecificFlags;
+            Aml->Address.SpecificFlags;
     }
 
     return (TRUE);
@@ -496,7 +491,6 @@ AcpiRsSetAddressCommon (
     ACPI_RESOURCE           *Resource)
 {
     ACPI_FUNCTION_ENTRY ();
-
 
     /* Set the Resource Type and General Flags */
 
--- a/source/components/resources/rscalc.c
+++ b/source/components/resources/rscalc.c
@@ -741,15 +741,11 @@ AcpiRsGetListLength (
             break;
 
         case ACPI_RESOURCE_NAME_SERIAL_BUS: {
-            /* Avoid undefined behavior: member access within misaligned address */
-
-            AML_RESOURCE_COMMON_SERIALBUS CommonSerialBus;
-            memcpy(&CommonSerialBus, AmlResource, sizeof(CommonSerialBus));
 
             MinimumAmlResourceLength = AcpiGbl_ResourceAmlSerialBusSizes[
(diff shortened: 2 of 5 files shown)


tamird:
Updated with the superior approach from #446.

SaketADumbre:
@tamird everything looks good to me except the usage of the pack() pragma on ARM or other RISC based architectures, which inherently do not support any packed attributes. Please take a look at this issue that I just reopened. Thanks!

https://github.com/acpica/acpica/issues/781

tamird:
Interesting, thanks for the cross-reference.

What should we do here? Fuchsia doesn't use ACPICA on anything other than x86, so I don't have a convenient platform on which to test sanitizer cleanliness on those architectures.

I suspect the situation is much on those architectures (because all the other types will start having alignment issues, not just the union). I think we should land this and tackle #781 separately. What do you think?

SaketADumbre:
That might be a reasonable plan, but I am gonna sit on this to think and strategize how to tackle all these connected issues. It is really one of the few places where packed attributes need to be used and I need to check in with some ARM64 maintainers how they handle misalignments (or rather guarantee contiguous memory alignment) in such cases. This is very interesting so far to balance out! Let me know if you have further suggestions and/or questions, thanks a lot! I really appreciate all your contributions!

tamird:
Sure thing. I do think we should land this for now as there is a new UBSAN check for misaligned pointers passed to memcpy, and this fixes that in the "proper" way.

I played with removing `#pragma pack` and there's a considerable number of problems; for example macros like `ACPI_MOVE_32_TO_32` cast both operands to `UINT32*` - sometimes one of the operands is a `char*` so it of course doesn't meet the alignment requirements of `UINT32`. So I think moving away from `#pragma pack` is going to take considerable work (and we should probably set up some CI, including on non-x86 architectures), and shouldn't block this.

Let me know if this makes sense to you.

SaketADumbre:
Hi @tamird 

It does make sense to me for now and I agree with you that in this case having packed attributes is the better option by a mile rathe than taking a different, more complex route in fixing possible misalignments! However, while I am still officially on vacation, I can take this opportunity to have a discussion with some ARM representatives regarding the lack of packed attribute support and how we can best overcome this hurdle together! I might just hold off merging this PR for the next couple of weeks until we have a plan on hand as adding more packed attributes might worsen problems on ARM and possibly other RISC based architectures. Right now, Apple's microarchitecture and possibly in the future Qualcomm's microarchitecture might beg to differ with this PR and the incompatibility gap would grow too big for us to support them all together! There has to be a simpler solution somewhere than having to suppress this warning type for each host/OS type separately in our build/make scripts when it is an architectural clash at the core of the problem!

tamird:
The simplest solution lies in plain sight: use proper alignment :).

Enjoy your vacation.

tamird:
@SaketADumbre have you had any updates here?

SaketADumbre:
Hi @tamird , I'll do something about it next week when I'm back, in the meantime I'm discussing it with an ARM person about it!

SaketADumbre:
@sudeep-holla can you please provide a perspective from the ARM ecosystem regarding the usage of packed attributes? Thanks a lot!

tamird:
@SaketADumbre ping!

sudeep-holla:
Hi @SaketADumbre , Sorry for lack of response. I didn't get any notification from Github. I had no issues with the change as such.

PR #998, opened 2025-01-19
exserial: don't forget to handle FFixedHW opregions for reading

d-tatianin:
The initial commit that introduced support for FFixedHW operation regions did add a special case in the AcpiExReadSerialBus If, but forgot to actually handle it inside the switch, so add the missing case to prevent reads from failing with AE_AML_INVALID_SPACE_ID.

Fixes: fad527b6e76 ("Add support for FFH Opregion special context data")

### DIFF
--- a/source/components/executer/exserial.c
+++ b/source/components/executer/exserial.c
@@ -359,6 +359,12 @@ AcpiExReadSerialBus (
         Function = ACPI_READ;
         break;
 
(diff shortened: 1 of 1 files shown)


SaketADumbre:
Thanks a lot @d-tatianin for this effective contribution. Glad you caught this! LGTM!

PR #1000, opened 2025-02-11
actbl3: add new TPM start methods

stuyod01:
Add TPM start method for Pluton and Arm FF-A as defined in the TCG ACPI specification v1.4.


SaketADumbre:
LGTM, thanks a lot @stuyod01

PR #1006, opened 2025-03-10
Introduce "nonstring" annotations

kees:
In order to distinguish character arrays from C Strings (i.e. strings with
a terminating NUL character), add support for the "nonstring" attribute
provided by GCC. (A better name might be "ACPI_NONCSTRING", but that's
the attribute name, so stick to the existing naming convention.)
    
GCC 15's -Wunterminated-string-initialization will warn about truncation
of the NUL byte for string initializers unless the destination is marked
with "nonstring". Prepare for applying this attribute to the project.

https://gcc.gnu.org/onlinedocs/gcc/Common-Variable-Attributes.html#index-nonstring-variable-attribute

### DIFF
--- a/source/include/actypes.h
+++ b/source/include/actypes.h
@@ -1578,4 +1578,8 @@ typedef enum
 #define ACPI_FLEX_ARRAY(TYPE, NAME)     TYPE NAME[0]
 #endif
 
+#ifndef ACPI_NONSTRING
+#define ACPI_NONSTRING	/* No terminating NUL character */
+#endif
+
 #endif /* __ACTYPES_H__ */
--- a/source/include/platform/acgcc.h
+++ b/source/include/platform/acgcc.h
@@ -224,4 +224,12 @@ typedef __builtin_va_list       va_list;
         }
 #endif
 
+/*
+ * Explicitly mark strings that lack a terminating NUL character so
+ * that ACPICA can be built with -Wunterminated-string-initialization.
+ */
+#if __has_attribute(__nonstring__)
+#define ACPI_NONSTRING __attribute__((__nonstring__))
+#endif
+
 #endif /* __ACGCC_H__ */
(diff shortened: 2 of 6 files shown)


ardbiesheuvel (approved):

kees:
Thanks for the review! I don't have commit access. Who should I poke about getting this landed? :)

source/include/actypes.h
```diff
@@ -1578,4 +1578,8 @@ typedef enum
 #define ACPI_FLEX_ARRAY(TYPE, NAME)     TYPE NAME[0]
 #endif
 
+#ifndef ACPI_NONSTRING
+#define ACPI_NONSTRING	/* No terminating NUL character */
```
SaketADumbre, line 1582: NUL -> NULL ?

source/include/platform/acgcc.h
```diff
@@ -224,4 +224,12 @@ typedef __builtin_va_list       va_list;
         }
 #endif
 
+/*
+ * Explicitly mark strings that lack a terminating NUL character so
```
SaketADumbre, line 228: NUL -> NULL ?

source/include/actypes.h
kees: Naming and spelling is weird, but "NULL" is a pointer, "NUL" is the "null character".

SaketADumbre: Hmmm, interesting. I mean I am not a naming or spelling wiz, but I just use the proper English word "Null" for everything, without all the technical differences. But I guess, I am okay with your technically specific usage.

ardbiesheuvel:
ASCII character `0` is called `NUL` not `NULL`, and it has nothing to do with pointers. There is really nothing weird about that.

https://en.wikipedia.org/wiki/ASCII

SaketADumbre:
Thanks a lot @kees for your contribution, LGTM! Thanks @ardbiesheuvel for the approval and comments!

PR #1013, opened 2025-04-08
Decrease `AcpiExTracePoint` verbosity

superm1:
Early in kernel boot pointers can't be used and so %p shows up incorrectly:

```
extrace-0138 ex_trace_point        : Method Begin [0x(____ptrval____):\M460] execution.
```

Later in the boot %p works, but it's not really actually useful when the pathname can resolve properly. Adjust the debug print so that if the Pathname resolves that the pointer isn't also printed:

```
extrace-0138 ex_trace_point        : Method Begin [\M460] execution.
```

### DIFF
--- a/source/components/executer/extrace.c
+++ b/source/components/executer/extrace.c
@@ -299,9 +299,9 @@ AcpiExTracePoint (
     if (Pathname)
     {
         ACPI_DEBUG_PRINT ((ACPI_DB_TRACE_POINT,
-            "%s %s [0x%p:%s] execution.\n",
+            "%s %s [%s] execution.\n",
             AcpiExGetTraceEventName (Type), Begin ? "Begin" : "End",
-            Aml, Pathname));
+            Pathname));
     }
     else
     {


SaketADumbre:
So there isn't much value in printing the pointer value if the Pathname works just fine? From the BIOS perspective and for debugging purposes, I would have thought the pointer value added some minor value at least .....

superm1:
It definitely isn't worth it in early boot (when ACPICA is put into Linux kernel).

With KASLR is it really still actually useful?

SaketADumbre:
That is a good question and I don't honestly know the answer and how useful exactly is it but if someone wants to contribute, we will listen!

SaketADumbre:
This is the conversation with GitHub Copilot when I asked about this PR change:

"How will this proposed PR change affect with KASLR enabled as the ACPICA code is native to Linux kernel?"


"The proposed change in the pull request modifies the debug trace messages in the file extrace.c within the ACPICA project. Specifically, it removes the inclusion of the Aml pointer in the debug message. Here's how this change impacts the behavior when Kernel Address Space Layout Randomization (KASLR) is enabled:

KASLR Overview:

KASLR randomizes the memory address space of the kernel, including the location of kernel code, data structures, and modules. This is done to improve security by making it harder for attackers to predict the memory layout.
Impact of the Change:

Prior to this change, the debug trace message included the raw pointer Aml as part of the output. With KASLR enabled, printing memory addresses (like Aml) in logs can unintentionally leak kernel memory locations, which undermines the security provided by KASLR.
After this change, the Aml pointer is no longer included in the trace message. This reduces the risk of leaking sensitive kernel memory addresses in debug logs, thereby improving security when KASLR is enabled.
Conclusion:

The proposed change enhances security by ensuring that kernel memory addresses are not exposed in debug logs. This is particularly important for systems with KASLR enabled, as it helps maintain the integrity of the randomization provided by KASLR.
Functionality-wise, this change does not affect the execution of the code but strictly modifies the debug output for better adherence to security best practices.
If you have further technical concerns or specific use cases in mind, feel free to ask!"

This answer is pretty good, but KASLR can also be disabled, I read some literature on KASLR and I think I agree with the security rationale at least, just not sure if the address will be any helpful during debug.

SaketADumbre:
I agree with you here, it would be a trivial revert, but until then this clearly helps the security aspect of the Linux kernel and we shouldn't really expose the addresses (even though they are randomized after each reboot) for adversaries to take advantage of in any negative manner! LGTM, thanks a lot @superm1 for your contributions!

PR #1016, opened 2025-04-14
Modify variable position

qiaozhe1:
Modify variable definition position to prevent potential undefined risks.

### DIFF
--- a/source/components/parser/psopinfo.c
+++ b/source/components/parser/psopinfo.c
@@ -181,7 +181,7 @@ AcpiPsGetOpcodeInfo (
     UINT16                  Opcode)
 {
 #ifdef ACPI_DEBUG_OUTPUT
-    const char              *OpcodeName = "Unknown AML opcode";
+    
 #endif
 
     ACPI_FUNCTION_NAME (PsGetOpcodeInfo);
@@ -207,7 +207,9 @@ AcpiPsGetOpcodeInfo (
 
 #if defined ACPI_ASL_COMPILER && defined ACPI_DEBUG_OUTPUT
 #include "asldefine.h"
-
+    
+    const char *OpcodeName = "Unknown AML opcode";
+    
     switch (Opcode)
     {
     case AML_RAW_DATA_BYTE:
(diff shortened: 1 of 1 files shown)


source/components/parser/psopinfo.c
```diff
@@ -181,7 +181,7 @@ AcpiPsGetOpcodeInfo (
     UINT16                  Opcode)
 {
 #ifdef ACPI_DEBUG_OUTPUT
```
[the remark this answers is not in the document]
qiaozhe1: Yes, thanks for your reminder. I noticed this problem too, I will delete this empty #ifdef.

SaketADumbre (changes requested):
../../../source/components/parser/psopinfo.c:208:5: error: ISO C90 forbids mixed declarations and code [-Werror=declaration-after-statement]
  208 |     const char *OpcodeName = "Unknown AML opcode";
      |     ^~~~~

SaketADumbre:
Sorry for catching this late as this is something that can easily skip my mind: compatibility with the older C standards like ISO C89 or C90 which forbid mixed declarations, so the const char *OpcodeName declaration needs to go back in its original place, but you can use the same #ifdef with && up there as it is down below. I just found this when building under Cygwin on Windows.

qiaozhe1:
> Sorry for catching this late as this is something that can easily skip my mind: compatibility with the older C standards like ISO C89 or C90 which forbid mixed declarations, so the const char *OpcodeName declaration needs to go back in its original place, but you can use the same #ifdef with && up there as it is down below. I just found this when building under Cygwin on Windows.

OK, thanks for pointing that out.

source/components/parser/psopinfo.c
```diff
@@ -180,8 +180,8 @@ const ACPI_OPCODE_INFO *
 AcpiPsGetOpcodeInfo (
     UINT16                  Opcode)
 {
-#ifdef ACPI_DEBUG_OUTPUT
-    const char              *OpcodeName = "Unknown AML opcode";
+#if defined ACPI_ASL_COMPILER && defined ACPI_DEBUG_OUTPUT
+    const char *OpcodeName = "Unknown AML opcode";
```
SaketADumbre, line 184: One last thing is that can you please leave the spacing as it was in the old line 184 (red/removed) here as well in this new line 184 (green/newly added)? This just creates some separation between the variable type and its actual name for better readability. Thanks, everything else looks good now other than this nitpick.

qiaozhe1: ok

SaketADumbre:
Thanks a lot for making all those requested changes, I really appreciate that! Now it LGTM and I can merge this PR and also thanks for contributing @qiaozhe1 !

Issue #1018, opened 2025-04-20
Heads up about GCC 15 changes made by Linus Torvalds


andy-shev:
In v6.15-rc3 a few patches were added to shut the GCC 15 up on some warnings. ACPICA is one of the sources of those warnings. You might be interested in fixing these in a portable way sooner. Here is the commit you may be interested in: https://web.git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=4b4bd8c50f4836ba7d3fcfd6c90f96d2605779fe.

With this a gentle reminder about issue #973. I will be happy to test any proposed fixes.

SaketADumbre:
Hi @andy-shev 

Thanks for opening this issue and for the pointer to the Linux commit. We have the same new type (I believe it is called ACPI_NONSTRING) in ACPICA and I can post a fix sometime this week for the 3/4 files in the commit that are part of ACPICA (not sure if .../drivers/acpi/tables.c is in ACPICA and what its corresponding filename is?). I can also propose a portable fix for #973 this week, I just tried the 2 suggestions from your end, but looks like one fails getting past MSVC and the other using ifdef debug still emits that warning and isn't really part of debug output in ACPICA, but a variable that keeps track of the number of carats in the namespace path.

andy-shev:
Hold on a bit: there is an ongoing discussion and seems the stuff should be already fixed by https://github.com/acpica/acpica/pull/1006:
https://lore.kernel.org/all/202504201840.3C1F04B09@keescook/T/#m76c3501925e2dfc802e515418b4d602aa1ff7a52

SaketADumbre:
[...]

PR #1019, opened 2025-04-22
Apply ACPI_NONSTRING

X0RW3LL:
Add ACPI_NONSTRING for destination char arrays without a terminating NUL character. This is a follow-up to 1035a3d453f7 where a few more destination arrays were missed

Closes: #1018

### DIFF
--- a/source/compiler/aslanalyze.c
+++ b/source/compiler/aslanalyze.c
@@ -575,7 +575,19 @@ ApCheckForGpeNameConflict (
     char                    Target[ACPI_NAMESEG_SIZE] ACPI_NONSTRING;
 
 
-    /* Need a null-terminated string version of NameSeg */
+    /**
+     * Need a null-terminated string version of NameSeg
+     *
+     * NOTE: during a review on Name[ACPI_NAMESEG_SIZE + 1] having an extra
+     *       byte[1], compiler testing exhibited a difference in behavior between
+     *       GCC and Clang[2] (at least; MSVC may also exhibit the same) in
+     *       how optimization is done. The extra byte is needed to ensure
+     *       the signature does not get mangled, subsequently avoiding
+     *       GpeNumber being a completely different return value from strtoul.
+     *
+     *       [1] https://github.com/acpica/acpica/pull/1019#discussion_r2058687704
+     *       [2] https://github.com/acpica/acpica/pull/1019#discussion_r2061953039
+     */
 
     ACPI_MOVE_32_TO_32 (Name, Op->Asl.NameSeg);
     Name[ACPI_NAMESEG_SIZE] = 0;
--- a/source/compiler/aslanalyze.c
+++ b/source/compiler/aslanalyze.c
@@ -571,7 +571,7 @@ ApCheckForGpeNameConflict (
 {
     ACPI_PARSE_OBJECT       *NextOp;
     UINT32                  GpeNumber;
-    char                    Name[ACPI_NAMESEG_SIZE + 1] ACPI_NONSTRING;
+    char                    Name[ACPI_NAMESEG_SIZE + 1];
     char                    Target[ACPI_NAMESEG_SIZE] ACPI_NONSTRING;
 
 
--- a/source/compiler/aslanalyze.c
+++ b/source/compiler/aslanalyze.c
@@ -571,8 +571,8 @@ ApCheckForGpeNameConflict (
 {
     ACPI_PARSE_OBJECT       *NextOp;
     UINT32                  GpeNumber;
-    char                    Name[ACPI_NAMESEG_SIZE + 1];
-    char                    Target[ACPI_NAMESEG_SIZE];
+    char                    Name[ACPI_NAMESEG_SIZE + 1] ACPI_NONSTRING;
+    char                    Target[ACPI_NAMESEG_SIZE] ACPI_NONSTRING;
 
 
     /* Need a null-terminated string version of NameSeg */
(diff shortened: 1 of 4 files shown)


source/compiler/aslanalyze.c
```diff
@@ -571,8 +571,8 @@ ApCheckForGpeNameConflict (
 {
     ACPI_PARSE_OBJECT       *NextOp;
     UINT32                  GpeNumber;
-    char                    Name[ACPI_NAMESEG_SIZE + 1];
-    char                    Target[ACPI_NAMESEG_SIZE];
+    char                    Name[ACPI_NAMESEG_SIZE + 1] ACPI_NONSTRING;
```
andy-shev, line 574: Is it really nonstring? What is that ` + 1` for?

X0RW3LL: Apologies for having taken a bit long to get back to you—I've been chasing this down, trying to understand exactly all the factors involved.

So, this function is called when the ASL analyzer performs the analysis walk during compilation (and/or possibly disassembly) checking for `_Lxx` vs `_Exx` name conflicts (same GPE number, different type). Once the preconditions are passed (i.e. name starts with `_`, followed by `L` or `E`, the 3rd/4th chars are verified as valid [hex values](https://github.com/acpica/acpica/blob/e374230f5fea5b2c92d9221ae3d5cad4d788569b/source/compiler/aslanalyze.c#L597) via a call to `strtoul`.

While `strtoul`'s argv1 (`const char *Sring`) is expected to be a NUL-terminated string (according to the String param description), I don't believe it is a strict requirement in this scenario. According to the ACPI specification version 6.5A, Sections 19.6.84[1] and 20.2.2[2], `MethodName` is evaluated as a `Namestring` data type.

Encoding for the relevant data types is as follows in AML:
```
NameString := <RootChar NamePath> | <PrefixPath NamePath>
NamePath := NameSeg | DualNamePath | MultiNamePath | NullName
NameSeg := <LeadNameChar NameChar NameChar NameChar>
```
This means `ApCheckForGpeNameConflict (ACPI_PARSE_OBJECT *Op)` will convert the endianness of \
`Op->Asl.NameSeg`, whose length can only ever be 4 bytes, storing the results in `Name`. Additionally, it can only contain the following characters/digits: `_A-Z0-9`.

My question is: do we really the extra NUL byte at this point?

I've also tested with the following changes and got both a successful compilation and DSDT analysis (disassembly and recompilation with level-triggered event control methods declared in the table)
```diff
diff --git a/source/compiler/aslanalyze.c b/source/compiler/aslanalyze.c
index 67f4ff42e1e0..d1e7adede1e9 100644
--- a/source/compiler/aslanalyze.c
+++ b/source/compiler/aslanalyze.c
@@ -571,14 +571,11 @@ ApCheckForGpeNameConflict (
 {
     ACPI_PARSE_OBJECT       *NextOp;
     UINT32                  GpeNumber;
-    char                    Name[ACPI_NAMESEG_SIZE + 1] ACPI_NONSTRING;
+    char                    Name[ACPI_NAMESEG_SIZE] ACPI_NONSTRING;
     char                    Target[ACPI_NAMESEG_SIZE] ACPI_NONSTRING;


-    /* Need a null-terminated string version of NameSeg */
-
     ACPI_MOVE_32_TO_32 (Name, Op->Asl.NameSeg);
-    Name[ACPI_NAMESEG_SIZE] = 0;
```

[1] https://uefi.org/specs/ACPI/6.5_A/19_ASL_Reference.html#method-declare-control-method
[2] https://uefi.org/specs/ACPI/6.5_A/20_AML_Specification.html#name-objects-encoding

SaketADumbre: I wonder if the strtoul() implementation varies across C compilers like GCC or Clang/LLVM or MSVC? I don't have the answer to it and MSVC is not open-source, maybe some behavioral testing might be needed to check the differences? Maybe not all of them "actually" need a NUL-terminated string as an argument?

SaketADumbre: For your reference:

https://devblogs.microsoft.com/oldnewthing/20240510-00/?p=109742#:~:text=The%20msvc%20implementation%20uses%20the%20capacity%20to,shrinks%20the%20string%20below%20the%20small%2Dstring%20threshold.

X0RW3LL: Thank you so much for the reference! I'm afraid I don't have the deep technical knowledge yet as to determine whether the same applies for C as opposed to C++. I'd love to be able to at least debug this and examine what's happening under the hood with the different compilers, but I've yet to figure out how I can get this to compile on Clang/MSVC. Latter I'd have to prep a Windows VM for, but for the former, I still have to figure out how to get it to compile. Very new to this, mind you, so it's taking me a little while to wrap my head around all the different moving parts.

X0RW3LL: As it turns out, there's definitely something different with the lack of the NUL terminator.
```c
GpeNumber = strtoul (&Name[2], NULL, 16);
```
**With** NUL, both GCC and Clang have `strtoul` return `GpeNumber` as 0x19

***Without*** NUL, GCC will still return `GpeNumber` as 0x19. Clang, however, returns it as 0x198

X0RW3LL: Thank you so much! I'll push another reverting only that change

andy-shev: I believe that ` + 1` in the code needs a comment. So next time even we will remember what's for and why NONSTRING is still applied.

X0RW3LL: To be honest, I was contemplating whether I should have commented it, but I thought the comment already on needing a null-terminated string was straight to the point enough. I'll push another with the comment, then, thank you!

SaketADumbre:
LGTM, thanks a lot once again @X0RW3LL for your contributions! I think this PR has been up for 2 weeks or so and it has been enough time for the community to provide feedback!

PR #1022, opened 2025-04-27
iASL/Templates: Use correct fd for printing info messages

X0RW3LL:
Print informational messages to stdout as opposed to stderr.

### DIFF
--- a/source/common/adisasm.c
+++ b/source/common/adisasm.c
@@ -481,12 +481,12 @@ AdDisassembleOneTable (
             "FieldName : FieldValue (in hex)\n */\n\n");
 
         AcpiDmDumpDataTable (Table);
-        fprintf (stderr, "Acpi Data Table [%4.4s] decoded\n",
+        fprintf (stdout, "Acpi Data Table [%4.4s] decoded\n",
             AcpiGbl_CDAT ? (char *) AcpiGbl_CDAT : Table->Signature);
 
         if (File)
         {
-            fprintf (stderr, "Formatted output:  %s - %u bytes\n",
+            fprintf (stdout, "Formatted output:  %s - %u bytes\n",
                 DisasmFilename, CmGetFileSize (File));
         }
 
@@ -584,16 +584,16 @@ AdDisassembleOneTable (
 
         AcpiDmDumpDataTable (Table);
 
-        fprintf (stderr, "Disassembly completed\n");
+        fprintf (stdout, "Disassembly completed\n");
         if (File)
         {
-            fprintf (stderr, "ASL Output:    %s - %u bytes\n",
+            fprintf (stdout, "ASL Output:    %s - %u bytes\n",
                 DisasmFilename, CmGetFileSize (File));
         }
 
         if (AslGbl_MapfileFlag)
         {
-            fprintf (stderr, "%14s %s - %u bytes\n",
+            fprintf (stdout, "%14s %s - %u bytes\n",
                 AslGbl_FileDescs[ASL_FILE_MAP_OUTPUT].ShortDescription,
                 AslGbl_Files[ASL_FILE_MAP_OUTPUT].Filename,
                 FlGetFileSize (ASL_FILE_MAP_OUTPUT));
@@ -630,7 +630,7 @@ AdReparseOneTable (
     ACPI_COMMENT_ADDR_NODE  *AddrListHead;
 
 
-    fprintf (stderr,
+    fprintf (stdout,
         "\nFound %u external control methods, "
         "reparsing with new information\n",
         AcpiDmGetUnresolvedExternalMethodCount ());
--- a/source/compiler/dttemplate.c
+++ b/source/compiler/dttemplate.c
@@ -255,7 +255,7 @@ DtCreateTemplates (
 
     if (AcpiGbl_Optind < 3)
     {
-        fprintf (stderr, "Creating default template: [DSDT]\n");
+        fprintf (stdout, "Creating default template: [DSDT]\n");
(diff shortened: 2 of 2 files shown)


SaketADumbre:
What is the motivation for this change? I see both pros and cons with changing stderr to stdout in the fprintf usages and I am not sure of the original intention of the author.

SaketADumbre:
Just for some references, I am pasting a conversation I had with Copilot, which, I am not sure if, is accessible for your use inside of GitHub:


What should I take proper of when changing fprintf(stderr, ...) to fprintf(stdout, ...)? The result might be used in an output file
When changing fprintf(stderr, ...) to fprintf(stdout, ...), there are several considerations to ensure that the modifications do not introduce issues, especially when the output might be redirected to a file:

Key Considerations
Separation of Error and Informational Messages:

stderr is typically used for error messages, warnings, or debugging information, as it is unbuffered and not affected by output redirection.
stdout is used for normal output, which may be redirected to a file or pipe.
Ensure that the messages being moved to stdout are not critical error messages that need immediate visibility, as they could get lost in a redirected output.
Buffering Behavior:

stdout is line-buffered or fully buffered by default, meaning output may not appear immediately, especially when redirected to a file.
stderr is unbuffered, ensuring that messages are printed immediately.
If the immediate visibility of these messages is important, consider using fflush(stdout) after each fprintf(stdout, ...).
Output Redirection Compatibility:

If the program's stdout output is redirected to a file, the informational messages will now be mixed with the main output.
If the program relies on parsing the output file, ensure that these messages do not interfere with the expected format.
Consistency of Usage:

Ensure that all related messages are consistently redirected. For example, if all diagnostic messages are being redirected to stdout, they should no longer be split between stderr and stdout.
Impact on Error Handling:

Moving error messages to stdout may impact tools or scripts that rely on stderr for detecting errors (e.g., build systems, logging tools, or monitoring scripts).
If the messages are critical errors, consider keeping them in stderr.
Documentation and User Expectations:

If users of the program expect error/logging messages to appear in stderr, document the change to avoid confusion.
Update any user guides or scripts that depend on the program's output behavior.
Testing:

Test the changes in scenarios where the program's output is redirected to a file, terminal, or pipe.
Ensure that the output format and functionality remain as intended.
Recommendations
If the messages being redirected to stdout are purely informational or debug messages, the change is less risky.
For critical error or warning messages, it is better to retain stderr to ensure they are visible and not buffered.
If mixing stdout and stderr content is unavoidable, consider providing a command-line option (e.g., --log-output=stderr|stdout) or an environment variable to allow users to choose where messages are directed.
Example of Safe Usage
To mimic stderr's unbuffered behavior when using stdout:

C
fprintf(stdout, "Message: %s\n", message);
fflush(stdout); // Ensure the output is written immediately
In summary, carefully assess the context and purpose of the messages before changing their output stream, and test the results to ensure the program behaves as expected in all scenarios.

What is the exact difference between the above 2 usages?

The exact difference between using `fprintf(stderr, ...)` and `fprintf(stdout, ...)` lies in how the output is directed and handled by the operating system:

### 1. **Output Stream Purpose**
   - **`stderr`**:
     - Used for error messages or diagnostics.
     - Not affected by `stdout` redirection (e.g., if the program's standard output is redirected to a file, `stderr` still goes to the terminal unless explicitly redirected).
     - Unbuffered by default, meaning the message appears immediately without waiting for a buffer to fill.
   - **`stdout`**:
     - Used for normal program output.
     - Affected by redirection (e.g., when the program's output is piped or redirected to a file, the messages sent to `stdout` will follow the redirection).
     - Typically buffered, meaning it waits until the buffer is full or flushed before writing output, which could lead to delays.

### 2. **Redirection Behavior**
   - **`fprintf(stderr, ...)`**: Messages will always appear on the standard error stream (usually the terminal) unless `stderr` is explicitly redirected.
   - **`fprintf(stdout, ...)`**: Messages will appear on the standard output stream, which might be redirected (e.g., to a file or pipeline).

### 3. **Buffering**
   - **`stderr`**: Does not buffer output, so messages are written immediately.
   - **`stdout`**: Buffers output (line-buffered in a terminal or fully buffered in a file), so messages might not appear immediately unless explicitly flushed (e.g., using `fflush(stdout)`).

### 4. **Use Case Implications**
   - **`stderr`**:
     - Ideal for error messages, warnings, and diagnostics because these outputs are meant to be visible immediately and are not mixed with the program's main output.
   - **`stdout`**:
     - Ideal for normal program output and informational messages that are part of the expected output. If redirected to a file, these messages will be included in the file.

### In Your Context
In the provided diff, the original code uses `fprintf(stderr, ...)`, meaning that the messages are treated as error or diagnostic output. After the change to `fprintf(stdout, ...)`, these messages will:
1. Be part of the standard output.
2. Potentially be redirected to a file or pipe along with the rest of the program's output.
3. Be subject to buffering, so the messages might not appear immediately unless flushed.

If the result is used in an output file, you should ensure:
- The messages now written to `stdout` do not interfere with the expected file format.
- You flush the output buffer (`fflush(stdout)`) if immediate visibility is required.

The exact difference is in how the messages are handled and whether they are visible immediately or mixed with redirected output.

X0RW3LL:
My best guess is that it might have been accidental. My main concern, and it's only a speculation as it stands at this point, is avoiding situations like certain outputs missing from logs (if any) in the case of output redirection, as well as things like lintian potentially mistaking stderr output for actual errors when they're really not. I'm aware the package wouldn't even build had that been the case, but I do think it's generally best to keep output directed to the fd that's relevant to its nature—i.e. info/success to stdout, and warnings/errors/failures to stderr

SaketADumbre:
Yeah, I agree and that makes sense! Thanks for the contributions, LGTM!

PR #1030, opened 2025-06-17
iASL: Fix _CRS CSI2 resource compilation and disassembly

s-ailus:
The iasl generates invalid AML _CRS CSI2 resource at byte 7, where the _PRT (bits 7--2) and _PHY bits (bits 1--0) are coalesced into lowest six bits of the byte 7, in what effectively is a bitwise or operation. This results in loss of information whenever the _PRT value is non-zero. A corresponding bug is also found in the disassembler, where the lowest 6 and 2 bits, respectively, are used to form the _PRT and _PHY values.

Fix both of these issues.

_CRS CSI2 descriptors, where the _PRT value is non-zero, generated before this patch are invalid.

### DIFF
--- a/source/compiler/aslrestype2s.c
+++ b/source/compiler/aslrestype2s.c
@@ -1469,7 +1469,7 @@ RsDoCsi2SerialBusDescriptor (
 
         case 2: /* Local Port Instance [Integer] (_PRT) */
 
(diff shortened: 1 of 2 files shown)


SaketADumbre:
Thanks for your contribution @s-ailus and good catch in spotting this one! LGTM!

s-ailus:
Thank you, Saket!

Issue #1027, opened 2025-06-07
Calling method with wrong number of arguments leads to unbootable Linux kernel


pkgw:
A recent Dell firmware update has caused various Dell laptops, and other systems, to become unbootable on several Linux distributions, including Fedora Linux. This is a particularly severe problem because Dell classifies the update as a security fix, so it can't be uninstalled.

In [Fedora Bugzilla 2355276][bz] I think that I've traced the issue down to an issue in the ACPICA implementation used in the Linux kernel.

[bz]: https://bugzilla.redhat.com/show_bug.cgi?id=2355276

Specifically, the problem seems to arise with a method in the Dell ACPI code named `\_SB.UBTC.RUCC`. The firmware updates (BIOS versions 1.12.0 and 1.13.0) add a third argument to this method, but it appears that one of the SSDT tables hasn't been updated to reflect this. Its "external" declaration of the RUCC symbol lists it as taking two arguments, and it's called with only two arguments in the AML code in that table. You can check out [the Bugzilla discussion][bz] for the details.

I'm unfamiliar with the ACPICA codebase, but as best I can tell it's not checking that the number of arguments passed to the method agrees with its actual implementation. And in the kernel, this is leading to memory corruption and a kernel panic. The following patch *against the Linux acpica tree* fixes the problem for me:

```diff
diff --git a/drivers/acpi/acpica/dsmethod.c b/drivers/acpi/acpica/dsmethod.c
index e809c2aed78a..a81de3265472 100644
--- a/drivers/acpi/acpica/dsmethod.c
+++ b/drivers/acpi/acpica/dsmethod.c
@@ -509,6 +509,17 @@ acpi_ds_call_control_method(struct acpi_thread_state *thread,
 	 */
 	this_walk_state->operands[this_walk_state->num_operands] = NULL;
 
+	if (this_walk_state->num_operands != obj_desc->method.param_count) {
+		printk(
+			"PKGW ACPI method exec `%4.4s` no=%d pc=%d\n",
+			method_node->name.ascii,
+			this_walk_state->num_operands,
+			obj_desc->method.param_count
+		);
+		status = AE_AML_UNINITIALIZED_ARG;
+		goto pop_walk_state;
+	}
+
 	/*
 	 * Allocate and initialize the evaluation information block
 	 * TBD: this is somewhat inefficient, should change interface to
@@ -539,7 +550,7 @@ acpi_ds_call_control_method(struct acpi_thread_state *thread,
[...]

PR #1031, opened 2025-06-18
dispatcher: Use AcpiDsClearOperands() in AcpiDsCallControlMethod()

jwrdegoede:
When deleting the previous walkstate operand stack AcpiDsCallControlMethod() was deleting ObjDesc->Method.ParamCount operands. But Method.ParamCount does not necessarily match ThisWalkState->NumOperands, it may be either less or more.

After correcting the for loop to check `i < ThisWalkState->NumOperands` the code is identical to AcpiDsClearOperands(), so just outright replace the code with AcpiDsClearOperands() to fix this.

### DIFF
--- a/source/components/dispatcher/dsmethod.c
+++ b/source/components/dispatcher/dsmethod.c
@@ -726,15 +726,7 @@ AcpiDsCallControlMethod (
      * Delete the operands on the previous walkstate operand stack
      * (they were copied to new objects)
      */
-    for (i = 0; i < ObjDesc->Method.ParamCount; i++)
-    {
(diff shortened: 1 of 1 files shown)


SaketADumbre:
Hi @jwrdegoede 

Thanks a lot for the timely contribution and also for testing this fix out! I added a commit since with your change, the variable i was unused and started throwing a warning, so removed it as it has no use any longer. I will close the Issue #1027 as completed after merging this PR, if GitHub doesn't do it automatically. Appreciate all your amazing work in helping address the issue at hand in a very prompt manner!

PR #1036, opened 2025-08-11
Fix largest possible resource descriptor index

dmantipov:
ACPI_RESOURCE_NAME_LARGE_MAX should be equal to the last actually used resource descriptor index (ACPI_RESOURCE_NAME_CLOCK_INPUT). Otherwise 'resource_index' in 'acpi_ut_validate_resource()' may be clamped incorrectly and resulting value may issue an out-of-bounds access for 'acpi_gbl_resource_types' array. Compile tested only.

Found by Linux Verification Center (linuxtesting.org) with SVACE. 
Fixes: 661feab5ee01 ("ACPICA: add support for ClockInput resource (v6.5)") 
Link: https://marc.info/?l=linux-acpi&m=175449676131260&w=2


SaketADumbre:
Thanks a lot for finding and fixing this unnoticed error! LGTM!

PR #1038, opened 2025-08-19
CEDT: Rename CXL Window coherency restrictions

davidlohr:
This has been renamed in more recent CXL specs (3.0+), as type3 (memory expanders) can also use HDM-DB for device coherent memory.


SaketADumbre:
Thanks a lot for this contribution and keeping our codebase up to date with the latest names! LGTM!

PR #1040, opened 2025-08-29
ACPICA: acpidump: fix return values in ap_is_valid_checksum()

kaushlen:
The function ap_is_valid_checksum() has a boolean name suggesting it should return TRUE/FALSE, but incorrectly returns AE_OK on success and has no explicit return on failure, leading to undefined behavior.

Fix by returning proper values:
- FALSE when checksum validation fails
- TRUE when checksum validation succeeds


SaketADumbre:
Thanks a lot @kaushlen for noticing this and promptly fixing it! LGTM!

PR #1044, opened 2025-09-13
ACPICA: replace ACPI_FREE with AcpiUtDeleteObjectDesc

GoodLuck612:
AcpiUtCreateInternalObject may allocate memory from a slab cache via kmem_cache_zalloc, but the code currently frees it with ACPI_FREE, which calls kfree. This mismatch prevents the object from being released properly and may lead to memory leaks or other issues.

Fix this by replacing ACPI_FREE with AcpiUtDeleteObjectDesc, which matches the allocation method used for internal objects.

### DIFF
--- a/source/components/namespace/nsxfname.c
+++ b/source/components/namespace/nsxfname.c
@@ -805,7 +805,7 @@ AcpiInstallMethod (
 ErrorExit:
 
     ACPI_FREE (AmlBuffer);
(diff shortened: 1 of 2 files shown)


SaketADumbre:
@GoodLuck612 Thanks for catching this! Very straightforward but highly effective PR! LGTM!

PR #1045, opened 2025-09-16
Add definition for ACPI C4 C-state

superm1:
Some AMD systems have support for a C4 C-state.  This isn't directly used in acpica, but the definitions are used in Linux and should be updated in acpica as well.

### DIFF
--- a/source/include/actypes.h
+++ b/source/include/actypes.h
@@ -751,8 +751,9 @@ typedef UINT64                          ACPI_INTEGER;
 #define ACPI_STATE_C1                   (UINT8) 1
 #define ACPI_STATE_C2                   (UINT8) 2
 #define ACPI_STATE_C3                   (UINT8) 3
-#define ACPI_C_STATES_MAX               ACPI_STATE_C3
-#define ACPI_C_STATE_COUNT              4
+#define ACPI_STATE_C4                   (UINT8) 4
+#define ACPI_C_STATES_MAX               ACPI_STATE_C4
+#define ACPI_C_STATE_COUNT              5
 
 /*
  * Sleep type invalid value


SaketADumbre:
Can I just ask why C4 state (not the C4 bomb LOL) is needed? Does it have any significant power savings advantage over the existing C3 state? Just asking in a general broad sense, not about anything that is AMD Intellectual Property! Thanks a lot!

superm1:
On a system that _exposes C4_ if the OS doesn't support it yes you will not be able to obtain the best power savings.
If the system only _exposes up to C3,_ then of course no implicit power savings by having a C4 state supported by the OS.

SaketADumbre:
Thanks for your response and I believe that also for enabling S0i3 you need C4. Anyway this PR LGTM, thanks a lot @superm1

SaketADumbre:
@superm1 I did find out in the ACPI Specification 6.6 that the supported C state types are still C1, C2 and C3 (say with _CST field values). The spec doesn't restrict the number of C states though, so I am just seeking clarification from you if you actually intend on defining a new C state type C4 or if it maps to HW C3 or some other mapping per the platform?

_Section 8.4.1.1: The entry semantics for these expanded states, (in other words), the considerations for entering these states, are conveyed to OSPM by the C state Type field and correspond to the entry semantics for C1 C2 and C3 as described in [Section 8.1.2](https://uefi.org/specs/ACPI/6.6/08_Processor_Configuration_and_Control.html#processor-power-state-c1) through [Section 8.1.4](https://uefi.org/specs/ACPI/6.6/08_Processor_Configuration_and_Control.html#processor-power-state-c3)._

I would strongly prefer ACPICA to follow the ACPI Spec as much as possible and as such if a new C state type is defined, then the ACPI Spec ECR needs to change the spec language to reflect the reality. Please let me know your thoughts on this and thanks a lot once again!

superm1:
> so I am just seeking clarification from you if you actually intend on defining a new C state type C4 or if it maps to HW C3 or some other mapping per the platform?

The best way I would describe it's usage on this given platform is that what C3 "used to do" is what C4 does now.   So if you don't support C4 you can't get to deepest state.  C3 is "kinda like" C2.5.

> I would strongly prefer ACPICA to follow the ACPI Spec as much as possible and as such if a new C state type is defined, then the ACPI Spec ECR needs to change the spec language to reflect the reality. Please let me know your thoughts on this and thanks a lot once again!

Given the section 8.1.5 you can see the behavior of C4..Cn mentioned already.  What do you have in mind for changes?

> ACPI introduced optional processor power states beyond C3 starting in ACPI 2.0. These power states, C4… Cn, are conveyed to OSPM through the _CST object defined in [_CST (C States)](https://uefi.org/specs/ACPI/6.6/08_Processor_Configuration_and_Control.html#cst-c-states) These additional power states are characterized by equivalent operational semantics to the C1 through C3 power states, as defined in the previous sections, but with different entry/exit latencies and power savings. See [_CST (C States)](https://uefi.org/specs/ACPI/6.6/08_Processor_Configuration_and_Control.html#cst-c-states) for more information.

SaketADumbre:
OK, interesting! Per section 8.1.5 then you won't need any changes in the ACPI Spec as C4....Cn map to either one of the three (C1/C2/C3) C state types in _CST! What I believe ACPICA differentiates is the C state type and not the actual C state number, but I might be wrong and need to verify this. The corresponding data for D or S states hint at this as they too don't go all the way to S10 or D10. If this turns out to be the case then I believe this PR should be reverted back because C4 would still remain a non-existent C state type. The C4 change would then go in the OS kernel driver which differentiates between C3 and C4 with different power saving levels. This is why I was/am asking regarding this change to make sure if it even makes any sense! You might want to make sure that just like me, you also don't misunderstand the difference between C state types vs numbers. It is on me that I rushed to merge this PR, sorry about that! I was oblivious to the difference between C state types (C1/C2/C3) vs numbers (C1....C10) and the ACPI Spec too doesn't help at all in this regard!

rafaeljw:
IMV this shouldn't be merged and this change will not be included in Linux.  The ACPI_STATE_Cx symbols are for FADT C-states (or for _CST C-state types), not for _CST C-state numbers.

@superm1 , you are confusing a C-state number with a C-state type (which admittedly are easy to confuse).

The numbers of C-states in _CST can go beyond 4 and it is not really useful to define symbols for all of these numbers because they are just numbers.  In particular, C4 is the C-state at index 3 in the _CST return package, nothing more.

If your platform uses fewer than 4 C-states (fewer than 4 C-states are returned by _CST), you shouldn't really talk about C4 from the ACPI perspective.

C-state types, on the other hand, are about the handling of the given C-state.  That is, what needs to be done before entering it and after exiting it.  So a C-state of type C1 will be handled like the FADT C1, a C-state of type C2 will be handled like the FADT C2, and a C-state of type C3 will be handled like the FADT C3.  However, there may be multiple C-states of each of these types in the _CST output package.

For example, if _CST returns 4 C-states, one of them may be of "C1" type and all of the other may be of type "C3".

superm1:
Thanks guys.  From the affected system that prompted all of this:

```c
        Name (_CST, Package (0x05)  // _CST: C-States
        {
            0x04, 
            Package (0x04)
            {
                ResourceTemplate ()
                {
                    Register (FFixedHW, 
                        0x02,               // Bit Width
                        0x02,               // Bit Offset
                        0x0000000000000000, // Address
                        ,)
                }, 

                0x01, 
                0x0001, 
                0x00000000
            }, 

            Package (0x04)
            {
                ResourceTemplate ()
                {
                    Register (SystemIO, 
                        0x08,               // Bit Width
                        0x00,               // Bit Offset
                        0x0000000000000414, // Address
                        0x01,               // Access Size
                        )
                }, 

                0x02, 
                0x0012, 
                0x00000000
            }, 

            Package (0x04)
            {
                ResourceTemplate ()
                {
                    Register (SystemIO, 
                        0x08,               // Bit Width
                        0x00,               // Bit Offset
                        0x0000000000000415, // Address
                        0x01,               // Access Size
                        )
                }, 

                0x03, 
                0x010E, 
                0x00000000
            }, 

            Package (0x04)
            {
                ResourceTemplate ()
                {
                    Register (SystemIO, 
                        0x08,               // Bit Width
                        0x00,               // Bit Offset
                        0x0000000000000416, // Address
                        0x01,               // Access Size
                        )
                }, 

                0x04, 
                0x012C, 
                0x00000000
            }
        })
```

That last C-state appears to have "type" 0x4 which is what threw me off for everything.  Looking at 8.4.2.1 I can see that's not actually sensible because C4 is not a valid "type".

I'll send a revert for this change and check with internal team about this.

superm1:
Revert: https://github.com/acpica/acpica/pull/1053

SaketADumbre:
Thanks a lot @superm1 for understanding and for the revert patch, you are awesome! I really appreciate that! Apologies from my end, I should have investigated more before merging this patch.

PR #1047, opened 2025-09-20
Add absent field_obj null check

jtstrs:
The AcpiEvAddressSpaceDispatch function designed 
in such a way that assigning  field_obj to NULL is valid case.
Cover the missed execution path with this check.

Fixes: 0acf24ad7e10 ("ACPICA: Add support for PCC Opregion special context data")


SaketADumbre:
@jtstrs Thanks a lot for fixing this missing case, LGTM!

PR #1050, opened 2025-09-26
Work around bogus -Wstringop-overread warning since GCC 10

xry111:
When ACPI_MISALIGNMENT_NOT_SUPPORTED, GCC can produce a bogus -Wstringop-overread warning, see https://gcc.gnu.org/PR122073.

To me it's very clear that we have a compiler bug here, thus just disable the warning.


SaketADumbre:
Hi @xry111 

We have a problem in gcc regardless of your prior change or not! MSVC compiler sees no issues, no data on llvm/clang, but GCC gets confused with the actual if statement and the #if (especially when it is false) and opening the pragma after the first if is problematic. I see an "else without if" error as the #if interferes when it is pushed and popped. GCC probably has the same underlying functionality/usage for a regular if and #if and the #endif weirdly ends the regular if! Hence, I propose that we enclose the entire if-else if-else construct with the pragma and I did verify that it doesn't throw the same error on GCC! I can push an update commit here and reference it in the comments for your referral.

SaketADumbre:
https://github.com/acpica/acpica/commit/e73917d2e917939f2862bd3097c0efefd589c638

This is my commit that addresses the gcc error, thanks a lot! Sorry about the weird run in with GCC, I didn't expect that since technically #if and a regular if should ideally be separate compiler constructs and one shouldn't ideally affect the other one!

xry111:
> [e73917d](https://github.com/acpica/acpica/commit/e73917d2e917939f2862bd3097c0efefd589c638)
> 
> This is my commit that addresses the gcc error, thanks a lot! Sorry about the weird run in with GCC, I didn't expect that since technically #if and a regular if should ideally be separate compiler constructs and one shouldn't ideally affect the other one!

I suppose that will work.  Should I close this PR or sync it with your commit?

xry111:
Updated my branch anyway.

SaketADumbre:
Thanks a lot @xry111 for this contribution and I really appreciate you for updating your PR! LGTM!

PR #1051, opened 2025-10-03
ACPICA: Add KEYP table definition

davejiang:
Software uses this table to discover the base address of the Key Configuration Unit (KCU) register block associated with each IDE capable host bridge.

[1]: Root Complex IDE Key Configuration Unit Software Programming Guide
     https://cdrdv2.intel.com/v1/dl/getContent/732838

### DIFF
--- a/source/include/actbl2.h
+++ b/source/include/actbl2.h
@@ -176,6 +176,7 @@
 #define ACPI_SIG_ERDT           "ERDT"      /* Enhanced Resource Director Technology */
 #define ACPI_SIG_IORT           "IORT"      /* IO Remapping Table */
 #define ACPI_SIG_IVRS           "IVRS"      /* I/O Virtualization Reporting Structure */
+#define ACPI_SIG_KEYP           "KEYP"      /* Key Programming Interface for IDE */
 #define ACPI_SIG_LPIT           "LPIT"      /* Low Power Idle Table */
 #define ACPI_SIG_MADT           "APIC"      /* Multiple APIC Description Table */
 #define ACPI_SIG_MCFG           "MCFG"      /* PCI Memory Mapped Configuration table */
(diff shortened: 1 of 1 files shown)


djbw:
Thanks @davejiang this is in support of the TDX Connect feature, RFC of this ACPICA enabling here:
http://lore.kernel.org/20250919142237.418648-10-dan.j.williams@intel.com

SaketADumbre:
@davejiang @djbw This isn't the complete KEYP ACPI table support that enables compilation, disassembly and template generation. Is it meant to be like this for now and later the full support would be added? Thanks a lot!

djbw:
> Is it meant to be like this for now and later the full support would be added?

That is my expectation because the immediate need is to just be able to write the consumer side of the kernel implementation. Later when we get to debug use cases that would motivate doing the deeper enabling, but for now unblocking kernel development is the top priority. v6.19 is possibly an intercept for initial [TDX Connect](https://www.intel.com/content/www/us/en/content-details/862706/intel-tdx-connect-architecture-specification.html) support.

SaketADumbre:
> > Is it meant to be like this for now and later the full support would be added?
> 
> That is my expectation because the immediate need is to just be able to write the consumer side of the kernel implementation. Later when we get to debug use cases that would motivate doing the deeper enabling, but for now unblocking kernel development is the top priority. v6.19 is possibly an intercept for initial [TDX Connect](https://www.intel.com/content/www/us/en/content-details/862706/intel-tdx-connect-architecture-specification.html) support.

Sounds good @djbw and that's fair! I would need the full support at some point in the future though! Thanks a lot!

davejiang:
Move to actbl2.h done. Changes pushed.

SaketADumbre:
Thanks @davejiang now it looks good! Please make sure to commit the full support in the future, thanks!

PR #1053, opened 2025-10-08
Revert "Add definition for ACPI C4 C-state"

superm1:
This reverts commit 6215f3c3397ef8b93bac41af20372828c7741fe5. The ACPI specification doesn't allow for anything but "Types" C1, C2, and C3.  We shouldn't be introducing a 4th type. This appears to be a BIOS bug.

### DIFF
--- a/source/include/actypes.h
+++ b/source/include/actypes.h
@@ -751,9 +751,8 @@ typedef UINT64                          ACPI_INTEGER;
 #define ACPI_STATE_C1                   (UINT8) 1
 #define ACPI_STATE_C2                   (UINT8) 2
 #define ACPI_STATE_C3                   (UINT8) 3
-#define ACPI_STATE_C4                   (UINT8) 4
-#define ACPI_C_STATES_MAX               ACPI_STATE_C4
-#define ACPI_C_STATE_COUNT              5
+#define ACPI_C_STATES_MAX               ACPI_STATE_C3
+#define ACPI_C_STATE_COUNT              4
 
 /*
  * Sleep type invalid value


SaketADumbre:
Thanks a lot @superm1 for your quick response, I really appreciate it!

Wer-Wolf:
I admit that i am a bit late with this, but section 8.1.5 ("Additional Processor Power States") of the ACPI specification does indeed specify additional C states beyond C3.

I think this revert was wrong.

superm1:
> I admit that i am a bit late with this, but section 8.1.5 ("Additional Processor Power States") of the ACPI specification does indeed specify additional C states beyond C3.
> 
> I think this revert was wrong.

The usage is for C state types, not C state numbers. I mixed them up from what looks like a BIOS bug.

SaketADumbre:
> I admit that i am a bit late with this, but section 8.1.5 ("Additional Processor Power States") of the ACPI specification does indeed specify additional C states beyond C3.
> 
> I think this revert was wrong.

That is exactly the conclusion I came to, when I first approved and merged the initial change, but somewhere in my mind, I wasn't fully convinced as I saw that even for D and S states, the numbers didn't go all the way up to like 10 or so, 10 being the standard max number in a lot of places as per the ACPI Specification. _CST also contributed to my doubts as it only allows either 1/2/3 in the "type" field. Therefore, I had to open a similar conversation with @rafaeljw and other Linux contributors and maintainers to clarify C states as there are all different kinds like C1E etc. which are non-standard in the ACPI Spec and sometimes enumerated in Linux with other ACPI alternatives like cpu_idle or intel_idle etc. That is when I clarified my confusion regarding C state types vs numbers. The revert is correct as things stand right now! Now a C4 type can be introduced in the ACPI Spec and I would gladly approve and merge it in ACPICA, but someone might need to open an ECR with the ASWG and define it properly in the spec and explain how it differentiates from the C3 type for example, which is already the deepest C state type which is equivalent to the "off" state. The deeper C state numbers (eg. C10) mean different things on different platforms, like voltage off or lack of contact with the power rails etc. I feel that the ACPI Spec needs to be more explicit in differentiating between C state types (1/2/3) vs C state numbers (1....10 or 1....n) so as not to confuse people who are unfamiliar with these exact terminologies!

Wer-Wolf:
I understand, in this case i am ok with the revert.

PR #1054, opened 2025-10-09
Add support for the Microsoft display mux _OSI string

Wer-Wolf:
See: https://learn.microsoft.com/en-us/windows-hardware/drivers/display/automatic-display-switch


SaketADumbre:
The reference webpage indicates the presence of Windows 11 version 24H2 and we have those _OSI string just a few lines above your change in the same file. If you are aware of these newer Windows 11 _OSI strings, could you please add them there, so that we don't miss out newer versions. Thanks a lot @Wer-Wolf for this contribution and for keeping ACPICA up to date with newer additions!

Wer-Wolf:
AFAIK there are no new _OSI strings for Windows 11 24h2 itself, the display mux feature has its own _OSI string.

SaketADumbre:
OK, no worries!

SaketADumbre:
Thanks for your contribution @Wer-Wolf and this PR LGTM!

PR #1055, opened 2025-10-11
Fix potential issue in runtime functional control ImplicitReturn tests.

KaiNorberg:
While working on making an AML interpreter and using the ACPICA runtime tests for validation I stumbled across this potential issue. Its possible I've missed something and/or misunderstood, but here are my findings. My tests were run using `SLCK == 1` ("Slack mode" enabled) and `SETN == 6`.

Starting at line 1050 in `tests/aslts/src/runtime/collections/functional/control/ImplicitReturn/store.asl`, there appears to be a mistake when setting the expected implicitly returned value of the m005(1) method. 
```
Store(0xdddd0000, i003)
CH03("", z138, 0x058, __LINE__, 0)
Store(m005(1), i003)
if (SLCK) {
  CH03("", z138, 0x059, __LINE__, 0)
  if (y901) {
    Store(0, Local0)
  } else {
    Store(Ones, Local0)
  }
  if (LNotEqual(i003, Local0)) {
    err("", z138, __LINE__, 0, 0, i003, Local0)
  }
} else {
  CH04("", 0, 0xff, z138, __LINE__, 0, 0)
}
```

The expected value, `Local0`, changes depending on the value of the `y901` flag, which specifies if "Predicate generates Implicit Return" and is set to `0` when `SETN == 6`. However, checking the definition of the `m005` method, we see that `r001 == 0` when `arg0 == 1` which it in this case does, meaning that the last big while loop is not executed and that the last section of the method that sets ends up setting the implicit result is the following.
```
Store(0xdddd0000, i000)
CH03("", z138, 0x070, __LINE__, 0)
Store(m004(), i000)
if (SLCK) {
  CH03("", z138, 0x071, __LINE__, 0)
  Store(0xaaaa001c, Local0)
  if (LNotEqual(i000, Local0)) {
    err("", z138, __LINE__, 0, 0, i000, Local0)
  }
} else {
  CH04("", 0, 0xff, z138, __LINE__, 0, 0)
}
if (r001) {
// While loop starts down here but there is nothing after the end of this if statement.
```

From my understanding of the expected implicit result behavior, the last evaluated expression should be returned by the method. In this case the last expression is `LNotEqual(i000, Local0)`. I considered the possibility that the `y901` flag being set means that all expressions used to determine a predicate are also not to be considered for implicit return, however in that case the last expression would be `Store(0xaaaa001c, Local0)` which does not align with either `0` or `Ones` as the expected result. So my understanding is that not using predicates for implicit results does not disqualify expressions used to determine the predicate.

Checking the definition of `m004()`, whose result is stored in `i000`.
```
Method(m004)
{
  if (fl00) {
    Return (0)
  }
  switch (Multiply(0xaaaa001a, 1)) {
    case (0) {
      Store(0xaaaa001b, i001)
    }
    case (0xaaaa001a) {
      Store(0xaaaa001c, i001)
    }
    default {
      Store(0xaaaa001d, i001)
    }
  }
}
```
We see that its implicit result is `0xaaaa001c`, since `fl00` is always `0`. This means that the last expression in `m005(1)`, `LNotEqual(i000, Local0)`, evaluates to false or `0` since they are in fact equal. Thus, the expected implicit return value from `m005(1)` is `0` and does not, as far as I can tell, change depending on if the `y901` flag is or is not set.

The fix would seem to be that in the test on line 1050, we should always expect the implicit return value `0` independent of the `y901` flag. As mentioned, I am more than welcome to being wrong on this, the implicit result behavior does seem quite finicky.

Also note that I have not checked that ACPICA actually passes this new version of the test, which would obviously be important to check if this is indeed an incorrect test.

### DIFF
--- a/tests/aslts/src/runtime/collections/functional/control/ImplicitReturn/store.asl
+++ b/tests/aslts/src/runtime/collections/functional/control/ImplicitReturn/store.asl
@@ -1052,11 +1052,7 @@ Method(mf6e,, Serialized)
 	Store(m005(1), i003)
 	if (SLCK) {
 		CH03("", z138, 0x059, __LINE__, 0)
-		if (y901) {
-			Store(0, Local0)
-		} else {
-			Store(Ones, Local0)
-		}
+		Store(0, Local0)
 		if (LNotEqual(i003, Local0)) {
 			err("", z138, __LINE__, 0, 0, i003, Local0)
 		}


KaiNorberg:
While doing some more work I discovered what appears to be an undocumented deviation from the ACPI spec, or at least I haven't been able to find any documentation for it.

I haven't delved into the ACPICA source code to verify, but it appears that the tests expect a zero-length string to convert to a buffer of length `1` containing `0x00`. 

Specifically `tests/aslts/src/runtime/collections/functional/control/switch4.asl` starting on line 1239, it calls `m0e5(Buffer(1){})` expecting the result to be `24`, which, checking the implementation of `m0e5`, would mean that a `""` is considered equal to a `Buffer(1){}`.

In the ACPI Specification Version 6.6 page 939 it says regarding conversion from a string to a buffer that
"If no buffer object exists, a new buffer object is created. If a buffer
object already exists, it is completely overwritten. If the string is
longer than the buffer, the string is truncated before copying. If the
string is shorter than the buffer, the remaining buffer bytes are set
to zero. In either case, the string is treated as a buffer, with each
ASCII string character copied to one buffer byte, including the null
terminator. A null (zero-length) string will be converted to a zero-
length buffer."

As in a zero-length string should convert to a zero-length buffer, if I'm understanding correctly. But it seems ACPICA simply always copies the null-terminator, resulting in a zero-length string becoming a buffer of length `1` containing `0x00` as mentioned above. As that seems to be the only way to pass all tests.

I assume this "issue" has been around for long enough to just be permanent now, checking the commit history does indeed confirm that it's very old. However, would it be possible to clarify this? Or has it already been documented and I've simply missed it? Thank you.

SaketADumbre:
Hi @KaiNorberg 

I am sorry I don't really have all the appropriate context and knowledge regarding the ASLTS and more precisely your concerns above as I have been only working on this project for a little bit over 4 years now and the ASLTS is very old at this point and we are always short on time to prioritize working on it and fixing test coverage! Apologies for the delay, but I will prioritize working on this PR soon, but I ran the ASLTS on your PR branch and I don't see any visible difference in the test count and the # of passed/failed tests! While ACPICA fully intends to follow the ACPI Specification as close as possible, there are a lot of examples where in some places the Spec can be too vague or too narrowly-defined and it is left up to us and the general public consensus on its exact interpretation and practicality (or lack thereof)! Please let me know your thoughts and if there is anything that needs to be discussed first regarding this topic and thanks a lot for contributing!

KaiNorberg:
Hi @SaketADumbre 

Thank you very much for your response and your work on ACPICA. I definitely understand why this kinda issue wouldn't be on the top of the list. (You know you are dealing with something legacy and weird when working on a project for four years is prefixed with "only").

Regarding not seeing any difference in passed/failed tests, that makes sense. The test case would expect the same value as it expects with my modification in the majority of cases. Its only when `SETN=6` that it fails. We see that `SETN` is documented as the following in runmode.asl line 442:

```
    /*
     * Settings number, used to adjust the aslts tests for different releases of ACPICA
     *
     * SETN - settings number of aslts:
     *        0 - release from Bob
     *        1 - release from Bob + my updates
     *        2 - new architecture of Method calculation
     *        3 - fixed bug 263,266
     *        4 - fixed bugs 275,276
     *        5 - fixed bugs 262 (corresponds to the 20070320 release of ACPICA)
     *        6 - 20074403
     *        all the greater - not used yet
     *
     * Used for to adjust some skippings of tests for different ACPICA releases
     * (set up this value manually). See Method SET2 below.
     *
     * Note: the value 5 of SETN corresponds to the 20070320 release of ACPICA.
     */
    Name (SETN, 0x05)
  ```
  
Since the default value is `0x05` the test would behave the exact same in most cases, as such nothing changing is expected.

To back up that this wouldn't happen with `SETN=5` we can check the `SET2` method in runmode.asl line 466.

```
            Case (0x05)
            {
                ...
                Y901 = 0x01 /* Predicate generates Implicit Return since ACPICA release 20080926 */
                ...
            }
            Case (0x06)
            {
                ...
                Y901 = 0x00
                ...
            }
```

We can see that with `SETN=5` the `Y901` flag is set to 0x01` while with `SETN=6` it is set to `0x00`, which will cause the test case to expect the incorrect value.

Regarding the spec being vague and narrow, I can certainly relate to that! I don't think ACPICA should change its behavior on zero-length strings, but it might be a good idea to have it documented somewhere? I don't know if ACPICA has a predefined means of documenting discrepancies.

To summarize, the faulty test case only occurs with a specific configuration (hence why it might have gone undiscovered) and the zero-length strings "issue" is probably just set in stone at this point but documenting it might be cool. Thank you for your time.

PR #1057, opened 2025-10-22
iASL: Fix EINJV2 disassembly

aegl:
ACPI specification added new "action" codes for EINJV2 (and then deprecated one of the new codes in ACPI v6.6).

The new ACPI_EINJV2_GET_ERROR_TYPE was added to enum AcpiEinjActions, but the corresponding AcpiDmEinjActions[] array of action names was not updated.

Result was that "iasl -d" indexed past the end of the action names array and printed spurious names for actions 0x10 and 0x11.

Fix by extending the array for the new action values (note that these are not contiguous with the existing ones, so add "Unassigned" strings for the values that were skipped).

Fixes: 6975cd07e20b ("actbl1: Add EINJv2 get error type action")

### DIFF
--- a/source/common/dmtable.c
+++ b/source/common/dmtable.c
@@ -299,6 +299,14 @@ static const char           *AcpiDmEinjActions[] =
     "Get Command Status",
     "Set Error Type With Address",
     "Get Execute Timings",
+    "Unassigned",
+    "Unassigned",
+    "Unassigned",
+    "Unassigned",
+    "Unassigned",
+    "Unassigned",
+    "EinjV2 Set Error Type(deprecated)",
+    "EinjV2 Get Error Type",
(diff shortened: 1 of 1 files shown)


SaketADumbre:
Thanks a lot @aegl for fixing this! I had an intuition (that something was out of order) back when I merged this PR - https://github.com/acpica/acpica/pull/977 and my old comment mentioned that the changes didn't consider 0xA to 0xF hex values and treated the ordering as decimal (like it used to be before that change). Since the ASWG ECR was already approved (strangely, without anyone questioning the hex ordering), my hands were tied and I didn't completely foresee this parsing to be problematic! Glad that we are getting it fixed now! Really appreciate your contributions! LGTM!

PR #1058, opened 2025-10-27
Abort AML bytecode execution when executing AML_FATAL_OP

Wer-Wolf:
The ACPI specification states that when executing AML_FATAL_OP, the OS should log the fatal error event and shutdown in a timely fashion. Windows complies with this requirement by immediatly entering a BSoD, effectively aborting the execution of the AML bytecode in question.

ACPICA however might continue with the AML bytecode execution should AcpiOsSignal() simply return AE_OK. This will cause issues because ACPI BIOS implementations might assume that the Fatal() operator does not return.

Fix this by aborting the AML bytecode execution in such a case by returning AE_ERROR. Also turn ACPI_SIGNAL_FATAL_INFO into a local variable because of its small size (12 bytes) and to ensure that AcpiOsSignal() always receives valid information about the fatal ACPI BIOS error.

### DIFF
--- a/source/components/executer/exoparg3.c
+++ b/source/components/executer/exoparg3.c
@@ -153,6 +153,7 @@
 #include "acpi.h"
 #include "accommon.h"
 #include "acinterp.h"
+#include "acoutput.h"
 #include "acparser.h"
 #include "amlcode.h"
 
@@ -201,8 +202,7 @@ AcpiExOpcode_3A_0T_0R (
     ACPI_WALK_STATE         *WalkState)
 {
     ACPI_OPERAND_OBJECT     **Operand = &WalkState->Operands[0];
-    ACPI_SIGNAL_FATAL_INFO  *Fatal;
-    ACPI_STATUS             Status = AE_OK;
+    ACPI_SIGNAL_FATAL_INFO  Fatal;
 
 
     ACPI_FUNCTION_TRACE_STR (ExOpcode_3A_0T_0R,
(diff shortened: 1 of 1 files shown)


Wer-Wolf:
Any thoughts on this?

SaketADumbre:
This is a good fix, thanks for contributing @Wer-Wolf and apologies for the delay. Everything looks good and in order!

SaketADumbre:
I see some regression in the ASL Test Suite after merging this PR/commit last night. The plan is to understand why some more tests fail (+19 tests per category) but also to leave room to revert this PR/commit if no results after further testing/tweaking. I am sorry I should have caught this sooner, but luckily I began investigating another PR that supposedly helps the ASLTS and hence ran the numbers. Also on the back burner (task for me) is the previous ASLTS regression that drops the test coverage overall (pass + fail tests reduced) and a git bisect might help pinpoint the exact commit (most likely scenario over multiple culprit commits).

ASLTS_regression_1058

Wer-Wolf:
I would love to use those test, sadly i cannot install them under Linux (ls cannot find the aml files)

Can you give me a hint?

Wer-Wolf:
I think the errors happen because some tests test object conversions using the `Fatal` opcode. This was only possible as long as said opcode always returned `AE_OK`, but will fail now.

Maybe i can introduce something similar to `ACPI_UNDER_DEVELOPMENT` for causing `Fatal` to continue the execution instead of aborting it.

SaketADumbre:
> I would love to use those test, sadly i cannot install them under Linux (ls cannot find the aml files)
> 
> Can you give me a hint?

You can run the test suite by going inside "tests" directory (cd tests) and simply running the ./aslts.sh script. Each individual test generates its corresponding aml and is then recompiled and the asl and dsl compared, but those transitionary files are deleted after use! You can even modify individual tests and set a few flags that help with control and logic flow, but the ASLTS is really old and not a finished product! I haven't been able to prioritize it up until now, but looks like it might be a future area for me to tackle!

SaketADumbre:
> I think the errors happen because some tests test object conversions using the `Fatal` opcode. This was only possible as long as said opcode always returned `AE_OK`, but will fail now.
> 
> Maybe i can introduce something similar to `ACPI_UNDER_DEVELOPMENT` for causing `Fatal` to continue the execution instead of aborting it.

I think those tests in question just rely on dumb logic to pass and expect AE_OK return status. I haven't looked into this yet, maybe something for you to investigate maybe? That being said, if it is indeed the cause for those ~19 tests per category failing, then we might need to reintroduce the goto cleanup part where regardless of the output, it simply returned AE_OK like before your change!

Wer-Wolf:
I will try to come up with a solution for this.

PR #1059, opened 2025-11-11
Complete definition of Cache Type Structure for PPTT v3

benhor01:
Hi,

I am looking to use  ACPI_PPTT_CACHE_V1 (struct acpi_pptt_cache_v1) in the headers exported to linux but found it only contains the single field added in PPTT v3 rather than the whole structure. I've reworked it to include the whole structure.

Is this kind of change acceptable? It could affect other users of acpica headers. Let me know if this can be done in a better way.

In linux I tried adding a new struct:
struct acpi_pptt_cache_v1_full {
 struct acpi_pptt_cache		f;
 struct acpi_pptt_cache_v1	extra;
} __packed;
but this is just papering over the problem.

See: https://lore.kernel.org/linux-arm-kernel/1a52079f-e12f-430b-b3c5-cb184f73c6c7@arm.com/

Thanks,

Ben


DCH-DYM:
这是来自QQ邮箱的假期自动回复邮件。
 
您好，我最近正在休假中，无法亲自回复您的邮件。我将在假期结束后，尽快给您回复。

SaketADumbre:
Is ACPI_PPTT_CACHE_V1 specific to only ARM ecosystem? Can you also help me understand why you'd need cache v1 directly in the header over the current implementation? Your change basically duplicates the newly added fields inside the cache v1 struct and either compiles the old cache or new cache v1 usage depending on if revision < 3 or not, which is just a refactoring of the original code IMHO. Thanks for updating the template, which signifies that your change works correctly as intended, but it doesn't look significantly different than the existing template.

benhor01:
Hi Saket,

Thanks for looking at this.

> Is ACPI_PPTT_CACHE_V1 specific to only ARM ecosystem? 

It's in the generic acpi spec and so is usable across architectures. In the linux kernel RISC-V and Loongarch also support building the cacheinfo from the PPTT and will use v1 of the cache structure if the PPTT is a late enough revision. The addition of cache_id was initiated by arm to give a way to reference one table from another.

> Can you also help me understand why you'd need cache v1 directly in the header over the current implementation?

This allows for cleaner code in linux and matches the other versioned structures in the acpi headers. The code is cleaner because:
1. No need for additional indirection through a new structure
2. The size of the structure can be used to check the version
3. It matches expectations.

However, this change is not compulsory, only nice to have, as the cache id support works without it.

> Your change basically duplicates the newly added fields inside the cache v1 struct and either compiles the old cache or new cache v1 usage depending on if revision < 3 or not, which is just a refactoring of the original code IMHO.

Yes. No functional change other than what's in the header was intended.

> Thanks for updating the template, which signifies that your change works correctly as intended, but it doesn't look significantly different than the existing template.

Indeed, just a PPTT version bump and adding the cache id.

Thanks,

Ben

SaketADumbre:
Hi Ben,

While you are right that from the Linux kernel perspective, this refactoring is cleaner, but from the ACPICA codebase perspective it adds some duplication and violates the DRY (Don't Repeat Yourself) coding principle, although in all honesty and practicality, this is not a big deal in this situation and the number of lines aren't blowing up significantly and hence I have no objections. I will go ahead and merge this PR, thanks for your contribution!

benhor01:
Thanks! It's appreciated and sorry it caused churn in ACPICA.

PR #1063, opened 2025-11-24
Fix asltests using the Fatal() opcode

Wer-Wolf:
Some asltests test the behavior of the Fatal() opcode and thus require that said opcode does not return an error when called.

Introduce a compile-time option called ACPI_CONTINUE_ON_FATAL to instruct the executor to continue the execution of AML bytecode when encountering a Fatal() opcode. Also update the asltest to use this new option.

This fixes the test regressions observed after #1058.

### DIFF
--- a/source/components/executer/exoparg3.c
+++ b/source/components/executer/exoparg3.c
@@ -225,11 +225,18 @@ AcpiExOpcode_3A_0T_0R (
 
         AcpiOsSignal (ACPI_SIGNAL_FATAL, &Fatal);
 
+#ifndef ACPI_CONTINUE_ON_FATAL
         /*
          * Might return while OS is shutting down, so abort the AML execution
          * by returning an error.
          */
         return_ACPI_STATUS (AE_ERROR);
+#else
+	/*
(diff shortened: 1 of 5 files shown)


Wer-Wolf:
Any thoughts on this?

SaketADumbre:
Hi @Wer-Wolf 

I will need to circle back to this PR next week since I need to test it (ASLTS test count verification) on my Linux machine. Thanks for following up to your original PR #1058 promptly and I appreciate your patience during these busy holiday times!

Wer-Wolf:
Any progress on this?

SaketADumbre:
Hi @Wer-Wolf

Apologies for the delayed response, I am back to work today from my vacation! I just finished testing this PR and can confirm that it does indeed revert the 19 test failures (for each of the 4 categories: 32 bit vs. 64 and opt. vs. non-opt.) and now I can see that those 19 tests are back to passing again! Thanks a lot for your contributions, I really appreciate them! LGTM!

Wer-Wolf:
Thank you :)

PR #1069, opened 2026-01-02
Add TPM2 start methods from TPM ACPI Spec 1.40

JohnAZoidberg:
See: https://trustedcomputinggroup.org/wp-content/uploads/TCG-ACPI-Specification-Version-1.4-Revision-15_pub.pdf

### DIFF
--- a/source/common/dmtable.c
+++ b/source/common/dmtable.c
@@ -568,6 +568,10 @@ static const char           *AcpiDmTpm2Subnames[] =
     "Reserved",
     "Reserved",
     "Command Response Buffer with ARM SMC",
+    "FIFO over I2C",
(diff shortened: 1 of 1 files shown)


JohnAZoidberg:
We use this on our Framework systems with Ryzen AI 300 CPUs.
See: https://github.com/FrameworkComputer/SoftwareFirmwareIssueTracker/issues/109

JohnAZoidberg:
@stuyod01 in #1000 (098c00258799f4dc32d718ceb4e83f70a9c9ffc2) added the definitions but didn't add the strings.

SaketADumbre:
Thanks a lot @JohnAZoidberg for this contribution! This was a good catch!

PR #1101, opened 2026-02-04
Update the _CPC definition to also support return package

SaketADumbre:
which is the sub-type of the parent package (with Integer and Buffer) as per ACPI Spec 6.6


mrhpearson:
Tested on Lenovo X1 Carbon G14 and confirmed it fixed the boot errors we were previously seeing - thanks!

Tested-by: Mark Pearson <mpearson-lenovo@squebb.ca>

JohnAZoidberg:
For reference, this is the dmesg warning:

> ACPI Warning: \_SB.PR00._CPC: Return Package type mismatch at index 24 - found Package, expected Integer/Buffer

SaketADumbre:
Thanks for testing this change @mrhpearson and @JohnAZoidberg

PR #1103, opened 2026-02-06
Fix FADT 32/64X length mismatch warning

seuros:
When the 64-bit address is set but BitWidth is 0, the spec says the legacy length should be used. That is valid firmware.

Skip the warning if BitWidth is 0.

This avoids false warnings like:
  `32/64X length mismatch in FADT/Gpe0Block: 128/0`

Tested on FreeBSD 16.0-CURRENT.


SaketADumbre:
Described from Section 5.2.3 to 5.2.9 in the ACPI 6.6 Specification! Thanks for contributing this fix @seuros !

PR #1104, opened 2026-02-10
Add Framework Computer Device HID IDs

JohnAZoidberg:
The different ACPI HID IDs we use on our systems.


JohnAZoidberg:
fyi @kiram9

source/common/ahids.c
```diff
@@ -203,6 +203,14 @@ const AH_DEVICE_ID  AslDeviceIds[] =
     {"DELLABCE",    "Dell Airplane Mode Switch Driver"},
     {"DLAC3002",    "Qualcomm Atheros Bluetooth UART Transport"},
     {"FTTH5506",    "FocalTech 5506 Touch Controller"},
```
SaketADumbre: Nitpick: Wouldn't this FTTH5506 go below all your changes @JohnAZoidberg according to the alphabetical order?

SaketADumbre:
To demonstrate that I am not such a perfectionist, I will leave that change as optional, but it would be better to take care of it just because the ordering is strictly alphabetical and I didn't see a single out of order instance there! Thanks for contributing @JohnAZoidberg

source/common/ahids.c
JohnAZoidberg: No problem, I updated it :D
Thanks for the quick feedback

SaketADumbre:
Thanks @JohnAZoidberg for your prompt response! Everything looks good now!

PR #1107, opened 2026-02-26
fix I2C LVR item count in the conversion table

akhilr-nv:
For ACPI_RSC_MOVE8, the 'Value' field in acpi_rsconvert_info is the item count and not a bit position like for the bitflags. Set 'Value' for the LVR field as '1' to fix this.

Conversion still works coincidentally with '0' because item_count is not reset between table entries, and the previous count value was taking effect.


SaketADumbre:
Thanks for catching and fixing this soon @akhilr-nv

PR #1108, opened 2026-02-28
Fix missing break in switch statment for case ACPI_MPAM_LOCATION_TYPE…

ColinIanKing:
Currently the ACPI_MPAM_LOCATION_TYPE_UNKNOWN case statement is missing a break statement, so it falls through to the default case where a fatal error message is reported. Fix this by adding the missing break statement.

Fixes: 005e24bcaa6e ("Add support for Arm's MPAM ACPI table version 2")

### DIFF
--- a/source/compiler/dttable2.c
+++ b/source/compiler/dttable2.c
@@ -603,6 +603,7 @@ DtCompileMpam (
                     break;
                 case ACPI_MPAM_LOCATION_TYPE_UNKNOWN:
                     InfoTable = AcpiDmTableInfoMpam1G;
+		    break;
                 default:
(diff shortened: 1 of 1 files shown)


ColinIanKing:
No, this is not a bug. Closing it.

SaketADumbre:
Hi @ColinIanKing 
Thanks for noticing this, although it is still the same "unknown" resource type case and we won't need a separate break statement but for it to take the default path! Great to see you here again and enjoy your retirement! I would like to express my heartfelt gratitude for your service all these years for both Intel and the open-source software community in general and I hope someone continues sponsoring your future contributions! Please keep in touch and let me know if there is anything I can do for you!

Issue #1115, opened 2026-03-16
Heap Buffer Overflow in AcpiPsGetNextSimpleArg


void0red:
## Summary
Heap-buffer-overflow in `AcpiPsGetNextSimpleArg` function at psargs.c, affecting multiple code paths (lines 560, 569, 578, 587, 601).

### Line 560
```
=================================================================
==3981419==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x50f0000000ef at pc 0x5dbb0c290d8e bp 0x7fff3d962b60 sp 0x7fff3d962b50
READ of size 1 at 0x50f0000000ef thread T0
    #0 0x5dbb0c290d8d in AcpiPsGetNextSimpleArg ../../../source/components/parser/psargs.c:560
    #1 0x5dbb0c290f14 in AcpiPsGetNextArg ../../../source/components/parser/psargs.c:971
    #2 0x5dbb0c2928d1 in AcpiPsGetArguments ../../../source/components/parser/psloop.c:263
    #3 0x5dbb0c2928d1 in AcpiPsParseLoop ../../../source/components/parser/psloop.c:546
    #4 0x5dbb0c297823 in AcpiPsParseAml ../../../source/components/parser/psparse.c:633
    #5 0x5dbb0c29a61c in AcpiPsExecuteTable ../../../source/components/parser/psxface.c:467
    #6 0x5dbb0c281437 in AcpiNsExecuteTable ../../../source/components/namespace/nsparse.c:269
    #7 0x5dbb0c281f35 in AcpiNsParseTable ../../../source/components/namespace/nsparse.c:436
    #8 0x5dbb0c27eb73 in AcpiNsLoadTable ../../../source/components/namespace/nsload.c:226
SUMMARY: AddressSanitizer: heap-buffer-overflow ../../../source/components/parser/psargs.c:560 in AcpiPsGetNextSimpleArg
```

### Line 569
```
=================================================================
==3134065==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x50c0000003b9 at pc 0x560402004dc2 bp 0x7ffd169811c0 sp 0x7ffd169811b0
READ of size 2 at 0x50c0000003b9 thread T0
    #0 0x560402004dc1 in AcpiPsGetNextSimpleArg ../../../source/components/parser/psargs.c:569
    #1 0x560402006c9b in AcpiPsGetArguments ../../../source/components/parser/psloop.c:222
[...]

Issue #1073, opened 2026-01-03
AddressSanitizer: heap-buffer-overflow ../../../source/components/parser/psargs.c:560 in AcpiPsGetNextSimpleArg


void0red:
Hello, developer
We are testing the iasl compiler and found this memory bug.

reproduce:
```bash
iasl -d fuzz_302.aml
```

AddressSanitizer log:
```log
==2303776==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x51100000013b at pc 0x64066984819e bp 0x7ffff2893320 sp 0x7ffff2893310
READ of size 1 at 0x51100000013b thread T0
    #0 0x64066984819d in AcpiPsGetNextSimpleArg ../../../source/components/parser/psargs.c:560
    #1 0x640669848324 in AcpiPsGetNextArg ../../../source/components/parser/psargs.c:971
    #2 0x640669849d60 in AcpiPsGetArguments ../../../source/components/parser/psloop.c:263
    #3 0x640669849d60 in AcpiPsParseLoop ../../../source/components/parser/psloop.c:546
    #4 0x64066984ef83 in AcpiPsParseAml ../../../source/components/parser/psparse.c:633
    #5 0x6406697d30d1 in AdParseTable ../../../source/common/dmtables.c:554
    #6 0x640669758e1d in AdDisassembleOneTable ../../../source/common/adisasm.c:504
    #7 0x640669758e1d in AdAmlDisassemble ../../../source/common/adisasm.c:391
    #8 0x6406697a53e5 in AslDoDisassembly ../../../source/compiler/aslstartup.c:368
    #9 0x6406697a53e5 in AslDoDisassembly ../../../source/compiler/aslstartup.c:351
    #10 0x640669718b9c in main ../../../source/compiler/aslmain.c:281
    #11 0x799632e29d8f in __libc_start_call_main ../sysdeps/nptl/libc_start_call_main.h:58
    #12 0x799632e29e3f in __libc_start_main_impl ../csu/libc-start.c:392
    #13 0x64066971ae54 in _start (/home/ikaros/acpica/generate/unix/bin/iasl+0xe9e54)

0x51100000013b is located 0 bytes to the right of 251-byte region [0x511000000040,0x51100000013b)
allocated by thread T0 here:
    #0 0x7996332b4887 in __interceptor_malloc ../../../../src/libsanitizer/asan/asan_malloc_linux.cpp:145
[...]

PR #1116, opened 2026-03-16
Fix heap-buffer-overflow in AcpiPsGetNextSimpleArg

void0red:
Add boundary checks before reading bytes from AML stream to prevent heap-buffer-overflow when parsing malformed AML.

Affected code paths: ARGP_BYTEDATA, ARGP_WORDDATA, ARGP_DWORDDATA, ARGP_QWORDDATA, ARGP_CHARLIST.

Fixes: #1115


void0red:
same as #1073

PR #1194, opened 2026-06-05
generate/efi: add missing utcksum.c source file

Icenowy:
The utcksum.c source file is currently missing from EFI builds, which make them to never succeed.

### DIFF
--- a/generate/efi/acpidump/Makefile
+++ b/generate/efi/acpidump/Makefile
@@ -41,6 +41,7 @@ OBJECTS = \
 	$(OBJDIR)/utascii.o\
 	$(OBJDIR)/utbuffer.o\
 	$(OBJDIR)/utclib.o\
+	$(OBJDIR)/utcksum.o\
 	$(OBJDIR)/utdebug.o\
 	$(OBJDIR)/utexcep.o\
 	$(OBJDIR)/utglobal.o\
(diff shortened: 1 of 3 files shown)


rafaeljw (approved):

maciejwieczorretman:
Is this for the gnu-efi path or the EDK one?
Does "make OS=efi" actually run for you successfully? I find there are some other build problems on my Arch Linux system (but I think they should be present on other distros too)
- ld having -nostdinc flag which ld doesn't support
- gnu-efi exporting memset and memcpy which colides with acpica exported functions
- objcopy having issues with the file format

Icenowy:
It's missing in both codepaths.
The .inf changes are for EDK and the Makefile change is for gnu-efi, although I only verified the resulting binary on EDK (I did got a riscv64 gnu-efi binary but it doesn't work; and I didn't bother to check x86 gnu-efi binary).

maciejwieczorretman:
Okay, I think I got what was wrong with my build and probably why the riscv64 binary didn't work too
Seems like OBJCOPYFLAGS doesn't have the .rodata section which causes a fault when running the binary
And x86 doesn't build right now because some objcopy flags are incorrectly used
I'm putting together a script to test it with qemu for future reference
And I'll merge #1194 and #1195 today I think. Maybe #1196 too if I can make it
Thanks for fixing these build issues :)

maciejwieczorretman:
Okay, merged it manually

PR #1195, opened 2026-06-05
generate/efi: adapt to EDK2 202605

Icenowy:
Recent EDK2's BaseLib has some more dependencies declared in a .inc file in MdePkg, and the toolchain name is changed to contain no version number.

Adapt to these changes in the DSC files and README.

Depends on #1194 to pass build.


maciejwieczorretman:
Merged this manually. Thanks for fixing this :)

PR #1196, opened 2026-06-05
EFI: add RISCV64 support

Icenowy:
Depends on #1194 and #1195 which serve as general EFI build fixes.

Building with StdLib requires https://github.com/tianocore/edk2-libc/pull/114 .


maciejwieczorretman:
Merged #1194 and #1195. I'll try to test this PR in qemu. Also I think I found what was wrong (the .rodata was missing) and added a few commits that fix it (and fix the build process for me). So you can also check if it works on riscv64 now. I'll try to check in qemu on my end

Icenowy:
I rebased this branch atop master, but unfortunately when I ran the gnu-efi-built riscv64 acpidump.efi, it's still `Command Error Status: Unsupported` .

The debug output of the EFI firmware is the following:

```
FS0:\> acpidump.efi
FSOpen: Open '\acpidump.efi' Success
FSOpen: Open '\acpidump.efi' Success
FSOpen: Open '\acpidump.efi' Success
FSOpen: Open '\acpidump.efi' Success
[Security] 3rd party image[0] can be loaded after EndOfDxe: VenHw(837DCA9E-E874-4D82-B29A-23FE0E23D1E2,0070001000000000)/HD(1,GPT,E01DE564-1E90-4CFD-A664-07714D7B0DDC,0x800,0x400000)/\acpidump.efi.
InstallProtocolInterface: 5B1B31A1-9562-11D2-8E3F-00A0C969723B 47F121040
Command Error Status: Unsupported
FSOpen: Open '\' Success
```

Icenowy:
The preamble of debug output of running a EDK2-built acpidump binary as a reference:

```
FS0:\> acpiedk2
FSOpen: Open '\acpiedk2.EFI' Success
FSOpen: Open '\acpiedk2.EFI' Success
FSOpen: Open '\acpiedk2.EFI' Success
FSOpen: Open '\acpiedk2.EFI' Success
[Security] 3rd party image[0] can be loaded after EndOfDxe: VenHw(837DCA9E-E874-4D82-B29A-23FE0E23D1E2,0070001000000000)/HD(1,GPT,E01DE564-1E90-4CFD-A664-07714D7B0DDC,0x800,0x400000)/\acpiedk2.EFI.
InstallProtocolInterface: 5B1B31A1-9562-11D2-8E3F-00A0C969723B 47F120040
Loading driver at 0x0047E2DB000 EntryPoint=0x0047E2DB240 acpidump.efi
InstallProtocolInterface: BC62157E-3E33-4FEC-9920-2D3B36D750DF 47F120E98
ProtectUefiImageCommon - 0x7F120040
  - 0x000000047E2DB000 - 0x0000000000009740
!!!!!!!!  Image Section Alignment(0x40) does not match Required Alignment (0x1000)  !!!!!!!!
ProtectUefiImage failed to create image properties record
InstallProtocolInterface: 752F3136-4E16-4FDC-A22A-E5F46812F4CA 83FFF350
```
(Then follows hexdumps of tables)

maciejwieczorretman:
Does riscv64 build correctly with these commands that I added to the README?
```bash
   # EDK_ARCH=$(uname -m | sed -e 's/x86_64/X64/' -e 's/i[3456789]86/IA32/')
   # build -p AcpiPkg/AcpiPkg.dsc -t GCC -a $EDK_ARCH -D DEBUG_ENABLE_OUTPUT=FALSE
```
I added sed rules for x86 and IA32 but if the arch string doesn't get recognized by the 'build' utility there it defaults to IA32. Maybe that happened in this case?

Icenowy:
Ah I mean the gnu-efi built thing is broken; the EDK2 built thing is okay.

BTW the EDK2 architecture should be "RISCV64".

maciejwieczorretman:
Okay, I'll recheck the gnu-efi path then
I know the build string is RISCV64, what is the string returned from "uname -m"? Does it need a different sed rule in the readme?

Icenowy:
The uname string is just `riscv64`.
Maybe add a path translating all lowercase and uppercase? This will then work for `aarch64` uname result too.

Icenowy:
Ah the `uname -m` result for EDK2 build string LOONGARCH64 is also just `loongarch64` .

Icenowy:
I propose:

```
EDK_ARCH="$(uname -m | sed -e 's/x86_64/X64/' -e 's/i[3456789]86/IA32/' -e 's/./\U&/g')"
```

maciejwieczorretman:
Yes, that should be okay. So the EDK2 way should work with how the PR is currently looking right?
About gnu-efi I suspect that the Makefile.config is missing stuff. Like around line 185 there is an if/else which check if arch is ia32 and if it's not then it goes to x86_64 flags. Maybe that part is missing some riscv64 flags?

Also the maybe a different output format is needed for the objcopy? I added the '-O efi-app-$(TARGET)' option since older 'target' was missing the gnu-efi options. But I can't verify what the riscv64 formats are supported from gnu-efi.

Anyway if you want to we can probably merge this already just with a sidenote what it is for EDK2 path only and that gnu-efi doesn't seem to work now on riscv64

Icenowy:
Yes, with this PR and EDK2 build system proper RISCV64 binary is produced.

source/os_specific/efi/osefitbl.c
```diff
@@ -704,10 +704,14 @@ OslTableInitialize (
         return (Status);
     }
 
-    Status = OslAddTableToList (ACPI_SIG_FACS, 0);
-    if (ACPI_FAILURE (Status))
-    {
-        return (Status);
+    if (!(Gbl_Fadt->Flags & ACPI_FADT_HW_REDUCED) ||
```
maciejwieczorretman, line 707: I think this part of the if clause can be removed? If you look at lines ~884 in this file (osefitbl.c) then there is a similar check already related to FACS.

And I didn't find a hard correlation between hardware reduced acpi and FACS missing. So I assume it's possible there may be non-reduced hardware that doesn't have FACS, and this check will let it through.

I can remove it and merge the three patches unless you think this makes sense here, then we can talk some more about it.

maciejwieczorretman (changes requested):
Found this one thing that may be a problem, let me know what you think. Generally I noticed that there are multiple places that do OslAddTableToList() with FACS but don't actually check MIN_FADT_FOR_[X]FACS. But I'll take care of that separately later. Cool that it got brought up here :)

source/os_specific/efi/osefitbl.c
Icenowy: In ACPI spec version 6.6, chapter 5.2.10 Firmware ACPI Control Structure (FACS), such a sentence exists: `This structure is optional if and only if the HARDWARE_REDUCED_ACPI flag in the FADT is set.`

Icenowy: Oh I followed the following sentence in the FADT definition:
```
If the HARDWARE_REDUCED_ACPI flag is set, and both this field and the X_FIRMWARE_CTRL field are zero, there is no FACS available.
```
I think I followed this sentence when writing the code, and it seems to indicate that an extreme side case is possible -- on non-HW-reduced ACPI platforms the FACS can be at 0x0 (although I think this sounds ridiculous).

Icenowy:
Oh weird the gnu-efi-built riscv64 .efi file has 2 MZ headers...
Something might be messed up.

Icenowy:
Oh the problem is that my gnu-efi is too old and that version of gnu-efi assumes objcopy has no `-O efi-app-riscv64` so it comes with its own PE header in the .text section.
Using `elf_riscv64_efi_local.lds` and `crt0-efi-riscv64-local.o` on the newest gnu-efi can still replicate this behavior (however it's weird that the crafted PE header will has subsystem set to EFI Runtime Driver instead of EFI Application; gnu-efi Makefile uses a weird 1-byte sed to fix this).
The solution is just to use new enough gnu-efi (4.0.0+) and objcopy (2.42+).

source/os_specific/efi/osefitbl.c
maciejwieczorretman: Ah okay, thanks for the chapter number. I was reading the 4.1 chapter and there was nothing there and in other places in the code. I'll add the hw reduced checks elsewhere too.

maciejwieczorretman:
Okay, so does the gnu-efi work too after you updated your binaries? All the other code looks good, so just wanted to make sure what works and what doesn't for the documentation.

Icenowy:
Yes after updating my gnu-efi it works quite well. For documentation purpose I think the version requirement could be documented as what I mentioned in my previous comment.

BTW the build system seems to have no option to configure for a gnu-efi installation not at `/usr/{include,lib}` , I have to hack the Makefile to use a non-system gnu-efi.

maciejwieczorretman:
Okay, if it works then I'll merge it in a second

And for now I'll keep the system gnu-efi only, but I'll come back to it once I deal with other serious bugs in issues/PRs

maciejwieczorretman:
Merged manually, I just moved one parenthesis in the big if clause so it matches the coding style.
Thanks for the patches and have a nice day :)

PR #1197, opened 2026-06-06
Set AML length for bytelist

Rongronggg9:
The AML length of bytelist is never set. Since the Fixes: commit, this leads to failures in the sanity check for resource templates, causing the decompiler to decompile valid resource templates as plain buffers.

Fix it by setting the AML length properly.

Fixes: 3ce09021b677 ("validate resource template buffer length and check allocation type")

### DIFF
--- a/source/components/parser/psargs.c
+++ b/source/components/parser/psargs.c
@@ -973,6 +973,8 @@ AcpiPsGetNextField (
 
                 ASL_CV_CAPTURE_COMMENTS_ONLY (ParserState);
                 Arg->Named.Value.Size = BufferLength;
+                Arg->Named.Length = (UINT32)
+                    ACPI_PTR_DIFF (PkgEnd, ParserState->Aml);
                 Arg->Named.Data = ParserState->Aml;
             }
 
@@ -1167,7 +1169,7 @@ AcpiPsGetNextArg (
 
             /* Fill in bytelist data */
 
-            Arg->Common.Value.Size = (UINT32)
+            Arg->Common.Value.Size = Arg->Named.Length = (UINT32)
                 ACPI_PTR_DIFF (ParserState->PkgEnd, ParserState->Aml);
             Arg->Named.Data = ParserState->Aml;
 


Rongronggg9:
To reproduce the issue:

Compile the DSL:
```dsl
DefinitionBlock ("", "SSDT", 1, "RSRC", "TEMPLATE", 0x00000001)
{
    Name (RBUF, ResourceTemplate ()
    {
        Memory32Fixed (ReadWrite,
            0xCB45F000,         // Address Base
            0x00004000,         // Address Length
            )
        Memory32Fixed (ReadWrite,
            0xCB463000,         // Address Base
            0x00004000,         // Address Length
            )
    })
}
```

Decompile it with 20260408:
```dsl
/*
 * Intel ACPI Component Architecture
 * AML/ASL+ Disassembler version 20260408 (64-bit version)
 * Copyright (c) 2000 - 2026 Intel Corporation
 * 
 * Disassembling to symbolic ASL+ operators
 *
 * Disassembly of rt_compiled.aml
 *
 * Original Table Header:
 *     Signature        "SSDT"
 *     Length           0x00000047 (71)
 *     Revision         0x01
 *     Checksum         0x77
 *     OEM ID           "RSRC"
 *     OEM Table ID     "TEMPLATE"
 *     OEM Revision     0x00000001 (1)
 *     Compiler ID      "INTL"
 *     Compiler Version 0x20260408 (539362312)
 */
DefinitionBlock ("", "SSDT", 1, "RSRC", "TEMPLATE", 0x00000001)
{
    Name (RBUF, Buffer (0x1A)
    {
        /* 0000 */  0x86, 0x09, 0x00, 0x01, 0x00, 0xF0, 0x45, 0xCB,  // ......E.
        /* 0008 */  0x00, 0x40, 0x00, 0x00, 0x86, 0x09, 0x00, 0x01,  // .@......
        /* 0010 */  0x00, 0x30, 0x46, 0xCB, 0x00, 0x40, 0x00, 0x00,  // .0F..@..
        /* 0018 */  0x79, 0x00                                       // y.
    })
}
```

With the patch:
(the same as the source DSL above)

This makes me doubt if the buggy commit has ever been tested seriously...

source/components/parser/psargs.c
```diff
@@ -973,6 +973,8 @@ AcpiPsGetNextField (
 
                 ASL_CV_CAPTURE_COMMENTS_ONLY (ParserState);
                 Arg->Named.Value.Size = BufferLength;
+                Arg->Named.Length = (UINT32)
+                    ACPI_PTR_DIFF (ParserState->PkgEnd, ParserState->Aml);
```
maciejwieczorretman: I believe in this context ParserState->PkgEnd is actually the outer package boundry, not the one that is parsed in AcpiPsGetNextField.
So I think to use variables appropriately it should be the local variable instead - 'PkgEnd'.

I dug into this and actually both versions work as long as BufferLength is smaller or equal to either PkgEnd or ParserState->PkgEnd. The problem was when Length wasn't initialized and it was 0, then the type was getting misinterpreted.

maciejwieczorretman (commented):
Thanks for sending the PR! Reproduced it and I see that your code works, just one comment there. I'll make the change and apply the patch next week unless you disagree about the PkgEnd.

source/components/parser/psargs.c
Rongronggg9: Thanks for pointing it out. It makes sense! I will fix it and rebase the PR to relieve your burden. :-)

maciejwieczorretman:
Modified the patch message a little to note down where that failure happened. Merged manually, thanks :)

PR #1199, opened 2026-06-11
ACPI 6.5 Add support for the MISC table

pawelchm-intel:
ACPI 6.5 Introduces MISC table and the definition for Miscellaneous GUIDed Table Entries. This commit adds support for this table, specifically in compiler, disassembler and template generator.
Refer to section 5.2.33. "Miscellaneous GUIDed Table Entries" of ACPI 6.5 specification for further information.


source/common/dmtable.c
```diff
@@ -1173,6 +1174,17 @@ AcpiDmDumpTable (
                             ACPI_CAST_PTR (char, Target));
             break;
 
+        case ACPI_DMT_MISC_DATA:
+            /*
+             * Calculate the length of the vendor data for the MISC table:
+             * Length = (Current Subtable ptr + EntryLength) -
+             *          Start of the vendor data (Target)
+             */
+            ByteLength = ((ACPI_CAST_PTR (char, Table) +
```
maciejwieczorretman: If EntryLength is smaller than 28 it will wrap around and underflow into max UINT32. Then after this switch block there is this check at line 1238:
```C
        if (CurrentOffset + ByteLength > TableLength)
```
which then will pass if EntryLength gets very big - which will pass since SubtableLength is also set to a non-zero value.

So I'd think we also need the same check that you have in the special dump function, whether the EntryLength is big enough

source/common/dmtbinfo2.c
```diff
@@ -1165,6 +1165,29 @@ ACPI_DMTABLE_INFO           AcpiDmTableInfoMchi[] =
...
+{
+    ACPI_DMT_TERMINATOR
+};
+
+/* GUIDed Entries */
+
+ACPI_DMTABLE_INFO           AcpiDmTableInfoMisc0[] =
+{
+    {ACPI_DMT_UUID,     ACPI_MISC0_OFFSET (EntryGuid[0]),          "Entry GUID", 0},
+    {ACPI_DMT_UINT32,   ACPI_MISC0_OFFSET (EntryLength),           "Entry Length", 0},
```
maciejwieczorretman: Without the DT_LENGTH flag you'd have to pass the size manually when compiling a table, not sure if it won't then accept bigger values too, like way bigger than the real size

source/compiler/dttemplate.h
```diff
@@ -1370,6 +1370,19 @@ const unsigned char TemplateMadt[] =
     0x00,0x00                                 /* 000002A8    ".."       */
 };
 
+const unsigned char TemplateMisc[] =
```
maciejwieczorretman: iasl -T MISC gives:
```C
*/ Incorrect checksum, should be D4 */
```
which I think pretty much no other tables get. I think setting some byte to D4 will fix this but I'll have to verify how that's handled elsewhere.

source/include/actbl2.h
```diff
@@ -1373,6 +1374,33 @@ enum AcpiKeypProtocolType
...
+ * Conforms to section "Miscellaneous GUIDed Table Entries" of ACPI
+ * Specification v6.5
+ *
+ ******************************************************************************/
+
+typedef struct acpi_misc_guid_entry {
+    UINT8       EntryGuid[16];
+    UINT32      EntryLength;
+    UINT32      Revision;
+    UINT32      ProducerId;
```
maciejwieczorretman: This should be 4 characters (UINT8 array) not 1 4xbyte number - spec says this is a string like 'ABCD'

source/common/dmtbdump2.c
```diff
@@ -1176,9 +1176,6 @@ AcpiDmDumpMcfg (
     UINT32                  Offset = sizeof (ACPI_TABLE_MCFG);
     ACPI_MCFG_ALLOCATION    *Subtable;
 
-
-    /* Main table */
```
maciejwieczorretman, line 1180: Is this here by mistake? It's related to a different table right?

source/common/dmtable.c
```diff
@@ -585,6 +585,7 @@ const ACPI_DMTABLE_DATA     AcpiDmTableData[] =
     {ACPI_SIG_MADT, NULL,                   AcpiDmDumpMadt, DtCompileMadt,  TemplateMadt},
     {ACPI_SIG_MCFG, NULL,                   AcpiDmDumpMcfg, DtCompileMcfg,  TemplateMcfg},
     {ACPI_SIG_MCHI, AcpiDmTableInfoMchi,    NULL,           NULL,           TemplateMchi},
+    {ACPI_SIG_MISC, AcpiDmTableInfoMisc,    AcpiDmDumpMisc, DtCompileMisc,  TemplateMisc},
```
maciejwieczorretman: I'm pretty sure putting AcpiDmTableInfoMisc here doesn't make a difference because it's used as a default method if you didn't specify any custom Dump/Compile handlers. Most entries have either the Info or Dump/Compile combination.

source/common/dmtbdump2.c
```diff
@@ -1213,6 +1210,58 @@ AcpiDmDumpMcfg (
...
+        return;
+    }
+
+    Subtable = ACPI_ADD_PTR (ACPI_MISC_GUID_ENTRY, Table, Offset);
+    while (Offset < Table->Length)
+    {
+
+        AcpiOsPrintf ("\n");
+        Status = AcpiDmDumpTable (Table->Length, Offset, Subtable,
+            sizeof (UINT8) * 16 + sizeof (UINT32) * 3, AcpiDmTableInfoMisc0);
```
maciejwieczorretman: Do we want to skip the data part when dumping the table?

source/common/dmtbdump2.c
```diff
@@ -1213,6 +1210,58 @@ AcpiDmDumpMcfg (
...
+        if (ACPI_FAILURE (Status))
+        {
+            return;
+        }
+
+        SubtableLength = Subtable->EntryLength;
+        if (SubtableLength < (sizeof (UINT8) * 16 + sizeof (UINT32) * 3))
```
maciejwieczorretman: Maybe we should put the:
```C
sizeof (UINT8) * 16 + sizeof (UINT32) * 3
```
into a constant? Or just make it:
```C
#define ACPI_MISC_MIN_ENTRY_LENGTH 28
```

maciejwieczorretman (changes requested):
I want to test a thing or two yet but leaving comments I think are relevant

source/compiler/dttemplate.h
maciejwieczorretman: Also this is not alphabetical, this entry should be after Mchi, the same as you did it in source/compiler/dtcompiler.h

source/include/actbl2.h
```diff
@@ -1373,6 +1374,33 @@ enum AcpiKeypProtocolType
 
 #define ACPI_KEYP_F_TVM_USABLE      (1)
 
+/*******************************************************************************
```
maciejwieczorretman: This is also not alphabetically places, since it's before L. It should go after Mchi the same as the extern template line did.

source/tools/acpisrc/astable.c
```diff
@@ -592,6 +592,7 @@ ACPI_TYPED_IDENTIFIER_TABLE           AcpiIdentifiers[] = {
     {"ACPI_TABLE_MRRM",                     SRC_TYPE_STRUCT},
     {"ACPI_TABLE_MSCT",                     SRC_TYPE_STRUCT},
     {"ACPI_TABLE_MSDM",                     SRC_TYPE_STRUCT},
+    {"ACPI_TABLE_MISC",                     SRC_TYPE_STRUCT},
```
maciejwieczorretman: This should also go after line 589 to be in alphabetical order

maciejwieczorretman (changes requested):
Also just some alphabetical order comments :b

source/common/dmtbdump2.c
pawelchm-intel: oh yes, it's a leftover ;)

source/common/dmtbdump2.c
pawelchm-intel: I assumed so, as it was intentionally left unspecified

source/common/dmtbdump2.c
pawelchm-intel: If we'd use this constant anywhere else, why not ;)

source/compiler/dttemplate.h
pawelchm-intel: I assumed the template is meant as an example for users to fill in or edit manually. I put this comment just as a highlight, as the value is template can be random (the actual checksum will be calculated anyway) :)

source/common/dmtbdump2.c
maciejwieczorretman: Do you mean the size is unspecified? We do know how long it is, right? It is encoded in EntryLength after all when someone writes and compiles the table.

maciejwieczorretman:
I think I solved the zero-length vendor data - look at this branch:
https://github.com/maciejwieczorretman/acpica/tree/pr_misc
I made a commit on top of your MISC patch, and I tested it with a few variants of 'iasl -T MISC' and then assembling and disassembling that. So so far it looks fine and it simplifies your patch a bit because then the special type is not needed for the vendor data

source/common/dmtbdump2.c
pawelchm-intel: I sat at the code and actually the data isn't skipped- the bound check in AcpiDmDumpTable() is Info->Offset > SubtableLength, and Vendor Data is at offset 28, so it's still dumped (checked the template output).
But you are right that this line is misleading, passing length where the parameter refers to the length of the subtable looks, as if the intent were to dump the header only. I'll pass Subtable->EntryLength instead, which is similar to how PMTT does it.

pawelchm-intel:
I've added your patch to the PR :) . Only needed to edit it, so it doesn't conflict with the changes applied in the meantime

maciejwieczorretman (approved):
Looks good after the changes, approved :)

maciejwieczorretman:
Merged to pending, closing

PR #1200, opened 2026-06-17
Utilities: Fix for AcpiUtVerifyChecksum warning display

mstrozek:
Remove redundant subtraction which is done already in AcpiUtGenerateChecksum.

Looks like it is a leftover from a refactor [1] where AcpiTbVerifyChecksum used to call AcpiTbChecksum which does not do the substraction (like the Acpi(Dm\Ut)GenerateChecksum does). 

Found by using iasl to generate checksum for SWFT tables.

[1]: https://github.com/acpica/acpica/commit/8ac4e5116f59d6f9ba2fbeb9ce22ab58237a278f


SaketADumbre:
Thanks a lot @mstrozek for this small but very useful fix!

PR #1203, opened 2026-06-22
Get rid of strncpy to stay in line with the Linux kernel

watologo1:
Closes: https://lore.kernel.org/all/79e9e913-0fb1-4110-804b-c3b5d0edafe4@kernel.org
Reported-by: Jiri Slaby <jirislaby@kernel.org>

### DIFF
--- a/source/components/utilities/utnonansi.c
+++ b/source/components/utilities/utnonansi.c
@@ -351,10 +351,23 @@ AcpiUtSafeStrncpy (
     char                    *Source,
     ACPI_SIZE               DestSize)
 {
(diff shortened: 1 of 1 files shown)


rafaeljw (commented):
I'm not sure if the new code is any better than the existing one.

strncpy() as used here is completely safe and the Linux kernel has its own library function that can be used here, so let it be.

rafaeljw:
This issue has been addressed in Linux (in 7.2-rc1), no need for ACPICA change.

rafaeljw:
No need for the code changes included in this pull request, closing.

PR #1204, opened 2026-06-30
iASL: Complete KEYP support

pawelchm-intel:
Add the iASL disassembler, table compiler, and template support for the KEYP (Key Programming Interface for Root Complex Integrity and Data Encryption) table.


SaketADumbre:
Follow-up fix is now available upstream.

Problem identified:
- The KEYP disassembler trusted malformed subtable metadata (especially variable root-port payload sizing), which could cause misparsing across subtable boundaries.
- The KEYP compiler trusted declared RootPortCount rather than validating against actual provided root-port entries.

Solution implemented:
- Disassembler hardening in KEYP path:
  - Validate subtable header bounds.
  - Validate subtable length against remaining table bytes.
  - Validate minimum config-unit size.
  - Validate root-port payload shape and exact RootPortCount-to-payload-size match.
- Compiler hardening in KEYP path:
  - Count actual root-port entries from input and fail if it does not match declared RootPortCount.

Validation rerun:
- KEYP round-trip still passes byte-exactly (ASL -> AML -> DSL -> AML).
- Malformed KEYP inputs are now rejected cleanly instead of being misparsed.

Upstream branch:
- https://github.com/acpica/acpica/tree/review/pr-1204-keyp

Commit:
- https://github.com/acpica/acpica/commit/caf29b252f019fb45948ba34ec20a1127792913d

SaketADumbre:
Hi @pawelchm-intel 

I fixed a couple of issues with the help of Copilot to ensure that we don't parse the subfields incorrectly in case a FW bug shows the incorrect number of "entries". We would ideally want to stop compilation/disassembly and issue an error in that case of mismatch. Thanks a lot for this contribution, it looks ready to be merged once you pull my commit in this PR!

pawelchm-intel:
I cleaned up this PR. I completed my changes and squashed my two commits into one.

In Saket's commit, I fixed an incorrect message identifier used for the entry count mismatch check (dttable1.c:3674). ASL_MSG_INVALID_LENGTH is rendered as "Length is larger than Min/Max window", which is misleading when reporting a port count mismatch:

Error 6049 - Length is larger than Min/Max window ^ (KEYP root port count 3 does not match actual entries 2)

A similar validation of the locality count in SLIT (dttable2.c:2607) uses ASL_MSG_ENTRY_LIST ("Invalid entry initializer list"), which is a much better fit for this case.

I've also added SoBs to the message of that commit.

maciejwieczorretman:
I'm looking into the PR right now

source/compiler/dttable1.c
```diff
@@ -3433,6 +3433,124 @@ DtCompileIvrs (
...
+                DtInsertSubtable (ParentTable, Subtable);
+                RootPortCount++;
+            }
+
+            if (RootPortCount != ConfigUnit->RootPortCount)
+            {
+                snprintf (MsgBuffer, sizeof (MsgBuffer),
+                    "KEYP root port count %u does not match actual entries %u",
+                    ConfigUnit->RootPortCount, RootPortCount);
+                DtFatal (ASL_MSG_INVALID_LENGTH, SubtableStart, MsgBuffer);
```
maciejwieczorretman: DtCompileSlit has a similar mismatch error message and it uses the ASL_MSG_ENTRY_LIST which will give the 'invalid entry initializer list' error message instead of the one for ASL_MSG_INVALID_LENGTH: 'Length is larger than Min/Max window'. Error message for ASL_MSG_INVALID_LENGTH doesn't make much sense together with the 'root port count X doesn't match actual entries Y'
```suggestion
                DtFatal (ASL_MSG_ENTRY_LIST, SubtableStart, MsgBuffer);
```

source/compiler/dttable1.c
```diff
@@ -3433,6 +3433,124 @@ DtCompileIvrs (
...
+                {
+                    return (Status);
+                }
+
+                ParentTable = DtPeekSubtable ();
+                DtInsertSubtable (ParentTable, Subtable);
+                RootPortCount++;
+            }
+
+            if (RootPortCount != ConfigUnit->RootPortCount)
```
maciejwieczorretman, line 3531: Table 2-2 in KEYP spec says that RootPortCount can't be zero. So this check is missing here

source/common/dmtbdump2.c
```diff
@@ -782,6 +782,139 @@ AcpiDmDumpIvrs (
...
+            }
+
+            RootPortLength = Subtable->Length - sizeof (ACPI_KEYP_CONFIG_UNIT);
+            if (RootPortLength % sizeof (ACPI_KEYP_RP_INFO))
+            {
+                AcpiOsPrintf ("Invalid KEYP root port array size\n");
+                return;
+            }
+
+            if ((((UINT32) ConfigUnit->RootPortCount) * sizeof (ACPI_KEYP_RP_INFO)) !=
```
maciejwieczorretman, line 871: Same thing that is in dttable1.c just in the other direction. Table 2-2 in KEYP spec says that RootPortCount can't be zero. So this check is missing here.

source/common/dmtbdump2.c
```diff
@@ -782,6 +782,139 @@ AcpiDmDumpIvrs (
...
+        {
+            AcpiOsPrintf ("Invalid KEYP subtable length\n");
+            return;
+        }
+
+        switch (Subtable->Type)
+        {
+        case ACPI_KEYP_TYPE_CONFIG_UNIT:
+
+            ConfigUnit = ACPI_CAST_PTR (ACPI_KEYP_CONFIG_UNIT, Subtable);
```
maciejwieczorretman, line 856: Table 2-2 specifies that ProtocolType can be only 1 or 2 - others are reserved. I see AcpiKeypProtocolType is even added to actbl2.h, just unused here.

Also the Version field is not checked and Table 2-2 also says that only value '1' is okay right now, others are reserved.

So these two things should be checked somewhere below I think.

source/compiler/dttable1.c
```diff
@@ -3433,6 +3433,124 @@ DtCompileIvrs (
...
+                &Subtable);
+            if (ACPI_FAILURE (Status))
+            {
+                return (Status);
+            }
+
+            ParentTable = DtPeekSubtable ();
+            DtInsertSubtable (ParentTable, Subtable);
+
+            ConfigUnit = ACPI_SUB_PTR (ACPI_KEYP_CONFIG_UNIT,
```
maciejwieczorretman, line 3511: Same as in dmtbdump2.c, Table 2-2 specifies that ProtocolType can be only 1 or 2 - others are reserved. I see AcpiKeypProtocolType is even added to actbl2.h, just unused here. The Version field is not checked and Table 2-2 also says that only value '1' is okay right now, others are reserved.

So these two things should be checked somewhere here and errors raised since these values are illegal technically.

maciejwieczorretman (changes requested):
Since the second commit has no description I suppose it can be squashed into the first one? Otherwise I left a few comments on stuff that isn't checked but the specification says should be reserved.

source/common/dmtbdump2.c
pawelchm-intel: You're right. Until a new revision of the spec allows version 2, I'm assuming that ProtocolType can only be 1 or 2, while Version must be 1.
In the compiler, this will be treated as a hard error (ASL_MSG_RESERVED_VALUE) to prevent building an invalid table.
In the disassembler, I'll just print a warning message but still allow the table to be disassembled for user to be able to check what's happening there.

source/compiler/dttable1.c
pawelchm-intel: You're right, I'll add the check

source/common/dmtbdump2.c
pawelchm-intel: as above, I'll add the check :)

source/compiler/dttable1.c
pawelchm-intel: Same thing as in the comment above, I'll address it in the way I described in my reply there.

maciejwieczorretman (approved):
No further comments on the code, looks nice :)
On the git side I'd still squash the two commits together and add Co-developed-by to Saket's signed-off-by tag if you intend to keep it

pawelchm-intel:
Ok, good idea. I've squashed them.

maciejwieczorretman:
Merged to pending, closing

PR #1205, opened 2026-07-02
Fix PCC OperationRegion command offsets

sudeep-holla:
ACPI 6.3, section 5.5.2.4.7.3, states that the PCC Operation Region is associated with the region of shared memory that follows the PCC signature.

The generic and extended PCC shared memory layouts include the 4-byte signature at offset 0, so their raw shared memory COMMAND fields are at offsets 4 and 12 respectively. Since AML field offsets for the PCC OperationRegion are relative to the region after that signature, ACPICA must look for those COMMAND fields at OperationRegion offsets 0 and 8.

Adjust the generic and master subspace command checks to use those OperationRegion-relative offsets. Otherwise writes to the COMMAND field can fail to invoke the PCC address space handler at the offset described by the PCC OperationRegion definition.

Fixes: a4849944e80f ("ACPI 6.3: add PCC operation region support for AML interpreter")

### DIFF
--- a/source/components/executer/exfield.c
+++ b/source/components/executer/exfield.c
@@ -191,12 +191,13 @@ static const UINT8      AcpiProtocolLengths[] =
 
 /*
  * The following macros determine a given offset is a COMD field.
- * According to the specification, generic subspaces (types 0-2) contains a
- * 2-byte COMD field at offset 4 and master subspaces (type 3) contains a 4-byte
- * COMD field starting at offset 12.
+ * According to the specification, the PCC OperationRegion begins after
+ * the PCC signature. The raw shared memory COMD offsets of 4 for generic
+ * subspaces (types 0-2) and 12 for master subspaces (type 3) therefore
+ * appear at OperationRegion offsets 0 and 8.
  */
-#define GENERIC_SUBSPACE_COMMAND(a)     (4 == a || a == 5)
-#define MASTER_SUBSPACE_COMMAND(a)      (12 <= a && a <= 15)
+#define GENERIC_SUBSPACE_COMMAND(a)     ((a) < 2)
+#define MASTER_SUBSPACE_COMMAND(a)      (((a) - 8) < 4)
 
 
 /*******************************************************************************
--- a/source/components/executer/exfield.c
+++ b/source/components/executer/exfield.c
@@ -53,13 +53,13 @@ static const UINT8      AcpiProtocolLengths[] =
 
 /*
  * The following macros determine a given offset is a COMD field.
- * According to the specification, generic subspaces (types 0-2) contains a
- * 2-byte COMD field at offset 4 and master subspaces (type 3) contains a 4-byte
- * COMD field starting at offset 12.
+ * According to the specification, the PCC OperationRegion begins after
+ * the PCC signature. The raw shared memory COMD offsets of 4 for generic
+ * subspaces (types 0-2) and 12 for master subspaces (type 3) therefore
+ * appear at OperationRegion offsets 0 and 8.
  */
-#define GENERIC_SUBSPACE_COMMAND(a)     (4 == a || a == 5)
-#define MASTER_SUBSPACE_COMMAND(a)      (12 <= a && a <= 15)
-
+#define GENERIC_SUBSPACE_COMMAND(a)     (0 == a || a == 1)
+#define MASTER_SUBSPACE_COMMAND(a)      (8 <= a && a <= 11)
 
 /*******************************************************************************
  *


rafaeljw (dismissed):

SaketADumbre:
Thanks for the fix, @sudeep-holla — the OperationRegion-relative offset correction is clearly right.

While reviewing this PR against the ACPI 6.6 spec, I found a few related issues and have put together two follow-on commits on branch [`fix/pcc-opregion-offsets`](https://github.com/acpica/acpica/tree/fix/pcc-opregion-offsets) that extend and complete the fix.

---

### Finding 1 — `GENERIC_SUBSPACE_COMMAND` was dead code

The macro was defined but never called in the write path. `AcpiExWriteDataToField` only checked `MASTER_SUBSPACE_COMMAND`, so writes to the COMD field of generic subspaces (types 0–2) never invoked the PCC region handler. This is fixed by wiring both macros into the write-path check.

### Finding 2 — comment said "type 3" but the offset check covers types 3–4

`ACPI_PCCT_EXT_PCC_SLAVE` (type 4) uses the same `ACPI_PCCT_EXT_PCC_SHARED_MEMORY` layout as the master, so its COMMAND field is also at OperationRegion offset 8–11. The updated comment now says "extended PCC subspaces (types 3–4)".

### Finding 3 — Type 5 (HW Registers, ACPI 6.4) was completely broken

`ACPI_PCCT_TYPE_HW_REG_COMM_SUBSPACE` has no shared-memory buffer — it communicates entirely through hardware registers. The existing code unconditionally allocated `InternalPccBuffer` for every PCC OperationRegion and returned `AE_AML_NO_OPERAND` when the buffer was missing, which would always fire for Type 5. The fix:

- Adds `AcpiDsPccIsHwRegSubspace()` in `dsfield.c` that walks the PCCT table to detect Type 5 channels, and skips buffer allocation for them.
- Restructures the read/write paths in `exfield.c` so that a null `InternalPccBuffer` falls through to the normal `AcpiExAccessRegion` dispatch instead of erroring — allowing Type 5 fields to be served by the registered address-space handler.

### ASLTS test updates

The existing PCC tests `m745` (generic) and `m746` (master) had `Offset(4)` in their Field definitions, inadvertently matching the old broken raw-memory offsets. Both methods are updated to place COMD at the correct OperationRegion-relative offsets (0 for generic, 8 for master) with matching expected values.

---

Commits on [`fix/pcc-opregion-offsets`](https://github.com/acpica/acpica/tree/fix/pcc-opregion-offsets):
- [`4e82c3e91e3b`](https://github.com/acpica/acpica/commit/4e82c3e91e3b295d367cf5152a42e212e7863a92) — executer: fix PCC OperationRegion COMD field offset detection
- [`0c3ef90db1cb`](https://github.com/acpica/acpica/commit/0c3ef90db1cbecdae756653130cd7266866e4f76) — executer/dispatcher: add Type 5 PCC subspace (HW Registers) support

@sudeep-holla — would appreciate your review, especially on the Type 5 fallthrough approach and whether there are any ARM-side firmware implications for the COMD offset change we should be aware of.

SaketADumbre:
### Verification test artifact

The following standalone ASL file can be used to independently verify the corrected COMD field offsets on any build. Compile with `iasl` and run under `acpiexec`; both tests should print `PASS`.

```sh
iasl pcc_opregion_offsets.asl
acpiexec -b "execute \_SB.TPCC" pcc_opregion_offsets.aml
```

<details>
<summary><code>pcc_opregion_offsets.asl</code></summary>

```asl
/*
 * pcc_opregion_offsets.asl
 *
 * Standalone verification test for PCC OperationRegion COMD field offsets.
 *
 * Background
 * ----------
 * ACPI 6.3 s5.5.2.4.7.3: a PCC OperationRegion begins immediately AFTER
 * the 4-byte PCC Signature in shared memory.  AML field offsets are
 * therefore 4 bytes less than the corresponding raw shared-memory offsets:
 *
 *   Generic shared memory (types 0-2):
 *     Raw +0  Signature (4 bytes)  <-- excluded from OperationRegion
 *     Raw +4  Command   (2 bytes)  --> OperationRegion offset 0-1  (COMD)
 *     Raw +6  Status    (2 bytes)  --> OperationRegion offset 2-3
 *
 *   Extended PCC shared memory (types 3-4):
 *     Raw +0  Signature (4 bytes)  <-- excluded from OperationRegion
 *     Raw +4  Flags     (4 bytes)  --> OperationRegion offset 0-3
 *     Raw +8  Length    (4 bytes)  --> OperationRegion offset 4-7
 *     Raw +12 Command   (4 bytes)  --> OperationRegion offset 8-11 (COMD)
 *
 * Running this test
 * -----------------
 *   $ iasl pcc_opregion_offsets.asl
 *   $ acpiexec -b "execute \_SB.TPCC" pcc_opregion_offsets.aml
 *
 * The acpiexec PCC handler fills InternalPccBuffer with sequential bytes
 * (Buffer[i] = i) on any COMD write.  Expected readback values follow
 * directly from that behaviour and are documented inline below.
 *
 * Both PCC channel indices used here (0x1, 0x3) are pre-registered by
 * the acpiexec built-in PCC address-space handler.
 */

DefinitionBlock ("pcc_opregion_offsets.aml", "DSDT", 2, "INTEL ", "PCCTEST", 1)
{
    Scope (\_SB)
    {
        /*
         * Generic PCC subspace (types 0-2), channel 0x1, 16-byte region.
         * COMD at OperationRegion offset 0 (raw shared-memory offset 4).
         */
        OperationRegion (PCC1, PCC, 0x1, 16)
        Field (PCC1, AnyAcc, NoLock, Preserve)
        {
            COMD, 16,   /* OperationRegion offset 0-1  */
            STAT, 16,   /* OperationRegion offset 2-3  */
            COSP, 64,   /* OperationRegion offset 4-11 */
        }

        /*
         * Extended PCC subspace (types 3-4), channel 0x3, 22-byte region.
         * COMD at OperationRegion offset 8 (raw shared-memory offset 12).
         */
        OperationRegion (PCC3, PCC, 0x3, 22)
        Field (PCC3, AnyAcc, NoLock, Preserve)
        {
            FLGS, 32,   /* OperationRegion offset 0-3   */
            LNGT, 32,   /* OperationRegion offset 4-7   */
            ECMD, 32,   /* OperationRegion offset 8-11  */
            ECSP, 80,   /* OperationRegion offset 12-21 */
        }

        /*
         * TPCC - run both PCC subspace tests, return 0 on full pass.
         *
         * After each COMD write the acpiexec handler fills the buffer
         * with bytes 0x00..N-1.  Expected readbacks:
         *
         *   m745 (generic, OLEN=16):
         *     COMD = 0x0100, STAT = 0x0302, COSP = 0x0B0A090807060504
         *
         *   m746 (extended, OLEN=22):
         *     FLGS = 0x03020100, LNGT = 0x07060504, COMD = 0x0B0A0908
         *     COSP[n] = 0x0C + n  (n = 0..9)
         */
        Method (TPCC, 0, Serialized)
        {
            Name (FAIL, 0)

            /* ---- m745: generic subspace ---- */
            Debug = "TEST m745: generic PCC subspace (types 0-2)"

            STAT = Buffer(2){0x55, 0x55}
            COSP = Buffer(8){0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA, 0xAA}
            COMD = Buffer(2){0x01, 0x02}     /* triggers PCC handler */

            If (COMD != 0x0100) {
                Debug = "FAIL m745: COMD"
                Debug = COMD
                FAIL = 1
            }
            If (STAT != 0x0302) {
                Debug = "FAIL m745: STAT"
                Debug = STAT
                FAIL = 1
            }
            If (COSP != 0x0B0A090807060504) {
                Debug = "FAIL m745: COSP"
                Debug = COSP
                FAIL = 1
            }
            If (!FAIL) { Debug = "PASS m745" }

            /* ---- m746: extended subspace ---- */
            Debug = "TEST m746: extended PCC subspace (types 3-4)"
            Name (F746, 0)
            Name (OLEN, 22)
            Name (CLEN, 12)

            FLGS = Buffer(4){0xAA, 0xAA, 0xAA, 0xAA}
            LNGT = Buffer(4){0xBB, 0xBB, 0xBB, 0xBB}
            ECSP = Buffer(10){0xCC, 0xCC, 0xCC, 0xCC, 0xCC,
                               0xCC, 0xCC, 0xCC, 0xCC, 0xCC}
            ECMD = Buffer(4){0x01, 0x02, 0x03, 0x04}   /* triggers PCC handler */

            If (FLGS != 0x03020100) {
                Debug = "FAIL m746: FLGS"
                Debug = FLGS
                F746 = 1
            }
            If (LNGT != 0x07060504) {
                Debug = "FAIL m746: LNGT"
                Debug = LNGT
                F746 = 1
            }
            If (ECMD != 0x0B0A0908) {
                Debug = "FAIL m746: ECMD"
                Debug = ECMD
                F746 = 1
            }

            Name (IDX, 0)
            While (OLEN - CLEN) {
                If (DerefOf(ECSP[IDX]) != (CLEN + IDX)) {
                    Debug = "FAIL m746: ECSP mismatch at index"
                    Debug = IDX
                    F746 = 1
                }
                IDX++
                OLEN--
            }
            If (!F746) { Debug = "PASS m746" }

            Return (FAIL | F746)
        }
    }
}
```
</details>

source/components/executer/exfield.c
```diff
@@ -191,12 +191,13 @@ static const UINT8      AcpiProtocolLengths[] =
...
- * COMD field starting at offset 12.
+ * According to the specification, the PCC OperationRegion begins after
+ * the PCC signature. The raw shared memory COMD offsets of 4 for generic
+ * subspaces (types 0-2) and 12 for master subspaces (type 3) therefore
+ * appear at OperationRegion offsets 0 and 8.
  */
-#define GENERIC_SUBSPACE_COMMAND(a)     (4 == a || a == 5)
-#define MASTER_SUBSPACE_COMMAND(a)      (12 <= a && a <= 15)
+#define GENERIC_SUBSPACE_COMMAND(a)     ((a) < 2)
+#define MASTER_SUBSPACE_COMMAND(a)      (((a) - 8) < 4)
```
maciejwieczorretman, line 200: I think the MASTER_SUBSPACE_COMMAND() should still consider a range of byte offsets (8-11)
Otherwise the fields of FLAGS and LENGTH will also be taken into account.
So I think the MASTER_SUBSPACE_COMMAND() part could look like the old version, just the numbers should be down by 4
```C
#define MASTER_SUBSPACE_COMMAND(a)      (8 <= a && a <= 11)
```

maciejwieczorretman (changes requested):
Left one comment about the byte offsets. Not sure about the GENERIC_SUBSPACE_COMMAND() not being used since you removed the call @sudeep-holla in 2021 (41be6afacfda) due to a kernel problem. So not sure if it should be reused now or the macro definition removed.

rafaeljw (commented):
The concern regarding the upper bound is valid in my view

source/components/executer/exfield.c
sudeep-holla: OK, I will more back to old style of checking with only offsets changed. I think the same comment applies for GENERIC_SUBSPACE_COMMAND(a) as well.

sudeep-holla:
> The concern regarding the upper bound is valid in my view

Agreed, will move back to existing style of checking with offsets changed.

sudeep-holla:
> Thanks for the fix, @sudeep-holla — the OperationRegion-relative offset correction is clearly right.
> 
> While reviewing this PR against the ACPI 6.6 spec, I found a few related issues and have put together two follow-on commits on branch [`fix/pcc-opregion-offsets`](https://github.com/acpica/acpica/tree/fix/pcc-opregion-offsets) that extend and complete the fix.
> ### Finding 1 — `GENERIC_SUBSPACE_COMMAND` was dead code
> 
> The macro was defined but never called in the write path. `AcpiExWriteDataToField` only checked `MASTER_SUBSPACE_COMMAND`, so writes to the COMD field of generic subspaces (types 0–2) never invoked the PCC region handler. This is fixed by wiring both macros into the write-path check.
> ### Finding 2 — comment said "type 3" but the offset check covers types 3–4
> 
> `ACPI_PCCT_EXT_PCC_SLAVE` (type 4) uses the same `ACPI_PCCT_EXT_PCC_SHARED_MEMORY` layout as the master, so its COMMAND field is also at OperationRegion offset 8–11. The updated comment now says "extended PCC subspaces (types 3–4)".
> ### Finding 3 — Type 5 (HW Registers, ACPI 6.4) was completely broken
> 
> `ACPI_PCCT_TYPE_HW_REG_COMM_SUBSPACE` has no shared-memory buffer — it communicates entirely through hardware registers. The existing code unconditionally allocated `InternalPccBuffer` for every PCC OperationRegion and returned `AE_AML_NO_OPERAND` when the buffer was missing, which would always fire for Type 5. The fix:
> 
>     * Adds `AcpiDsPccIsHwRegSubspace()` in `dsfield.c` that walks the PCCT table to detect Type 5 channels, and skips buffer allocation for them.
> 
>     * Restructures the read/write paths in `exfield.c` so that a null `InternalPccBuffer` falls through to the normal `AcpiExAccessRegion` dispatch instead of erroring — allowing Type 5 fields to be served by the registered address-space handler.
> 
> 
> ### ASLTS test updates
> 
> The existing PCC tests `m745` (generic) and `m746` (master) had `Offset(4)` in their Field definitions, inadvertently matching the old broken raw-memory offsets. Both methods are updated to place COMD at the correct OperationRegion-relative offsets (0 for generic, 8 for master) with matching expected values.
> 
> Commits on [`fix/pcc-opregion-offsets`](https://github.com/acpica/acpica/tree/fix/pcc-opregion-offsets):
> 
>     * [`4e82c3e91e3b`](https://github.com/acpica/acpica/commit/4e82c3e91e3b295d367cf5152a42e212e7863a92) — executer: fix PCC OperationRegion COMD field offset detection
> 
>     * [`0c3ef90db1cb`](https://github.com/acpica/acpica/commit/0c3ef90db1cbecdae756653130cd7266866e4f76) — executer/dispatcher: add Type 5 PCC subspace (HW Registers) support
> 
> 
> @sudeep-holla — would appreciate your review, especially on the Type 5 fallthrough approach and whether there are any ARM-side firmware implications for the COMD offset change we should be aware of.

@SaketADumbre I assume the above is orthogonal to this change right ? I will take a look at your changes soon.

sudeep-holla:
> Left one comment about the byte offsets. Not sure about the GENERIC_SUBSPACE_COMMAND() not being used since you removed the call @sudeep-holla in 2021 ([41be6af](https://github.com/open-acpica/acpica/commit/41be6afacfdaec2dba3a5ed368736babc2a7aa5c)) due to a kernel problem. So not sure if it should be reused now or the macro definition removed.

Ah are you suggesting to drop it completely ? Should that be a separate change as this is being pushed as fix and that is rather more clean-up category.

maciejwieczorretman:
> > Left one comment about the byte offsets. Not sure about the GENERIC_SUBSPACE_COMMAND() not being used since you removed the call @sudeep-holla in 2021 ([41be6af](https://github.com/open-acpica/acpica/commit/41be6afacfdaec2dba3a5ed368736babc2a7aa5c)) due to a kernel problem. So not sure if it should be reused now or the macro definition removed.
> 
> Ah are you suggesting to drop it completely ? Should that be a separate change as this is being pushed as fix and that is rather more clean-up category.

If you're still able it'd be best to check if that same issue happens after you brought back GENERIC_SUBSPACE_COMMAND() in the if() clause. Since the offsets that this PR fixes were wrong then (right?) I assume it should be fine now?

maciejwieczorretman:
Okay after some though, since we have a new scheme where we push things through a testing branch, I think I'll put it there and run some kernel checks from there

maciejwieczorretman:
Merged to testing, let me know if we want to revert that 2021 patch

sudeep-holla:
> Merged to testing, let me know if we want to revert that 2021 patch

I think revrerting will break the systems. Since Saket is addressing other subspace(1/2/5) types, it can be addressed as part of it IMO.

maciejwieczorretman:
Okay, then I'll close this PR now, I think we'll move your commit from testing to the master branch next week

PR #1212, opened 2026-07-27
Add macro for each bit of the Flags field in the IOVT IOMMU structure

lixianglai:
According to the IOVT specification, define macros for each bit of the Flags field in the IOVT IOMMU structure:

Bit 0: PCI device flag.
1: The IOMMU is a PCI device.
0: The IOMMU is a platform device.

Bit 1: Proximity domain valid.
Bit 2: Whether the devices managed by the IOMMU can be classified by PCI segment
1: All devices under the root PCI bridge can be managed.
0: Manageable devices are placed in the device entry list.

Bit3: Hardware capability support.
Bit4: MSI interrupt address bypass supported.
Bit5-31: Reserved, must be zero.

### DIFF
--- a/source/include/actbl2.h
+++ b/source/include/actbl2.h
@@ -1190,6 +1190,14 @@ typedef struct acpi_iovt_iommu
 
 } ACPI_IOVT_IOMMU;
 
+/* Masks for Flags field above */
+
+#define ACPI_IOVT_IOMMU_PCI_DEVICE           (1<<0)
+#define ACPI_IOVT_IOMMU_PXM_VALID            (1<<1)
+#define ACPI_IOVT_IOMMU_MAGAGE_BY_SEGMENT    (1<<2)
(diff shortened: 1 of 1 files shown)


maciejwieczorretman:
Should these definitions be used anywhere?

lixianglai:
We are submitting an IOMMU driver‑related patch to the Linux community. In this patch, we will parse the IOVT table and use these macro definitions during the process.

lixianglai:
These macros were originally defined directly in the kernel's actbl2.h header file, but the kernel community recommends that we submit them to the ACPICA community now and then synchronise them into the kernel.

Thanks！

maciejwieczorretman:
Ah okay, I found the discussion on the kernel mailing list
I'll add the mail archive link to the commit message so it's easier to link it to your driver patches

lixianglai:
OK, thanks!

maciejwieczorretman:
I merged to master manually so closing the PR
Added link to the kernel lore and slightly shortened the subject

PR #1213, opened 2026-08-04
executer/dispatcher: fix PCC OperationRegion COMD field offsets and add Type 5 support

SaketADumbre:
Extends and completes the fix from PR #1205 (@sudeep-holla).

## Background

ACPI 6.3 §5.5.2.4.7.3 specifies that a PCC OperationRegion begins immediately **after** the 4-byte PCC Signature in shared memory. AML field offsets are therefore 4 bytes less than the corresponding raw shared-memory offsets:

| Layout | Raw offset | Field | OperationRegion offset |
|---|---|---|---|
| Generic (types 0–2) | 4–5 | Command (UINT16) | **0–1** |
| Extended (types 3–4) | 12–15 | Command (UINT32) | **8–11** |

---

## Commit 1 — fix COMD offset macros and wire generic subspace handler

`source/components/executer/exfield.c`

**Problem 1:** `GENERIC_SUBSPACE_COMMAND` and `MASTER_SUBSPACE_COMMAND` were comparing against raw shared-memory offsets (4/5 and 12–15) instead of OperationRegion-relative offsets (0/1 and 8–11). This caused the PCC region handler to never be invoked on COMD writes for any subspace type, because the offsets used in real firmware AML never matched.

**Problem 2:** `GENERIC_SUBSPACE_COMMAND` was defined but never called in the write path — `AcpiExWriteDataToField` only checked `MASTER_SUBSPACE_COMMAND`. Writes to the COMD field of generic subspaces (types 0–2) silently succeeded without invoking the PCC handler.

**Problem 3:** The macro comment said "master subspaces (type 3)" but `ACPI_PCCT_EXT_PCC_SLAVE` (type 4) uses the identical `ACPI_PCCT_EXT_PCC_SHARED_MEMORY` layout, so the offset check legitimately covers both types 3 and 4.

**Fix:** Correct both macros to use OperationRegion-relative offsets, wire `GENERIC_SUBSPACE_COMMAND` into the write-path check alongside `MASTER_SUBSPACE_COMMAND`, and update the comment.

---

## Commit 2 — add Type 5 (HW Registers) PCC subspace support

`source/components/dispatcher/dsfield.c`, `source/components/executer/exfield.c`

**Problem:** `ACPI_PCCT_TYPE_HW_REG_COMM_SUBSPACE` (ACPI 6.4, type 5) uses hardware registers for communication and has **no shared-memory buffer**. The existing code unconditionally allocated `InternalPccBuffer` for every PCC OperationRegion and returned `AE_AML_NO_OPERAND` when the buffer was missing, which always fires for type 5 regions.

**Fix:**
- Add `AcpiDsPccIsHwRegSubspace()` in `dsfield.c` — walks the PCCT table to identify type 5 channels by index and skips `InternalPccBuffer` allocation for them.
- Restructure the read and write paths in `exfield.c` so that a null `InternalPccBuffer` falls through to the normal `AcpiExAccessRegion` dispatch (the registered PCC address-space handler) instead of returning an error. This is the correct path for type 5 regions.

---

## ASLTS test updates

The existing PCC tests `m745` (generic) and `m746` (master) in `tests/aslts/src/runtime/collections/functional/region/regionfield.asl` were written using `Offset(4)` in the Field definitions, which inadvertently matched the old broken raw-memory offsets. Both are updated to place COMD at the correct OperationRegion-relative offsets with matching expected values.

---

## Testing

Full ASLTS suite run clean. A standalone verification ASL (`pcc_opregion_offsets.asl`) is attached in the review comment on PR #1205 for independent validation.

### DIFF
--- a/source/components/dispatcher/dsfield.c
+++ b/source/components/dispatcher/dsfield.c
@@ -633,6 +633,64 @@ AcpiDsGetFieldNames (
 
 /*******************************************************************************
  *
+ * FUNCTION:    AcpiDsPccIsHwRegSubspace
+ *
+ * PARAMETERS:  ChannelIndex    - PCC channel number from OperationRegion address
+ *
+ * RETURN:      TRUE if the channel maps to a Type 5 (HW Registers) subspace
+ *
+ * DESCRIPTION: Walk the PCCT to determine whether a PCC channel uses the
+ *              HW-Registers-based model, which has no shared memory buffer.
+ *
+ ******************************************************************************/
+
+static BOOLEAN
+AcpiDsPccIsHwRegSubspace (
+    UINT8                   ChannelIndex)
+{
+    ACPI_STATUS             Status;
+    ACPI_TABLE_HEADER       *Table;
+    ACPI_SUBTABLE_HEADER    *Subtable;
+    UINT32                  Offset;
+    UINT8                   Channel = 0;
+    BOOLEAN                 Result = FALSE;
+
+
+    Status = AcpiGetTable (ACPI_SIG_PCCT, 1, &Table);
+    if (ACPI_FAILURE (Status))
+    {
+        return (FALSE);
+    }
+
+    Offset = sizeof (ACPI_TABLE_PCCT);
+    while (Offset < Table->Length)
+    {
+        if (Offset + sizeof (ACPI_SUBTABLE_HEADER) > Table->Length)
+        {
+            break;
+        }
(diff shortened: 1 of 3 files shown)


maciejwieczorretman:
Also the second commit says it wires the GENERIC_SUBSPACE_COMMAND into the code but it actually does the type 5 handling. And there is no description in the commit message about the type 5.

maciejwieczorretman:
Now we'll focus on the main PCC fix, we can get back to this later if needed

Issue #996, opened 2025-01-10
iASL Warning: There were 212 external control methods found during disassembly, but only 468 were resolved (4294967040 unresolved).


sliedes:
Possibly related (I see the same wraparound in outputs pasted there): #702 

Git tag: R2024_12_12

I just decompiled a dsdt.dat from a ThinkPad E14 Gen 6
(attached: [dsdt.dat.zip](https://github.com/user-attachments/files/18383247/dsdt.dat.zip)). It shows this warning with obviously bogus numbers:

```
     * iASL Warning: There were 212 external control methods found during
     * disassembly, but only 468 were resolved (4294967040 unresolved). Additional
     * ACPI tables may be required to properly disassemble the code. This
     * resulting disassembler output file may not compile because the
     * disassembler did not know how many arguments to assign to the
     * unresolved methods. Note: SSDTs can be dynamically loaded at
     * runtime and may or may not be available via the host OS.
```

I see similar numbers in bug #702 (a wider issue) from 2023, although they were not the focus there.

maciejwieczorretman:
Merged the fix to pendng, closing this one

PR #1221, opened 2026-08-24
iASL: Fix the truncated count of external control methods

AbhinavMir:
#996


maciejwieczorretman:
Hi, generally all looks okay and I want to merge the patches
But I noticed you didn't leave your signed-off-by git tag, could you add them to the two patches? It is later required when porting patches to the kernel. I can also add them if you're okay with that

maciejwieczorretman:
I merged your fix patch to pending with you as Originally-by: tag. We probably don't want the shell script in the tests since it only tests for this one very specific condition and it's very unlikely it will ever be a problem again.
Thanks again for the fix :)

AbhinavMir:
Thank you! I was not monitoring my inbox well!

PR #1223, opened 2026-08-27
Add the _Ixx object

maciejwieczorretman:
ACPI 6.6 specification, chapter 5.6.4.3 mentions an _Ixx object (where
xx is a two-digit hexadecimal number) which presence indicates that the
corresponding GPE is enabled in the low power S0 idle state.

Add a new reserved name type to symbolize an object. Since the
specification mentions that just the presence of the object is required,
any one of Name(), Method(), Device(), etc. with the _Ixx name should be
a valid indication.

Update comments and the disassembler path.


rafaeljw (approved):
Looks good to me

pawelchm-intel (approved):
Thank you for the PR, looks good

PR #1224, opened 2026-08-27
Follow up memory safety fixes

maciejwieczorretman:
Pull requests #1216 and #1217 provided some very good memory safety fixes.
While checking them out I noticed more similar problems in direct vicinity of
where these two PRs made changes. More detailed description of both problems are
located in the patch messages.

### DIFF
--- a/source/compiler/dttable2.c
+++ b/source/compiler/dttable2.c
@@ -1576,30 +1576,34 @@ DtCompilePptt (
         {
         case ACPI_PPTT_TYPE_PROCESSOR:
 
+            if (!Subtable->Buffer)
+            {
+                AcpiOsPrintf ("Pptt subtable buffer pointer is NULL\n");
+                return (AE_NULL_OBJECT);
+            }
+
             PpttProcessor = ACPI_SUB_PTR (ACPI_PPTT_PROCESSOR,
                 Subtable->Buffer, sizeof (ACPI_SUBTABLE_HEADER));
-            if (PpttProcessor)
+
+            /* Compile initiator proximity domain list */
+
+            PpttProcessor->NumberOfPrivResources = 0;
+            while (*PFieldList)
             {
-                /* Compile initiator proximity domain list */
-
-                PpttProcessor->NumberOfPrivResources = 0;
-                while (*PFieldList)
+                Status = DtCompileTable (PFieldList,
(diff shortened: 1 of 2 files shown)


source/compiler/dttable2.c
```diff
@@ -1576,30 +1576,34 @@ DtCompilePptt (
...
 
+            if (!Subtable->Buffer)
+            {
+                AcpiOsPrintf ("Pptt subtable buffer pointer is NULL\n");
+                return (AE_NULL_OBJECT);
+            }
+
             PpttProcessor = ACPI_SUB_PTR (ACPI_PPTT_PROCESSOR,
                 Subtable->Buffer, sizeof (ACPI_SUBTABLE_HEADER));
-            if (PpttProcessor)
```
pawelchm-intel, line 1581: Nice catch :)

source/components/namespace/nsxfname.c
```diff
@@ -579,6 +579,15 @@ AcpiInstallMethod (
     }
     Path = AcpiPsGetNextNamestring (&ParserState);
 
+    /*
+     * Path can be NULL, and Aml can be PkgEnd which combined with the Aml++
+     * a few lines below will cause AmlLength to underflow.
+     */
+    if (!Path || (ParserState.Aml >= ParserState.PkgEnd))
```
pawelchm-intel, line 586: Good catch, and a guard to add!

source/components/namespace/nsxfname.c
pawelchm-intel, line 576: I'm wondering that we are checking ParserState.PkgEnd > ParserState.AmlEnd (and also dereferencing ParserState.AmlEnd in AcpiPsGetNextPackageLength, which is called from AcpiPsGetNextPackageEnd), while AmlEnd doesn't appear to be set anywhere before this line..

maciejwieczorretman: Look at the pending branch, I just recently merged a PR that fixed that setting AmlEnd problem. 

Namespace: Initialize ParserState.AmlEnd in AcpiInstallMethod
(972ae1e276ddfe5c8c727978d179a776c9ccad83)

pawelchm-intel: Thanks, I missed that one.

PR #1225, opened 2026-08-29
Respect OSDT overrides in acpi_ns_execute_table()

dvdkon:
Currently there is only partial support for OSDT tables (which should override already-defined methods) -- it works when tables are loaded via `AcpiNsOneCompleteParse()`, but not for `AcpiNsParseTable()`, which is what's actually used in e.g. `acpiexec` or the Linux kernel. This PR remedies that by adding support.

### DIFF
--- a/source/components/dispatcher/dswload2.c
+++ b/source/components/dispatcher/dswload2.c
@@ -287,7 +287,14 @@ AcpiDsLoad2BeginOp (
         {
             /* Execution mode, node cannot already exist, node is temporary */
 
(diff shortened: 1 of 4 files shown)


dvdkon:
Is there some path to merging this or getting OSDT tables working, or is that not possible and I should find some other way to fix my issue? @d-tatianin

d-tatianin:
I'm afraid I'm not the right person to ask that, I just pointed out what I knew. You should ask the maintainers probably.

PR #1229, opened 2026-09-09
iASL: Validate the entire Switch pattern before rewriting the parse tree

maciejwieczorretman:
While loops with if/else-if ladders that return on each step would disassemble incorrectly where the if/else-if part would become case instead.
Fixes issue #1219


rafaeljw (approved):
Looks reasonable to me.

maciejwieczorretman:
Merged to pending

PR #1231, opened 2026-09-09
ACPICA: Add ACPI_MSG_DEBUG to set KERN_DEBUG for debug output

superm1:
acpi_debug_print() emits ACPI debug output using multi-part acpi_os_printf() and acpi_os_vprintf() calls without any loglevel prefix.

When acpi_os_vprintf() processes a chunk without a loglevel prefix (printk_get_level() returns 0), it prepends KERN_CONT. Because the initial header chunk lacks a loglevel prefix, it either attaches to a preceding unfinalized log message (inheriting its level) or, when split across printk buffer records, falls back to the default message loglevel rather than KERN_DEBUG. This can cause dmesg to misinterpret and colorize debug messages as errors.

Define ACPI_MSG_DEBUG as KERN_DEBUG in aclinux.h (with an empty fallback in acutils.h for ACPICA OS-independence) and prefix debug message headers and context switch output in acpi_debug_print() with ACPI_MSG_DEBUG so that printk records are properly tagged with KERN_DEBUG.

### DIFF
--- a/source/components/utilities/utdebug.c
+++ b/source/components/utilities/utdebug.c
@@ -182,6 +182,7 @@ AcpiDebugPrint (
         if (ACPI_LV_THREADS & AcpiDbgLevel)
         {
             AcpiOsPrintf (
+                ACPI_MSG_DEBUG
                 "\n**** Context Switch from TID %u to TID %u ****\n\n",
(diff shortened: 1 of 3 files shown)


maciejwieczorretman:
Looks good, merged to pending :)

PR #1234, opened 2026-09-16
UBRT sub-table compiler and disassembler support

hanjun-guo:
This series completes the UBRT (UnifiedBus Root Table) tooling on top
of the main-table support merged in commit 332440884 ("acpica: add
UBRT disassembler, compiler, and integration support"). The original
work introduced the UBRT signature, the main table dump/compile path,
and the sub-table entry array, but not the sub-table body. This series adds the 
sub-table bodies that follow the entry array, dispatched by the entry Type field
(ubc / ummu / ub_reserved_mem).

Patch 1 (compiler) adds the sub-table field info tables and extends
DtCompileUbrt() to:

- compile each sub-table body in order: the common header, the
type-specific fields, and the nested UBC/UMMU structures and memory
ranges;
- auto-count UbcCount / UmmuCount / MemoryRangesCount;
- enforce the spec-mandated fixed entry sizes (384/160 bytes) by
zero-padding short Vendor Info buffers and rejecting oversized ones,
so generated tables are always conformant;
- back-fill the sub-table TotalSize (via DT_LENGTH) and the 32-bit
checksum, defined such that the valid data area sums to zero as
little-endian 32-bit words, with a partial trailing word zero-padded
in its high bytes.

Patch 2 (disassembler) extends AcpiDmDumpUbrt() to walk the same
sub-table bodies, dumping the common header once per body followed by
the type-specific fields and the UBC/UMMU structures and memory ranges
— including the feature set and interrupt attribute flag bits, using
the info tables introduced in patch 1. Unknown types are dumped as
their common header with a note. The UBRT template is updated with one
entry of each of the three sub-table types so that iasl -T UBRT
emits the full table structure.

Testing:
- iasl -T UBRT emits the full template.
- ASL -> AML -> DSL -> AML round-trips are byte-identical for the
template, full-size/short Vendor Info and entries-only cases.
- make builds clean with -Werror; iasl / acpiexec / acpibin
used in the round-trip tests are built via aslts.sh.

Patch 1 and patch 2 are split for review but patch 2 relies on the
field info tables added in patch 1, so they should be applied
together.

Hanjun Guo (2):
  compiler: Add UBRT sub-table compiler support
  disassembler: Add UBRT sub-table disassembler support

 source/common/dmtbdump3.c    | 188 +++++++++++++++++++-
 source/common/dmtbinfo2.c    | 110 +++++++++++-
 source/compiler/dttable2.c   | 407 ++++++++++++++++++++++++++++++++++++++++++++-
 source/compiler/dttemplate.h |  94 ++++++++++-
 source/include/acdisasm.h    |   7 +
 source/include/actbinfo.h    |  10 ++
 6 files changed, 807 insertions(+), 9 deletions(-)


maciejwieczorretman:
I'm checking out the PR now

maciejwieczorretman:
So the idea is that the subtables are laid out just after the main table? I thought the idea in the docs was that the main table has entries for subtables, and each entry has a physical address that points into that subtable. Right now I think that address field is unused? (I'm still browsing what's happening where)

I mean this looks like it should work, I'm just wondering if there will be no side effects from moving away from the physical pointers idea. Also I think there was a comment in actbl2.h that said that UBRT would use pointers for this purpose, so that'd need to change

source/common/dmtbdump3.c
```diff
@@ -663,6 +669,182 @@ AcpiDmDumpUbrt (
...
+
+        Type = ACPI_ADD_PTR (ACPI_UBRT_SUBTABLE, Table,
+            sizeof (ACPI_TABLE_UBRT) +
+            (i * sizeof (ACPI_UBRT_SUBTABLE)))->Type;
+
+        UbrtHeader = ACPI_ADD_PTR (ACPI_UBRT_HEADER, Table, BodyOffset);
+        TotalSize = UbrtHeader->TotalSize;
+
+        if ((TotalSize < sizeof (ACPI_UBRT_HEADER)) ||
+            (BodyOffset + TotalSize) > Table->Length)
```
maciejwieczorretman: TotalSize is firmware input and if TotalSize is near the UINT32 limit then if you add something to it it will wrap around and pass the check, which would be incorrect. So I think it would be more correct to do:
```C
TotalSize > (Table->Length - BodyOffset))
```

maciejwieczorretman (commented):
One thing I noticed so far. I think there is also one thing missing from a subtable count vs size validation corner case but I'm still looking into how that works

